PR #63, opened 2014-11-14
Intial support of a ptrans distribution

tsegismont:
For now, there's only a simple env/start script facility, shamelessly copied from the Wildfly equivalents.

We can add instructions on how to use as a service later.

### DIFF
--- a/clients/ptranslator/pom.xml
+++ b/clients/ptranslator/pom.xml
@@ -115,4 +115,30 @@
     </plugins>
   </build>
 
+  <profiles>
+    <profile>
+      <id>publish</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>dist</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>single</goal>
+                </goals>
+                <configuration>
+                  <descriptor>src/assembly/dist/assembly.xml</descriptor>
+                </configuration>
+              </execution>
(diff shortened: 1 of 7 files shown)


tsegismont:
Dropped the env and start scripts. The distribution file is just a zip with:
* license
* sample `ptrans.conf` file
* sample `log4j.xml` file
* `ptrans-all` jar file
* `README` file with instructions on how to run/stop/configure get help

pilhuhn:
Why o why?
Providing a readme is good, but you know what each admin is going to do first? Re-create the start.sh script, as he does not want to remember how to start that thing via java -jac ... -<uh, forgot which options> ...

Also wrt Licence in assets - can't be the license from the projects root be pulled in?

tsegismont:
> Why o why?

The answer is 42 :)

> Providing a readme is good, but you know what each admin is going to do
> first? Re-create the start.sh script, as he does not want to remember
> how to start that thing via java -jac ... - ...
>

I wasn't happy at all with the start and env script. I had barely (and 
shamelessly) copied/pasted the scripts from Wildfly.

So instead of waiting to get something acceptable, I went ahead and 
merged the changes so that the next release gets a distribution file 
with minimal instructions.

I'll create a new JIRA for a start script.

In December I had the opportunity to talk with someone from the OSAS 
team who is managing servers. He works with Zabbix. When I asked him 
what would make him inclined to test something else, his answers was "I 
probably won't test the tool if I can't install a native package".
So it's not only start scripts which I hope to add in the end.

> Also wrt Licence in assets - can't be the license from the projects root
> be pulled in?

It could be done this way. I feared that a change in the project 
structure would lead to a missing LICENSE file.

PR #92, opened 2014-12-17
Keycloak integration

jpkrohling:


jpkrohling:
I've just added a new commit with changes to the paths, so that the tenant ID is resolved from the URL, as discussed on the mailing list. 

However, there was already an endpoint making use of the "/tenants" prefix: TenantHandler. As far as I could see, it's meant to be used by users on the admin realm, to create new tenants. As a temporary measure, I've renamed it to "tenant", but either a better name or a refactoring is needed. Perhaps all the "/tenants/" (and now "tenant") endpoints could be grouped into a single handler? 

What's still missing after this change:

- Rebasing (it seems there are conflicts)
- Re-test of the UI, to see which changes are required there. I expect that only a couple of changes would be required though.

@mtho11 : About the sed script vs. Windows: what will be the strategy for the start.sh? If it's going to be replaced entirely by a Maven goal, then I'd wait until this is done and then adapt it with the Keycloak changes. The changes to the start.sh are not required by the Keycloak integration per se, and those particular sed commands are only to pre-configure some realms and other json files required by Keycloak, so, those can be removed for now.

jpkrohling:
@mtho11  By the way: I've just tried to rebase this branch and it seems it gets a lot of conflicts on the start.sh. So, depending on the approach to be taken with regards to replace it with a Maven goal, I'll either discard my changes or adapt it based on the latest start.sh

mtho11:
@jpkrohling we can discuss the whole scripting/mvn strategies for other platforms (like Win/Mac) once @pilhuhn is back (Jan 6th). @stefannegrea may have some ideas on this as well.

jpkrohling:
Alright, sounds reasonable. The current state of the PR is:

- REST endpoints have been changed, so that /tenants/{tenantId}/... is being used by MetricsHandler on tenant-specific operations
- REST endpoint for TenantHandler have been changed to /tenant/
- UI has been changed to reflect the changes: from what I could see, only metric-data-service was affected.

Still to discuss:

- Refactor of the REST endpoints, to keep all /tenants/ (ie: tenant-specific endpoints for metrics + tenant operations from TenantHandler) in one handler
- start.sh 

Because of the start.sh, I'll not be rebasing this PR for now.

jpkrohling:
@pilhuhn , so, should I remove the changes related to the start.sh from the PR?

ui/console/src/main/webapp/scripts/controllers/realm-selector-controller.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+    }
+
+    export class SelectRealmModalInstanceController {
+        public static $inject = ['$modalInstance', 'Auth'];
+        constructor(private $modalInstance:any, private Auth:Services.AuthService) {
+        }
+
+        realms = ['acme-other-affairs', 'acme-roadrunner-affairs'];
```
jpkrohling: Just realized that this wasn't removed/changed before sending the PR! This needs to be externalized.

jpkrohling:
I've rebased the branch, but there are two things missing:

- Unprotected REST endpoint to query the possible realms (to use on the console's realm selector modal)
- start.sh - I've stashed my changes and need to refactor it based on the new version

jpkrohling:
So, the last two points were done, but instead of having an unprotected REST endpoint, I'm adding a static realms.json file, created by start.sh . That's because I'm not sure if having it being served as a REST endpoint is the best option. This way, we have more time to discuss and improve this, if needed.

Also, it would be good to check the `sed` parts, as I'm not sure if OS X's `sed` would be happy with those. As I don't have an OS X, it's a bit hard for me try it out.

mtho11:
I'm getting some compile errors this time.

INFO] RHQ Metrics ....................................... SUCCESS [2.215s]
[INFO] RHQ Metrics Test Utils ............................ SUCCESS [1.109s]
[INFO] RHQ Metrics API ................................... SUCCESS [1.945s]
[INFO] Embedded Cassandra for RHQ Metrics ................ SUCCESS [0.223s]
[INFO] Embedded Cassandra Service for RHQ Metrics ........ SUCCESS [0.720s]
[INFO] Embedded Cassandra EAR for RHQ Metrics ............ SUCCESS [1.816s]
[INFO] RHQ Metrics Core .................................. SUCCESS [5.849s]
[INFO] RHQ Metrics REST-api .............................. SUCCESS [3.967s]
[INFO] RHQ Metrics clients ............................... SUCCESS [0.195s]
[INFO] clients-common .................................... SUCCESS [0.556s]
[INFO] Protocol Translator ............................... SUCCESS [1.149s]
[INFO] RHQ Metrics Rest Tests ............................ FAILURE [1.601s]
[INFO] metrics-explorer .................................. SKIPPED
[INFO] RHQ Metrics Console ............................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 21.880s
[INFO] Finished at: Mon Jan 19 20:47:06 PST 2015
[INFO] Final Memory: 76M/3017M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.wildfly.plugins:wildfly-maven-plugin:1.0.2.Final:start (start-wildfly) on project rhq-metrics-rest-tests: JBOSS_HOME '.keycloak/keycloak-appliance-dist-all-1.1.0.Beta2/keycloak' is not a valid directory. -> [Help 1]

Not sure what I'm missing or what the completion criteria is here to merge.
Maybe @pilhuhn  can try compiling on OS/X?

jpkrohling:
That's because the Keycloak used in the tests is not available. Similarly to the Cassandra setup, there's a .travis.install.keycloak.sh . Running it before performing the build should prepare the environment.

mtho11:
@jpkrohling ok compile is fine. start.sh is complaining now.
./start.sh: line 422: unexpected EOF while looking for matching `''

Something like http://mojo.codehaus.org/versions-maven-plugin/set-mojo.html might work or just mvn writing to properties file that start.sh reads from would also remove the OS specific stuff. As we will soon needs a Windows version this might be the best route.

I know others use FreeBSD as a proxy for OS/X.

jpkrohling:
I was under the impression that non-Linux scripts would be worked out at a later time. Is this not true anymore? In any case, it's a bit hard to convert it entirely to Maven, as there are other tools than sed involved, like openssl and uuidgen. The output of those is what sed is using to replace the placeholders on the templates. Perhaps a better longer term solution is to create our own plugin that would create the certs and UUIDs, generating the files on demand. But again, these files are only for a "--dev" setup, so, not sure if it's worth the trouble.

By the way: I've just seen that Github says the PR has conflicts. I'll rebase it again and send an update to this PR tomorrow.

pilhuhn:
Creating a maven plugin to generate those sounds good for the whole keycloak project (?)

jpkrohling:
@mtho11 , I was just checking that line, to see if I could guess what could be done to get it working on OS X, but this seems strange. The original line from master looks like this:

https://github.com/rhq-project/rhq-metrics/blob/master/start.sh#L313

    WFLY_VERSION=`grep "<version.wildfly>" pom.xml | sed -E 's/^.*y>(8.*l)<.*$/\1/'`

And the line from your message is:
https://github.com/jpkrohling/rhq-metrics/blob/JPK-KCAuthentication-Draft/start.sh#L422

    KC_VERSION=`grep "<version.keycloak>" pom.xml | sed -E 's/^.*k>(1.*)<.*$/\1/'`

Can you confirm that the original one works?

jpkrohling:
Closing, as efforts on this PR are now on hold. See thread "Keycloak + Hawt.io" on the hawkular-dev mailing list.

PR #94, opened 2014-12-17
Embedded Cassandra

stefannegrea:


jsanda:
This is intended primarily for development and/or getting started with RHQ Metrics right? What are the steps that the developer/user has to go through with this approach? What are the pros/cons vs using the [wildfly-cassandra](https://github.com/rhq-project/wildfly-cassandra) subsystem @heiko-braun wrote? I know that the latter builds from Cassandra's trunk, but I wouldn't think modifying it to working against a tagged version of Cassandra would be too difficult.

stefannegrea:
The goal was to have a second method of delivering an embedded Cassandra but not dependent on the Wildfly container. It is very simple to deploy (just an ear that will be published to maven repositories) and the code lives in the same repository as the main project.

As you pointed out, this is targeted for developers and people that want to test RHQ Metrics. I would like this to be release 0.2.6 to get feedback on the concept and refine it further. 

The widlfy-cassandra subsystem has better integration with the main container targeted by the project. So beyond release 0.2.6 of RHQ Metrics, the plan is to look into making the external subsystem an alternative to deploying an embedded Cassandra. The two methods are not necessarily competing and we could ultimately drive the code for both out of the same repository.

metrics-core/pom.xml
```diff
@@ -51,6 +51,30 @@
...
+      <version>4.0.23.Final</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.cassandra</groupId>
+      <artifactId>cassandra-clientutil</artifactId>
+      <version>2.1.1</version>
+    </dependency>
+
+    <dependency>
```
jsanda: I didn't notice this before. Why does metrics-core have cassandra server dependencies now? And the version of Netty should be the same as the one used by the DataStax driver.

jsanda:
There is one thing I would like to see if we could change before merging. I am -1 on the compile dependencies on the new modules. I am thinking about the changes to the RHQMetrics class in particular. We are not working directly with any of Cassandra's internal APIs. We are still going over the native, CQL protocol. So from a client stand point, we don't really care whether Cassandra is running in the same process or not. Do we really need any special logic in the metrics-core module in order to use the in-memory node? For instance, could the ```MetricsServiceCassandra.verifyNodeIsUp``` method be moved into the EmbeddedCassandraService class?

metrics-core/pom.xml
stefannegrea: It is a mistake on my side to not clean the pom properly. Earlier in the development cycle, I had the code in the metrics-core to get things going. I will fix it.

PR #96, opened 2014-12-18
Add Kubernetes support and simplify Docker build

vnugent:


docker/Dockerfile
```diff
@@ -1,34 +1,9 @@
 #Dockerfile for rhq-metrics
 
-FROM jboss/base-jdk:7
+FROM vnguyen/wildfly-jdk8
```
pilhuhn: Is there no jboss/* version available?

docker/README.adoc
```diff
@@ -1,5 +1,8 @@
-==== docker build
-`docker build --force-rm=true --no-cache=true -t rhq-metrics:latest .`
```
pilhuhn: Should we keep that for people that only run Docker and not Kubernetes?

vnugent: Right.  I'll update the Readme with instructions to launch Cassandra container and link with rhq-metrics the plain Docker way.

docker/Dockerfile
vnugent: I could be more specific with WildFly version.  vnguyen/wildfly-jdk8 image was built from 8.2.0.Final.

PR #102, opened 2014-12-19
Added ListSeriesEvent and extracted ListSeriesHandler

tsegismont:
This is not ready to be merged, I just pushed an example in order to gather feedback.

This is an attempt to make REST handlers classes shorter. CDI events decouples the JAX-RS annotated bean from the actual handling code (and vice versa).

We could have for example:
```java
@Observes @NewData NumericMetricEvent
```

Then both InfluxHandler and vanilla handler could fire a ```@NewData NumericMetricEvent```

WDYT?


tsegismont:
Closing without merging as:
* no interest shown
* we might want to minimize adherence to CDI

PR #103, opened 2014-12-19
JSONP: convert entities when media type is 'application/javascript'

tsegismont:
Before the change, the JsonPFilter allowed JSONP for any resource, regardless of the media type (it could even wrap XML).

Now JSONP needs to be explicitely allowed (with @Produces)

Being a MessageBodyWriter JsonPProvider lets the JAX-RS runtime determine the content length.


tsegismont:
Rebased on master.

Can we merge this one?

My opinion is that allowing JSONP on demand with `@Produces(APPLICATION_JAVASCRIPT)` is a better option:
* for security reasons (do not enable JSONP on all endpoints by default)
* for standard conformance (content-type)

jsanda:
As long as we have some test coverage in place, I am fine with merging.

tsegismont:
@jsanda I've enabled JSONP on the PingHandler and added some tests (see https://github.com/tsegismont/rhq-metrics/blob/936dad4b729988b3666043262691db9983ebb2d3/rest-tests/src/test/groovy/org/rhq/metrics/rest/JsonPITest.groovy)

PR #107, opened 2014-12-23
JMX on CDI beans with Deltaspike

tsegismont:
This is not ready for merge.

Deltaspike provides tooling to expose CDI beans as JMX beans.

JMX operations could solve issue #95 : Add the ability to discover or set C* nodes post-start up


tsegismont:
Closing without merging.

I did some experiments with Deltaspike. Then considering that:
* we want to minimize dependencies of production code
* we'd use a narrow part of it (JMX)
I came to the conclusion that it wasn't a good idea.

I'm not throwing away all the work on configuration and management though. I'm going to open a new PR with the relevant parts of this one.

PR #119, opened 2015-01-12
Config management

tsegismont:
This pull requests introduces JMX support (see commit 26da171) and configuration through dependency injection (see commit  969cf9b)

Here's an example of JMX support:
```java
@MBean(keyProperties = @KeyProperty(key = "type", value = "Configuration"))
@Description("Configuration management")
@MXBean
public interface ConfigurationManagement {
    @Description("Lists the configuration property keys. Some properties may be missing if they were not loaded yet.")
    String[] getConfigurationPropertyKeys();

    @Impact(MBeanOperationInfo.INFO)
    @Description("Get the value of a configuration property")
    String getConfigurationProperty(
        @Description("External form of the configuration property key") @OperationParamName("key") String key);

    @Impact(MBeanOperationInfo.ACTION)
    @Description("Set the value of a configuration property")
    void updateConfigurationProperty(@OperationParamName("key") String key, @OperationParamName("value") String value);
}
```

And of configuration:
```java
    @Inject
    @Configurable
    @ConfigurationProperty(BACKEND)
    private String backend;
```

I tested the new features manually and I'm working on some automated tests.

Any comments?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright 2014 Red Hat, Inc.
+  ~ Copyright 2015 Red Hat, Inc.
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
@@ -153,6 +153,13 @@
         <artifactId>antlr4-runtime</artifactId>
         <version>${antlr.version}</version>
       </dependency>
+      <!-- Tests dependencies -->
+      <dependency>
+        <groupId>junit</groupId>
+        <artifactId>junit</artifactId>
+        <version>4.11</version>
+        <scope>test</scope>
+      </dependency>
       <dependency>
         <groupId>org.assertj</groupId>
         <artifactId>assertj-core</artifactId>
@@ -160,9 +167,9 @@
         <scope>test</scope>
       </dependency>
       <dependency>
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>4.11</version>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-core</artifactId>
+        <version>1.10.19</version>
         <scope>test</scope>
       </dependency>
       <!-- TestNG is not RHQ-Metrics main testing framework -->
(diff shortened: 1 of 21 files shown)


tsegismont:
More about ```@Configurable``` (reproducing Javadoc here):

Qualifier for an injected field or parameter which value should be fetched from configuration. For example:
```java
       @Configurable
       @ConfigurationProperty(ConfigurationKey.BACKEND)
       @Inject
       String backendType;
```

Configuration is a set of key/value pairs (configuration properties) defined by the following sources, in order of precedence:
* system properties (```-Dkey=value```)
* external ```java.util.Properties``` file, which path is defined by the ```metrics.conf``` system property; by default, ```<user.home>/.metrics.conf``` is used
* internal ```java.util.Properties``` file (```META-INF/metrics.conf```)

Any field or parameter annotated with ```@Configurable``` must also be annotated with ```@ConfigurationProperty```. The value of the latter specifies the configuration property key. Otherwise the configuration producer method will throw an instance of ```IllegalArgumentException```.

In most cases, a configuration property should at least be defined in the internal properties file, in order to provide a default value. Configuration values may be modified at runtime (through JMX). When a bean can/should adapt to such changes, the injected field or parameter type can be wrapped with ```javax.enterprise.inject.Instance```. The configuration value will be fetched on each call to ```javax.enterprise.inject.Instance.get()```.

stefannegrea:
Why JMX?

tsegismont:
> Why JMX?

It's the standard for management, so it gives the ability to work with common tools (JConsole, VisualVM, jmxtrans)

It's widely used:
* Cassandra http://www.datastax.com/documentation/cassandra/2.0/cassandra/operations/ops_monitoring_c.html
* Tomcat https://tomcat.apache.org/tomcat-8.0-doc/monitoring.html
* ActiveMQ http://activemq.apache.org/jmx.html
... etc

Also, I understood that using JMX+Jolokia would make our life easier for future work with Fabric8.

tsegismont:
If there is no consensus on the JMX part of the PR, could we at least merge the configuration part of this PR? If we agree, I'll:
* break this PR into two
* update the config PR to take `rhq-metrics.cassandra-cql-port`, `rhq-metrics.cassandra-nodes` and `cassandra.keyspace` into account

jsanda:
That sounds a like good, reasonable approach.

tsegismont:
Closing this PR without merging as there's no agreement to support JMX.

@jsanda I've created #127 with the "external config" aspects only

PR #135, opened 2015-01-26
ptrans PID file

tsegismont:
Added a "-p" to set the path to the PID file

ptrans will try to acquire a write lock on the PID file

Java has no built-in support for self PID so the PID is provided by
JNR (the library built by the JRuby team).

### DIFF
--- a/clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/Main.java
+++ b/clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/Main.java
@@ -18,12 +18,15 @@ package org.rhq.metrics.clients.ptrans;
 
 import static org.rhq.metrics.clients.ptrans.OptionsFactory.CONFIG_FILE_OPT;
 import static org.rhq.metrics.clients.ptrans.OptionsFactory.HELP_OPT;
+import static org.rhq.metrics.clients.ptrans.OptionsFactory.PID_FILE_OPT;
 
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Properties;
+
+import jnr.posix.POSIXFactory;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
@@ -47,6 +50,7 @@ public class Main {
     private final String[] args;
     private final OptionsFactory optionsFactory;
     private PTrans ptrans;
+    private PidFile pidFile;
 
     private Main(String[] args) {
         this.args = args;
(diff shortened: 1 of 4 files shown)


tsegismont:
Can someone please review this short PR? Thanks

clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/Main.java
```diff
@@ -78,10 +82,18 @@ private void start() throws Exception {
...
+            System.err.printf("Configuration file %s does not exist or is not readable.%n",
+                configFile.getAbsolutePath());
             System.exit(1);
         }
+        if (cmd.hasOption(PID_FILE_OPT)) {
+            File file = new File(cmd.getOptionValue(PID_FILE_OPT));
+            pidFile = new PidFile(file);
+            boolean locked = pidFile.tryLock(POSIXFactory.getPOSIX().getpid());
+            if (!locked) {
+                System.exit(1);
```
pilhuhn: Should we inform the user why it exited?

clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/PidFile.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
+import java.nio.charset.StandardCharsets;
+
+/**
```
pilhuhn: I think (and I am sloppy here often as well) it is a good idea to at least provide a short class level comment about the class.

clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/Main.java
tsegismont: > Should we inform the user why it exited?

It's done in the PidFile class

clients/ptranslator/src/main/java/org/rhq/metrics/clients/ptrans/PidFile.java
tsegismont: > I think (and I am sloppy here often as well) it is a good idea to at
> least provide a short class level comment about the class.

Done. See commit dcc7c44e0fa0be15e85fc8a1a7a729d120c69e5a

tsegismont:
I've just added "PID file must be deleted on exit", commit 2e8ae37002d16c3d59189073d29ea285f66f82d0

It was on my PID file check list but totally missed it in the implementation.

PR #136, opened 2015-01-27
clients-common and ptrans update following REST changes

tsegismont:
* Metrics JSON format has changed
* Post data URI change


pilhuhn:
So 2 out of 4 changed files are formatting?
Ah no, in one formatting change was actually some code change hidden

PR #150, opened 2015-02-11
HWKMETRICS-16 ptrans has no integration tests

tsegismont:
The goal isn't to test a full collectd/ptrans/metrics stack (this could be done in another PR)

The goal is to verify that when metrics REST API changes, ptrans isn't broken.
It happened and I believe there's at least one metrics release under the RHQ org which has a broken ptrans.

So now the ptrans module is built after rest-tests.
It starts a Wildfly server with the metrics REST API.
Then RestForwardingHandlerITest sends a metric through a RestForwardingHandler and verifies that it landed in the database.

Side work (in separate commits)
Move rest-tests Wildfly configuration out of the standard tests resources directory.
Updated rest-tests POM file


stefannegrea:
Good changes!

tsegismont:
Thanks!

pilhuhn:
Great stuff, @tsegismont

tsegismont:
Le 12/02/2015 10:07, Heiko W. Rupp a écrit :
> Great stuff, @tsegismont <https://github.com/tsegismont>
>

Thanks!

PR #155, opened 2015-02-16
Domain Simplification and Refactoring

stefannegrea:


api/metrics-api-jaxrs/pom.xml
```diff
@@ -24,7 +24,7 @@
   <parent>
     <artifactId>hawkular-metrics-parent</artifactId>
     <groupId>org.hawkular.metrics</groupId>
-    <version>0.3.0-SNAPSHOT</version>
+    <version>0.3.1-SNAPSHOT</version>
```
tsegismont: What's the reason for the version bump?

api/metrics-api-jaxrs/pom.xml
```diff
@@ -141,6 +141,7 @@
       <plugin>
         <artifactId>maven-war-plugin</artifactId>
         <configuration>
+          <attachClasses>true</attachClasses>
           <archive>
```
tsegismont: Is that actually needed? Which classes are shared in other projects?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -16,29 +16,20 @@
...
-import static java.util.stream.Collectors.summarizingDouble;
-import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toMap;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
-import static javax.ws.rs.core.Response.Status;
-import static org.hawkular.metrics.api.jaxrs.CustomMediaTypes.APPLICATION_VND_HAWKULAR_WRAPPED_JSON;
 import static org.hawkular.metrics.core.api.MetricsService.DEFAULT_TENANT_ID;
+import gnu.trove.map.TLongObjectMap;
+import gnu.trove.map.hash.TLongObjectHashMap;
```
tsegismont: Minor: shouldn't be here, but after "com"

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -100,355 +96,239 @@
             + "a metric before storing data. Doing so however allows clients to prevent naming collisions and to "
             + "specify tags and data retention.")
     @ApiResponses(value = { @ApiResponse(code = 200, message = "Metric definition created successfully"),
-            @ApiResponse(code = 400, message = "Metric with given id already exists or request is otherwise incorrect"),
```
tsegismont: The create numeric/availability metric part has been updated in #168 . I'll have to rebase the PR once this one is merged.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/callback/NoDataCallback.java
```diff
@@ -0,0 +1,63 @@
...
+    public void onSuccess(Object result) {
+        response.resume(Response.ok().type(MediaType.APPLICATION_JSON_TYPE).build());
+    }
+
+    @Override
+    public void onFailure(Throwable t) {
+        if (t instanceof MetricAlreadyExistsException) {
+            Error errors = new Error("A metric input id already exists " + ":"
+                    + Throwables.getRootCause(t).getMessage());
+            response.resume(Response.status(Status.BAD_REQUEST).entity(errors).type(APPLICATION_JSON_TYPE).build());
```
tsegismont: Lets keep this implementation to merge the PR ASAP.  Just want to let you know that this will change for [HWKMETRICS-26 Tracker - REST-API uses return codes incorrectly](https://issues.jboss.org/browse/HWKMETRICS-26).

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -472,14 +352,54 @@ public void findNumericData(
...
+                }
+            });
         } else {
             if (bucketWidthSeconds == 0) {
                 outputFuture = Futures.transform(dataFuture, new CreateSimpleBuckets(start, end, numberOfBuckets,
                     skipEmpty));
             } else {
                 ListenableFuture<List<? extends Object>> bucketsFuture = Futures.transform(dataFuture,
-                    new CreateFixedNumberOfBuckets(numberOfBuckets, bucketWidthSeconds));
+                        new Function<NumericMetric, List<? extends Object>>() {
```
tsegismont: Any reason for making this class anonymous instead of inner?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -704,40 +433,29 @@ public void findAvailabilityData(@Suspended final AsyncResponse asyncResponse,
...
-
-            @Override
-            public void onFailure(Throwable t) {
-                asyncResponse.resume(t);
-            }
+        ListenableFuture<List<Availability>> outputfuture = Futures.transform(future,
+                new Function<AvailabilityMetric, List<Availability>>() {
+                    @Override
+                    public List<Availability> apply(AvailabilityMetric input) {
+                        if (input == null) {
```
tsegismont: Here null is used. In another function above, it's `NoResultException` instead. I believe these functions are perfect candidate for usage of Java 8 Optional (`Optional.ofNullable`). Besides, this code could be turned into something as simple as:
```java
 outputfuture = Futures.transform(future, input-> Optional.ofNullable(input).map(input->input.getData()));
```

api/metrics-api-jaxrs/pom.xml
stefannegrea: Snapshots are published automatically to Nexus, I do not want to force depending projects to upgrade overnight to the new API. There will be a few differences, so it is better to bump the version and let each project migrate at a later time.

api/metrics-api-jaxrs/pom.xml
stefannegrea: This is no longer needed, it was needed when I started experimenting. I will remove it.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
stefannegrea: That code needs to be simplified. It is by far the most complicated method and will need help from @jsanda to migrate that to a simpler version.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
stefannegrea: Unfortunately, we will need to support Java 7 for the time being. There are still a few places to change back to Java 7; we should not add more Java 8 constructs.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/callback/NoDataCallback.java
stefannegrea: The code is now consolidated in a single place. So updating it should be much simpler going forward.

core/metrics-core-impl/pom.xml
```diff
@@ -72,6 +77,31 @@
...
+    <dependency>
+      <groupId>com.wordnik</groupId>
+      <artifactId>swagger-core_2.10</artifactId>
+      <version>${swagger.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+      <scope>provided</scope>
```
tsegismont: Why provided? Here we suppose that the metrics library will always be used in an environment where Jackson is present.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Tenant.java
```diff
@@ -32,8 +35,10 @@
 
     private String id;
 
+    @JsonInclude(Include.NON_EMPTY)
     private List<AggregationTemplate> aggregationTemplates = new ArrayList<>();
```
tsegismont: On the up side, it make our lives easier to annotate the model in the core-api library. On the flip side, now you need to get a Jackson dependency if you want to embed the Metrics core library.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
```diff
@@ -701,9 +700,9 @@ private void dropKeyspace(String keyspace) {
                     return Futures.immediateFuture(results);
                 }
                 List<ResultSetFuture> insertFutures = new ArrayList<>();
-                tags.forEach((k, v) -> dataAccess.insertAvailabilityTag(k, v.orElse(""), data));
+                tags.forEach((k, v) -> dataAccess.insertAvailabilityTag(k, v, data));
                 for (Availability a : data) {
-                    insertFutures.add(dataAccess.updateDataWithTag(a, MetricUtils.flattenTags(tags)));
+                    insertFutures.add(dataAccess.updateDataWithTag(a, tags));
```
tsegismont: I haven't looked at the details, why all these changes in the tags model? (It used to be `Map<String, Optional<String>>` everywhere)

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
jsanda: The code is complicated. I checked with Mike Thompson a while back, and most of it (at least at that time) was not actually used. We need to find out what is actually used, get rid of the rest, and then refactor. If it is not used, I don't want to keep it. It is just technical debt.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Tenant.java
jsanda: I have expressed similar concern about introducing Jackson dependency in the core library. It would be worthwhile to see if we can do things in a way to avoid that.

rest-tests/pom.xml
```diff
@@ -182,6 +120,84 @@
           -Xrunjdwp:transport=dt_socket,address=8797,server=y,suspend=y</maven.surefire.debug>
       </properties>
     </profile>
+
+    <profile>
```
tsegismont: This new profile is not mandatory: you can simply start individual tests (or the whole suite) from your IDE

api/metrics-api-jaxrs/pom.xml
tsegismont: I disagree. If projects depend on SNAPSHOTS it's because they want the latest thing, should it break their code over night. If they want more stability, then let's release 0.3.0 and then bump to 0.3.1-SNASPHOT

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
pilhuhn: On 10 Mar 2015, at 18:01, jsanda wrote:

>>       } else {
>>           if (bucketWidthSeconds == 0) {
>>               outputFuture = Futures.transform(dataFuture, new 
>> CreateSimpleBuckets(start, end, numberOfBuckets,
>>                   skipEmpty));
>>           } else {
>>               ListenableFuture<List<? extends Object>> bucketsFuture 
>> = Futures.transform(dataFuture,
>> -                    new CreateFixedNumberOfBuckets(numberOfBuckets, 
>> bucketWidthSeconds));
>> +                        new Function<NumericMetric, List<? extends 
>> Object>>() {
>
> The code is complicated. I checked with Mike Thompson a while back, 
> and most of it (at least at that time) was not actually used. We need 
> to find out what is actually used, get rid of the rest, and then 
> refactor. If it is not used, I don't want to keep it. It is just 
> technical debt.
>

Yes that (additional) bucketing code can go.
I implemented that some time ago to play with bucketing in the sense 
that we do e.g. not display a full 8h period of data, but only a 60 mins 
time span where all values from x hours: yy minutes in land in the 
respective yy bucket
to better see if there are repeating patterns like a spike on minute 5 
each hour.

-- 

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
tsegismont: *Sigh*. I miss the "if it's not on the devel list, it did not happen"

api/metrics-api-jaxrs/pom.xml
pilhuhn: On 10 Mar 2015, at 18:15, Thomas Segismont wrote:

>
> I disagree. If projects depend on SNAPSHOTS it's because they want the 
> latest thing, should it break their code over night. If they want more 
> stability, then let's release 0.3.0 and then bump to 0.3.1-SNASPHOT

I think here making a huge api change just to have other consumers fix 
their code out of the blue is wrong.
So +1 on release 0.3.0 and then apply the change to 0.3.1, so consumers 
can use 0.3.0 while adapting their
code to those changes.


-- 

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Tenant.java
stefannegrea: This PR is just the first part of a larger effort, but I want to get it merged as soon as possible because it gets out of control. The goal is to have Jackson library part of the core-impl, while core-api be just interfaces.

tsegismont: I can see only way to do this: maintain a separate model, dedicated to transfer in the JAX-RS and WebSocket API. In case you're worried about bean mappings, Gunnar (from the Hibernate team) has created a nice tool for this: http://mapstruct.org/

tsegismont: > This PR is just the first part of a larger effort, but I want to get it
> merged as soon as possible because it gets out of control. The goal is

Agreed.

> to have Jackson library part of the core-impl, while core-api be just
> interfaces.

This goal is precisely what's being questioned :)

Nobody wants core-api alone. (By the way, now that we chose to support 
C* only, I'm not sure why it's still needed.)

jsanda: Even if we only provide a single implementation, separating API from the impl is a cleaner approach IMO and makes it easier for people to consume and also makes it easier to test. Nobody wants core-api alone, but that's all client's should have to worry about. It also makes it possible to provide alternative implementations if someone were so inclined.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
stefannegrea: To simplify the model and API, a tag is a tuple (key,value) with both required. We should provide richer query capabilities, but tag creation should be as simple as possible.

api/metrics-api-jaxrs/pom.xml
stefannegrea: We will probably not release 0.3.0, with snapshots getting published on every build there is no incentive to make this a public official release. 

Once everybody catches up on 0.3.1 or newer we should make the first official release of Hawkular Metrics.

PR #157, opened 2015-02-19
HWKMETRICS-25 ptrans might wait an arbitrary amount of time before forwarding metrics

tsegismont:
Ganglia, statsd and collectd sources use a metric batcher. If the data incoming rate is low and the batch size high, then ptrans might wait an arbitrary amount of time before forwarding metrics.

For example:
* collectd sending only 1 metric, one value per minute
* batch size set to 5
-> then ptrans will wait 5 minutes before sending a batch of 5 metrics

This PR introduces a batch.delay parameter in ptrans configuration file.
Now MetricBatcher class will output a list of metrics when batch.size is reached or no metric was added since (batch.delay) seconds.

### DIFF
--- a/clients/ptranslator/ptrans.conf
+++ b/clients/ptranslator/ptrans.conf
@@ -48,5 +48,7 @@ rest.close-after=200
 # Maximum number of metrics to spool if the server is not reachable
 spool.size=10000
 
-# Minimum batch size of metrics to be forwarded (from one source)
-batch.size=5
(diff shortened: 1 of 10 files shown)


tsegismont:
This PR is ready for review and merge. [The build passes](https://travis-ci.org/hawkular/hawkular-metrics/builds/51348675).

tsegismont:
Thanks!

PR #160, opened 2015-02-25
Allow running rest-test multiple times without emptying the datastore in...

burmanm:
... between. Adds random UUID infront of the tenant_id. This makes the test work when running tests against running Hawkular-metrics instance from the IDE for example. Helps when improving the tests.


tsegismont:
Looks good to me. @jsanda , @stefannegrea any objection?

stefannegrea:
Good change. Merged.

PR #161, opened 2015-03-02
Wildfly Maven Plugin: skip server start when tests are disabled

tsegismont:
Upgrade Wildfly Maven Plugin version -> now we can skip server start when tests are disabled

### DIFF
--- a/api/metrics-api-jaxrs/pom.xml
+++ b/api/metrics-api-jaxrs/pom.xml
@@ -162,8 +162,11 @@
         <groupId>org.wildfly.plugins</groupId>
         <artifactId>wildfly-maven-plugin</artifactId>
         <configuration>
-          <jvmArgs>-Dhawkular-metrics.backend=em_cass -Xdebug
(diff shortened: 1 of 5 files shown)


tsegismont:
Build failed initially because the new Wildfly Maven plugin version requires to use `<javaOpts>` instead of `<jvmArgs>`.

Note that there's a warning which appears now when building the whole project. I've reported the issue: https://github.com/wildfly/wildfly-maven-plugin/issues/31

stefannegrea:
@tsegismont ready to be merged?

tsegismont:
It can be merged, if we all agree that the warning in the build log is not a blocker.

The report I've made to the Wildfly Maven Plugin issue tracker has been categorized has a minor bug. It may be fixed in the next version.

tsegismont:
Rebased on master. I've [pushed a fix for the minor bug](https://github.com/wildfly/wildfly-maven-plugin/pull/32) (build warning, see above). Should we merge or wait until a new version of the plugin is released?

tsegismont:
Looks like setting different ports for ptrans Wildfly servers (see #171) makes the warning go away... Merging.

PR #165, opened 2015-03-04
SonarQube issues - try-with and entrySets

burmanm:
These fix some critical issues in other functions than REST-API. Use of entrySet instead of keySet when all the entries are iterated over and use of try-with instead of Closeable (and also fix potential resource leak at the same time).

### DIFF
--- a/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/DataAccessImpl.java
+++ b/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/DataAccessImpl.java
@@ -611,9 +611,7 @@ public class DataAccessImpl implements DataAccess {
     private ResultSetFuture executeTagsBatch(Map<String, String> tags,
         BiFunction<String, String, BoundStatement> bindVars) {
         BatchStatement batchStatement = new BatchStatement(BatchStatement.Type.UNLOGGED);
-        for (String tag : tags.keySet()) {
-            batchStatement.add(bindVars.apply(tag, tags.get(tag)));
-        }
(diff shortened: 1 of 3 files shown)


tsegismont:
LGTM, except, that the import order is not correct. See https://github.com/hawkular/hawkular-build-tools/blob/master/ide-configs/eclipse/hawkular-eclipse-preferences-java-codestyle-organizeimports.importorder

Can you update your settings and push a new commit to minimize this diff and future ones?

burmanm:
I have https://github.com/hawkular/hawkular-build-tools/blob/master/ide-configs/idea/settings.jar imported. It generates this kind of output.

tsegismont:
Le 04/03/2015 15:45, Michael Burman a écrit :
> I have
> https://github.com/hawkular/hawkular-build-tools/blob/master/ide-configs/idea/settings.jar
> imported. It generates this kind of output.

Then settings.jar is wrong :)

lucasponce:
I put a fix for it in the last settings.jar, the order is still incorrect ?

Thanks,
Lucas

tsegismont:
Maybe Micke imported an older one

burmanm:
"Current branch master is up to date.". So either IDEA is bugging with it or yes.. could be IntelliJ though, as it now shows mixture of rhq import order & hawkular order in the same page.

burmanm:
I'll make an issue on the hawkular-build-tools, settings. jar doesn't include our special rules of imports. So wasn't my IntelliJ

tsegismont:
Micke, any chance you manually change the settings in IDEA before IJ settings.jar is updated? Let's merge this PR soon.

burmanm:
Hep, here you go (forgot the whole thing..). I still have different alphabetic order in SchemaManager, does that need fixing also?

tsegismont:
No, that's for the next PR ;) Thanks Micke!

PR #167, opened 2015-03-05
[HWKMETRICS-29] Matching handling for all tagged data calls

burmanm:
Move all the tagged data data callbacks to their own class and at the same time fix the difference between return codes of these methods (all of them return 204 now on empty resultset).


stefannegrea:
Thanks for the change!

PR #175, opened 2015-03-16
HWKMETRICS-34 Metrics REST API Required for MVP2: Average, Median, 95th Percentile for a date range?

tsegismont:


jsanda:
Do we need our own custom Duration class? Would either org.joda.time.Duration or something from the java.time package suffice?

tsegismont:
It's not meant to be used outside of JAX-RS parameters. I can update the 
Javadoc to make this clear, or if you fear it's going to be misused, I 
can switch to the JODA class.

jsanda:
Maybe I do not understand the purpose. Is it to help with/simplify JSON mapping?

tsegismont:
Le 16/03/2015 16:41, jsanda a écrit :
> Maybe I do not understand the purpose. Is it to help with/simplify JSON
> mapping?

Yes

jsanda:
Thanks for the clarification! Maybe make a note of it in the class javadoc.

stefannegrea:
Thanks Thomas!

PR #179, opened 2015-03-18
HWKMETRICS-35 Availability REST API Calculations Required for MVP2 given a date range?

tsegismont:
Starting PR for HWKMETRICS-35

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
@@ -523,21 +523,30 @@ public class MetricHandler {
 
     @GET
     @Path("/{tenantId}/metrics/availability/{id}/data")
-    @ApiOperation(value = "Retrieve availability data.", response = Availability.class)
-    @ApiResponses(value = { @ApiResponse(code = 200, message = "Successfully fetched availability data."),
-            @ApiResponse(code = 204, message = "No availability data was found.")})
-    public void findAvailabilityData(@Suspended final AsyncResponse asyncResponse,
-        @PathParam("tenantId") String tenantId, @PathParam("id") final String id,
-        @ApiParam(value = "Defaults to now - 8 hours", required = false) @QueryParam("start") Long start,
-        @ApiParam(value = "Defaults to now", required = false) @QueryParam("end") Long end) {
-
+    @ApiOperation(
+            value = "Retrieve availability data. When buckets or bucketDuration query parameter is used, the time "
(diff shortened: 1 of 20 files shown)


tsegismont:
I'll continue working on this tomorrow. After the last commit, now it's "only" a matter of computing the desired stats and adding the new methods in `MetricsService`.

tsegismont:
I've just added `AvailabilityBucketDataPoint` and `AvailabilityBucketedOutputMapper`. An important thing to note about the logic to create the bucket summary:
* each point determines the availability type between this point and the next point
* the first point determines availability type between the start of the bucket and the second point
* the last point determines availability type between this point and the end of the bucket

This might not be the desired logic for the long term. I suggest we stick to something simple for the next milestone.

tsegismont:
@mtho11 It's almost there. Here's what's missing:
* connection between the REST handler and the new method in MetricsService
* an integration test

jsanda:
Hey @tsegismont, I have looked through a good bit of the PR so far. Everything looks good. I really like AvailabilityBucketedOutputMapperTest, despite being JUnit ;-) This is a good example of executable documentation. There is a small concern I have around the bucketing code. The changes in and of themselves make perfect sense. The bucketing is a way to do downsampling. In other words, we need to squeeze 1,000 data points into 100 data points. For this API addition, we just need to perform some aggregations. In fact, I want to say that actual work that needs to be performed is in the AvailabilityBucketedOutputMapper.newPointInstance method. We definitely need downsampling, but I am a bit concerned that downsampling when we just need one or more aggregation functions may be a bit heavyweight. For example, suppose we want to add a custom aggregation function that just reports the count of reported availabilities over a specified date range. We don't need buckets. We just need to apply the function. Now when we do downsampling as with the bucketing, we are going to want to apply one or more aggregation functions.

tsegismont:
I understand your point. The truth is I have ignored the problem on purpose to focus on the immediate needs of milestone 2. It's definitely something we'll have to come back to.


-----Original Message-----

tsegismont:
I've removed the input data sort and now the mapper only use a "reverse" view. Order is checked while iterating.

jsanda:
Just to be clear about my downsampling vs aggregation comments, this was not an existing problem as we had not been doing anything with availability except returning the raw data points prior to this work. My point was that I thin we only need the work done in the AvailabilityBucketedOutputMapper.newPointInstance method, which performs the aggregations. I realize that this is time sensitive; so, I am fine with revisiting this. The bucketing code that previously had only been used with numeric data needs some work. I am not entirely familiar with it and just have not made a point to spend more time on it. It is definitely an area to discuss/revisit.

Are you going to add REST API tests?

tsegismont:
Yes, just pushed the ITest. The PR is ready for final review/merge.

PR #180, opened 2015-03-18
HWKMETRICS-36

jsanda:


stefannegrea:
Thank you for the PR!

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -46,6 +47,16 @@
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 
+import com.google.common.base.Function;
```
tsegismont: It's been agreed to follow an import order. Can you update your IDE settings and the PR to get rid of the import order changes? Thanks

stefannegrea:
@tsegismont sorry but I did not see your comment prior to merge.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -444,6 +445,66 @@ public void findNumericData(
     }
 
     @GET
+    @Path("/{tenantId}/metrics/numeric/{id}/periods")
+    @ApiOperation(value = "Retrieve periods for which the condition holds true for each consecutive data point.",
+        response = List.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "Successfully fetched periods."),
+        @ApiResponse(code = 204, message = "No numeric data was found."),
+        @ApiResponse(code = 400, message = "Missing or invalid payload")})
```
tsegismont: There's no payload in a GET request

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
```diff
@@ -800,6 +801,40 @@ private void dropKeyspace(String keyspace) {
         });
     }
 
+    @Override
```
tsegismont: There might be an issue regarding the result when no data is found. In other places, null is returned instead of an empty list. Then in the JAX-RS module SimpleDataCallback checks for null and eventually returns a 204 response.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CassandraBackendITest.groovy
```diff
@@ -15,17 +15,14 @@
...
-import static org.junit.Assert.assertNotNull
-import static org.junit.Assert.assertTrue
-
 import org.hawkular.metrics.core.impl.DateTimeService
 import org.joda.time.DateTime
 import org.junit.Test
 
+import static org.joda.time.DateTime.now
+import static org.joda.time.Seconds.seconds
+import static org.junit.Assert.*
```
tsegismont: Groovy imports should be like Java imports, no imports with '*'

rest-tests/src/test/groovy/org/hawkular/metrics/rest/RESTTest.groovy
```diff
@@ -56,9 +56,9 @@ class RESTTest {
     }
   }
```
tsegismont: What happened here? :smile:

tsegismont:
Le 18/03/2015 17:21, Stefan Negrea a écrit :
> @tsegismont <https://github.com/tsegismont> sorry but I did not see your
> comment prior to merge.

No biggie, as long as they're answered and issues fixed in another PR ;)

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
jsanda: That's a good point. For now, what about updating SimpleDataCallback to return a 204 for either null or an empty collection. I say **for now** because I don't like returning nulls, and it has to do with how some of the queries and mapper classes are implemented. NumericMetricMapper for example returns null when there is no data. It cannot return the NumericMetric object having no data because it doesn't have the tenant id or the metric id. But we already know those. I think this should be addressed separately though.

tsegismont: Le 18/03/2015 17:40, jsanda a écrit :
> That's a good point. For now, what about updating SimpleDataCallback to
> return a 204 for either null or an empty collection. I say *for now*
> because I don't like returning nulls, and it has to do with how some of
> the queries and mapper classes are implemented. NumericMetricMapper for
> example returns null when there is no data. It cannot return the
> NumericMetric object having no data because it doesn't have the tenant
> id or the metric id. But we already know those. I think this should be
> addressed separately though.

I'm fine with a SimpleDataCallback change if we come back to this later. 
Can you open a JIRA? We should handle empty results consistently in the 
long run.
I also prefer to return emptyList(), where it makes sense.

jsanda:
Do we really need a PR for a small code formatting change?

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
jsanda: I will open a separate PR for the change to SimpleDataCallback and create the JIRA ticket.

tsegismont:
Le 18/03/2015 17:43, jsanda a écrit :
> Do we really need a PR for a small code formatting change?

I guess it's too late.

PR #181, opened 2015-03-18
HWKMETRICS-40

Jiri-Kremser:
Don't create the generated/swagger-ui directory in the project root when building with docgen profile


burmanm:
Prr, I recommend PR #182 instead, since build.directory notation is deprecated. Otherwise the same change (I didn't notice this before I submitted mine).

stefannegrea:
Per @burmanm #182 supersedes this PR.

PR #188, opened 2015-03-25
Do not return null from core API methods

jsanda:
This PR cleans up several of the core API methods so that we no longer return null from any methods. When a single object is to be returned, it is wrapped in an Optional. When there are multiple objects to be returned, they are wrapped in either a Collection or in a Map. This makes the semantics for _no data_ in the REST API more clear. Simply put, for a single value there is no data when the Optional is empty, an for multiple values there is no data when the collection or map is empty.

This PR also contains some refactoring for response handling. I have removed the use of SimpleDataCallback in most places in favor of some helper functions. Initially I added the necessary logic to SimpleDataCallback, which made the code a bit complex.


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/ApiUtils.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.common.base.Throwables;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import org.hawkular.metrics.api.jaxrs.ApiError;
+
+/**
+ * @author jsanda
+ */
+@Dependent
```
tsegismont: Is there a reason to use the `@Dependent` scope? Can't an instance of this class be shared?

tsegismont:
There is still a problem with your IDE import settings

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CassandraBackendITest.groovy
```diff
@@ -34,6 +34,66 @@ class CassandraBackendITest extends RESTTest {
   }
 
   @Test
+  void findMetricsWhenThereIsNoData() {
```
tsegismont: Nice to have tests for theses cases!

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/NumericDataMapper.java
```diff
@@ -70,28 +67,32 @@ public NumericDataMapper(boolean includeWriteTime) {
...
-        metric.addData(rowConverter.getData(firstRow));
-
-        for (Row row : resultSet) {
-            metric.addData(rowConverter.getData(row));
-        }
-
-        return metric.getData();
+        List<NumericData> data = new ArrayList<>();
+
+        return StreamSupport.stream(resultSet.spliterator(), false).map(row ->
```
tsegismont: IMO, `map(NumericDataMapper::rowToMetric)` would be more readable.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/NumericDataMapper.java
```diff
@@ -70,28 +67,32 @@ public NumericDataMapper(boolean includeWriteTime) {
...
 
-    private NumericMetric getMetric(Row row) {
-        NumericMetric metric = new NumericMetric(row.getString(ColumnIndex.TENANT_ID.ordinal()), getId(row),
-                row.getMap(ColumnIndex.METRIC_TAGS.ordinal(), String.class, String.class),
-                row.getInt(ColumnIndex.DATA_RETENTION.ordinal()));
-        metric.setDpart(row.getLong(ColumnIndex.DPART.ordinal()));
-
-        return metric;
-    }
+//    private NumericMetric getMetric(Row row) {
```
tsegismont: No mercy: rename to `rowToMetric` and reuse in `apply`... or remove commented code

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
```diff
@@ -95,13 +93,7 @@
...
     public static final int DEFAULT_TTL = Duration.standardDays(7).toStandardSeconds().getSeconds();
 
-    private static final Function<List<ResultSet>, Void> RESULT_SETS_TO_VOID = resultSets -> null;
-
-    private static final NumericMetricMapper NUMERIC_METRIC_MAPPER = new NumericMetricMapper();
-
-    private static final AvailabilityMetricMapper AVAILABILITY_METRIC_MAPPER = new AvailabilityMetricMapper();
-
-    private static final TenantMapper TENANT_MAPPER = new TenantMapper();
+    public static final TenantMapper TENANT_MAPPER = new TenantMapper();
```
tsegismont: Why keep TENANT_MAPPER here if all mappers move to `Functions`?

tsegismont:
Other than that, the changes look good to me. In the future, we might try to avoid lambdas of more than 2-3 lines, and prefer method references. Otherwise, stack traces are impossible to read.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/ApiUtils.java
jsanda: I had to use Dependent scope since the class has public fields. Initially I tried making it application scope but that caused the CDI subsystem to fail to initialize.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CassandraBackendITest.groovy
jsanda: I spent a whole lot of time my import settings last Friday. The settings for static imports were not configured in the build tools repo. I think getting the settings aligned is still a bit a work in progress.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/ApiUtils.java
tsegismont: Le 25/03/2015 13:04, jsanda a écrit :
> I had to use Dependent scope since the class has public fields.
> Initially I tried making it application scope but that caused the CDI
> subsystem to fail to initialize.

Why not turn the final fields into constants then? That's the intention 
anyway, right? Or just keep the class a utility class.

jsanda: They are constants - they are marked final. Initially I had them as static constants in MetricHandler. Since they are more general purpose, I decided it would be better for them to be a separate class. They are also used in TenantHandler and in InfluxSeriesHandler. I just tested with making ApiUtils application scoped and making the fields/methods static. No CDI start up errors, so I will move ahead with that change.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CassandraBackendITest.groovy
tsegismont: Le 25/03/2015 13:05, jsanda a écrit :
> I spent a whole lot of time my import settings last Friday. The settings
> for static imports were not configured in the build tools repo. I think
> getting the settings aligned is still a bit a work in progress.

The settings JAR is a work in a progress, the convention isn't :)

static java.*
blank line
static javax.*
blank line
static javax.*
blank line
static javax.*
blank line
static javax.*

tsegismont:
I don't really get it. What's the purpose of APIUtil as a CDI bean if all members (fields and methods) are static? Can you add a private constructor and remove the `@ApplicationScoped`
To be honest I don't really a value in turning a utility class in something injectable here.

jsanda:
I made it a managed bean/class because I was anticipated the need for having a thread pool available for submitting tasks to avoid doing work on driver I/O threads. The executeAsync method is the only one that falls into this category. I can update that method to accept an Executor argument, and make the class a true static utility class.

jsanda:
Actually, I do not think there is a need for an Executor argument. As long as the async work done by the core API is not performed on an I/O thread then it should be fine.

tsegismont:
@jsanda Travis build failed because of a CDI issue. I didn't investigate.

tsegismont:
Thanks @jsanda

PR #189, opened 2015-03-25
HWKMETRICS-26 REST-API uses return codes incorrectly

tsegismont:
Please don't merge. I'm waiting for #188 to be merged. Then I'll rebase/adjust this one.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
@@ -21,6 +21,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
+import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.badRequest;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.executeAsync;
 import static org.hawkular.metrics.core.api.MetricsService.DEFAULT_TENANT_ID;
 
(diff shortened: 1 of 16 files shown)


tsegismont:
The PR is ready for review and merge.

The changes could have been limited to a few tags related methods in `MetricHandler`. But while I was adding Swagger docs and code for checking invalid tag list all over the place, I realized that converting tags the way we did was wrong. So I've created a new JAX-RS param object, `Tags`. This leads to much cleaner code for handling invalid format.

Besides, I wanted the generated REST doc to look better, so I did some research on how to customize the documentation of path and query parameters when their type is a custom one. Hence the changes in the Swagger filter object to override the default documentation mechanism for `Tags` and `Duration` param.

As  as side work, I've updated the `Duration` param converting code to make it consistent with the converting code for `Tags`.

tsegismont:
Travis and Jenkins builds failing because of the ongoing attack against GH...

tsegismont:
retest this please

jsanda:
My question about  removing the decodeTags method was not about the customer type converter. Adding the type converter is a great change. My question was the implementation of the TagsConverter.fromString method.

Where is BadRequestExceptionMapper used?

tsegismont:
Le 27/03/2015 15:19, jsanda a écrit :
> My question about removing the decodeTags method was not about the
> customer type converter. Adding the type converter is a great change. My
> question was the implementation of the TagsConverter.fromString method.

Oops, sorry, I didn't get that. I implemented it like this because with 
the Java 8 `Stream` version, the only way to detect invalid format is to 
wrap the code with a try-catch-anyexception block.

>
> Where is BadRequestExceptionMapper used?
>

`BadRequestExceptionMapper` is not explicitly created. It's annotated 
with `@Provider`. It's not a new class, I've added it in the past weeks.

A benefit of this provider is that we can customize (with `ApiError`) 
any issue which would arise in the process of matching URI+METHOD with 
handler methods.

tsegismont:
Green builds. But @jshaughn said that we should hold off until the alpha 2 work is done.

PR #190, opened 2015-03-26
One mustache to rule them all.

Jiri-Kremser:
Using the remotely sourced mustache file to standardize the format of
generated docs across all the components.


rhqci:

tsegismont:
Thanks @Jiri-Kremser

api/metrics-api-jaxrs/pom.xml
```diff
@@ -177,14 +176,13 @@
...
             <artifactId>swagger-maven-plugin</artifactId>
-            <version>2.3.3</version>
             <configuration>
               <apiSources>
                 <apiSource>
                   <locations>org.hawkular.metrics.api.jaxrs</locations>
                   <apiVersion>1.0</apiVersion>
                   <basePath>http://localhost:8080/hawkular-metrics/</basePath>
-                  <outputTemplate>${basedir}/src/main/resources/rest-doc/asciidoc.mustache</outputTemplate>
+                  <outputTemplate>https://raw.githubusercontent.com/hawkular/hawkular.github.io/swagger/asciidoc.mustache</outputTemplate>
```
pilhuhn: Is that cached? Or how would I build the docs offline?

Jiri-Kremser: > Is that cached? Or how would I build the docs offline?

No cache, normal http get is performed. So when building with `-Pdocgen` in train it'd fail the build. Workaround is to download the file and enter the local path to it. 

I could have added another profile for that called `docgenOffline` with a link to the local file (per component), but it would defeat the purpose of the change -> having the standardized form of documentation.

Jiri-Kremser: btw. it fails on DNS resolving java.net.UnknownHostException, perhaps if you put `raw.githubusercontent.com` to your local hosts file it could respect the 'caching headers'. Everything on 'http://raw.*' has the max-age set only to 300seconds though. The 'caching headers' could be prolonged by using the `rawgit.com` service instead. 

imho it doesn't worth doing

tsegismont: Le 27/03/2015 09:42, Heiko W. Rupp a écrit :
> Is that cached? Or how would I build the docs offline?
>

Good catch. How about moving the mustache file to the build tools module 
and unzip the file before generating the doc.

Everyone will have the build tools module locally because it's required 
for building the project.

pilhuhn: On 27 Mar 2015, at 12:20, Thomas Segismont wrote:
> module
> and unzip the file before generating the doc.
>
> Everyone will have the build tools module locally because it's required
> for building the project.

Sounds good to me.

PR #191, opened 2015-03-30
REST service generator based on annotation processing

tsegismont:
**Not ready yet**

The generator bridges the gap between the JAX-RS AsyncResponse model and the C* driver ListenableFuture model.

It frees us from copy/pasting #executeAsync calls and avoid writing possibly long, nested lambdas (not super easy to read, not debugger-friendly, not checkstyle-friendly).

It's a first shot so it still needs some polishing (no Javadoc, more code to factor out).

It should work fine with the docgen profile, but needs to be checked.


tsegismont:
I've check the docgen, it works fine.

jsanda:
I have some reservations about this PR. First, we are not copying/pasting calls of executeAsync. We call executeAsync in several method using different arguments. That is code reuse, not copy/paste. 

The long, nested lambdas to which you refer are simply the same method bodies prior to introducing the executeAsync method and are not wrapped in a function. The caller of that function, executeAsync, handles the futures callback code, thus eliminating what was previously some boiler plate code. 

I also disagree that the code is more difficult to read. I am not saying it is easier, but I do not think that wrapping those method bodies in lambda functions make it less readable. I suppose it is a matter of taste/preference. 

Can you elaborate on the debugger-friendly aspect? I have no problem setting breakpoints in or stepping through a lambda function.

It is not checkstyle-friendly because it is new. I would rather put it that checkstyle is not lamba-friendly :)

Lastly, I am -1 on code generation.

tsegismont:
More on "bridges the gap between the JAX-RS AsyncResponse model and the C* driver ListenableFuture model."

The tool processes interfaces (`XXXService`) annotated with JAX-RS `@Path` and generates an abstract implementation (`XXXServiceBase`). This base class implements the JAX-RS service interface.

Let's have a look at the `TenantSevice`. It has all the JAX-RS and Swagger annotations. Each method must have one parameter of type `AsyncResponse`:
```java
@Path("/tenants")
@Consumes(APPLICATION_JSON)
@Produces(APPLICATION_JSON)
@Api(value = "/tenants", description = "Tenants related REST interface")
public interface TenantsService {

    @POST
    @ApiOperation(value = "Create a new tenant. ", notes = "Clients are not required to create explicitly create a "
                                                           + "tenant before starting to store metric data. It is "
                                                           + "recommended to do so however to ensure that there "
                                                           + "are no tenant id naming collisions and to provide "
                                                           + "default data retention settings. ")
    @ApiResponses(value = {
            @ApiResponse(code = 201, message = "Tenant has been succesfully created."),
            @ApiResponse(code = 400, message = "Missing or invalid retention properties. ",
                         response = ApiError.class),
            @ApiResponse(code = 409, message = "Given tenant id has already been created.",
                         response = ApiError.class),
            @ApiResponse(code = 500, message = "An unexpected error occured while trying to create a tenant.",
                         response = ApiError.class)
    })
    void createTenant(
            @Suspended AsyncResponse asyncResponse, @ApiParam(required = true) Tenant params,
            @Context UriInfo uriInfo
    );
```

The generated class implements the service interface and has an abstract companion method:
```java
/*
 * Generated class file
 */
public abstract class TenantsServiceBase implements org.hawkular.metrics.api.jaxrs.service.TenantsService {

    @Override
    public void createTenant(
        javax.ws.rs.container.AsyncResponse asyncResponse,
        org.hawkular.metrics.core.api.Tenant params,
        javax.ws.rs.core.UriInfo uriInfo
    ) {
        ListenableFuture<Response> future = _createTenant(
            params,
            uriInfo
        );
        addCallback(asyncResponse, future);
    }

    protected abstract ListenableFuture<Response> _createTenant(
        org.hawkular.metrics.core.api.Tenant params,
        javax.ws.rs.core.UriInfo uriInfo
    );
 
    private static void addCallback(AsyncResponse asyncResponse, ListenableFuture<Response> future) {
        Futures.addCallback(
                future, new FutureCallback<Response>() {
                    @Override
                    public void onSuccess(Response response) {
                        asyncResponse.resume(response);
                    }

                    @Override
                    public void onFailure(Throwable t) {
                        String exceptionMsg = Throwables.getRootCause(t).getMessage();
                        String msg = "Failed to perform operation due to an error: " + exceptionMsg;
                        asyncResponse.resume(Response.serverError().entity(new ApiError(msg)).build());
                    }
                }
        );
    }
```

The companion method has all the original service method arguments, except the `AsyncResponse` one. It returns an `ListenableFuture<Response>`. The generated code takes care of transforming the future into something which will eventually call `AsyncResponse#resume`.

So now implementing the service (`XXXHandler`) boils down to implementing the companion method:
```java
@ApplicationScoped
public class TenantsHandler extends TenantsServiceBase {

    // TODO: add back retention settings

    @Inject

    @Override
    protected ListenableFuture<Response> _createTenant(Tenant params, UriInfo uriInfo) {
        if (params == null) {
            return emptyPayload();
        }
        ListenableFuture<Void> insertFuture = metricsService.createTenant(params);
        URI location = uriInfo.getBaseUriBuilder().path("/tenants").build();
        ListenableFuture<Response> responseFuture = Futures.transform(insertFuture, created(location));
        FutureFallback<Response> fallback = alreadyExistsFallback(
                TenantAlreadyExistsException.class,
                TenantsHandler::alreadyExistsError
        );
        return Futures.withFallback(responseFuture, fallback);
    }

    private static ApiError alreadyExistsError(TenantAlreadyExistsException e) {
        return new ApiError("Tenant with id " + e.getTenantId() + " already exists");
    }
```

As mentioned before the advantages are that the processor frees us from copy/pasting #executeAsync calls and avoid writing possibly long, nested lambdas.

Besides the processor enforces a single pattern for service implementation: **return a ListenableFuture<Response>**. No matter if your code is checking arguments, or transforming results from async calls to the C* driver, there's one thing to do, return a response future.

And `ResponseUtils` has a bunch of handy methods to help building the response future.

By the way, the JAX-RS and Swagger annotations are held by the service interface, so it makes the service implementation easier to read (no annotations explosion).

stefannegrea:
I second @jsanda , this seems to complicate the simple prescriptive service code. I see little benefit from doing this given that it adds one more level of indirection. 

If we were to generate thousands of these methods, I could see the benefit. Given the small number of end-points (combined with a disciplined approach on organizing them) I will think it will be simpler to keep the current implementation. The PR will make simple code really confusing.

tsegismont:
I was writing my previous comment while you sent yours. My answers inline.

Le 31/03/2015 17:24, jsanda a écrit :
> I have some reservations about this PR. First, we are not
> copying/pasting calls of executeAsync. We call executeAsync in several
> method using different arguments. That is code reuse, not copy/paste.

You're right. I was referring to the way I'm working with it: copy past 
the block from another method, then edit the lamdba.

>
> The long, nested lambdas to which you refer are simply the same method
> bodies prior to introducing the executeAsync method and are not wrapped
> in a function. The caller of that function, executeAsync, handles the
> futures callback code, thus eliminating what was previously some boiler
> plate code.

Not exactly. The #executeAsync construct has changed the way the service 
method is implemented: the lambda passed as argument must return a 
Listenable<Future>. Before, service method had to deal with FutureCallbacks

Note that I started this PR because I think #executeAsync was a good 
step forward. The difference with the annotation processor is that the 
lambda passed as argument to #executeAsync is now the actual handler method.

>
> I also disagree that the code is more difficult to read. I am not saying
> it is easier, but I do not think that wrapping those method bodies in
> lambda functions make it less readable. I suppose it is a matter of
> taste/preference.
>

The nested lambda blocks are at least 8 column indented. Given that we 
work with ListenableFuture<List<AvaiabilityMetric> instances, it leaves 
just a few column for actual code.

> Can you elaborate on the debugger-friendly aspect? I have no problem
> setting breakpoints in or stepping through a lambda function.

Debugger-friendly was a mistake, I meant stack trace friendly. When your 
lambda is 1,2,3 lines it's fine, it does a few things. Having lambdas of 
30 lines is a mistake IMO.

>
> It is not checkstyle-friendly because it is new. I would rather put it
> that checkstyle is not lamba-friendly :)
>

I meant that the service handler code with #executeAsync forces you to 
loose a bunch of columns before you reach the margin.


> Lastly, I am -1 on code generation.

Why? Just on principles?

Generating boiler plate code with an annotation processor is very useful.

jsanda:
>The companion method has all the original service method arguments, except the AsyncResponse one. It returns an ListenableFuture<Response>. 

The Supplier function does exactly that. It returns a ListenableFuture<Response>.

>The generated code takes care of transforming the future >into something which will eventually call AsyncResponse#resume.

The executeAsync method does this transformation.

Maybe I misunderstand something, but I fail to see the benefits here. 

With respect to the code generation, I only see it adding boiler plate, not reducing it. Now instead of passing a function to executeAsync, we write an interface, generate an implementation of said interface, and implement a companion method which is essentially the equivalent of what is passed to executeAsync.

tsegismont:
Le 31/03/2015 17:56, jsanda a écrit :
>     The companion method has all the original service method arguments,
>     except the AsyncResponse one. It returns an ListenableFuture.
>
> The Supplier function does exactly that. It returns a ListenableFuture.

Yes. The principle is the same. But instead of writing nested code, you 
implement a method of a generated abstract class.

>
>     The generated code takes care of transforming the future >into
>     something which will eventually call AsyncResponse#resume.
>
> The executeAsync method does this transformation.

See above. The PR is derived from the idea of #executeAsync

>
> Maybe I misunderstand something, but I fail to see the benefits here.
>
> With respect to the code generation, I only see it adding boiler plate,
> not reducing it. Now instead of passing a function to executeAsync, we

Your're not adding boiler plate when boiler plate is **generated**.

> write an interface, generate an implementation of said interface, and

Generation is automated (integrated to the build with an annotation 
processor).

> implement a companion method which is essentially the equivalent of what
> is passed to executeAsync.

Yes it's equivalent.

jsanda:
@tsegismont your last comment confirms my understanding. This PR involves generating an interface with a stub implementation that has an abstract method we implement. We already accomplish this by passing a java.util.function.Supplier<ListenableFuture<Response>> to executeAsync. 

I see no reason for the code generation. It simply adds boiler plate code that we do not need. If you prefer to *not* write lambda functions, then write an anonymous inner class or a top-level class that implements the function interface.

Unless there is something I missed or misunderstand, I prefer not to spend more time on this. There are other more important, higher priority things on which I want to focus.

tsegismont:
OK

tsegismont:
Closed rejected

PR #193, opened 2015-03-31
[HWKMETRICS-13]Adding InfluxDB java client support test

jkandasa:


rhqci:

tsegismont:
I'll have a look tomorrow morning. Thanks


-----Original Message-----

jkandasa:
@tsegismont  Hi Thomas, There are two tests failed due to incorrect value from hawkular-metrics server.
1.Test Sum
2.Test Mean
Input:10.9,11.9,12.9,13.9,14.9,15.9,16.9,17.9,18.9
Expected Sum:134.1
Expected Mean:14.9
Result from hawkular-metrics:
Sum:134.10000000000002
Mean:14.900000000000002

tsegismont:
You can't use assertEquals with double values. IIRC, junit has an assertDoubleEquals method which takes an extra delta argument.


-----Original Message-----

rest-tests/src/test/java/org/hawkular/metrics/test/InfluxDatabaseITest.java
```diff
@@ -0,0 +1,231 @@
...
+        Serie serieSource = new Serie.Builder(timeSeries)
+        .columns("time", "sum")
+        .values(new Double(startTime+8000L), 134.1D)
+        .build();
+        List<Serie> series = influxDB.query(dbName,
+                                            "select sum(value) from "+timeSeries
+                                            +" where time > now() - 30s group by time(30s)",
+                                            TimeUnit.MILLISECONDS);
+        Assert.assertNotNull(series);
+        Assert.assertEquals(serieSource.toString(), series.get(0).toString());
```
tsegismont: Ok now I see the issue. You can't compare the two series as string. You may create an `#assertSeriesEquals` method, in this class or its parent. In this method you could iterate over the points and check that the timestamps and values are "equal" within a delta.

See http://junit.org/javadoc/latest/org/junit/Assert.html#assertEquals%28java.lang.String,%20double,%20double,%20double%29

tsegismont:
This is a very good contribution @jkandasa , thanks. Waiting for the test failure fix before merging.

jkandasa:
@tsegismont Thank you Thomas. I have changed 'assertDoubleEquals' for double values with delta as '1e-13' for double and '1e-7' for float

tsegismont:
@jkandasa perfect! Merging now.

pilhuhn:
(Just a more general thought, not related to this PR, but triggered by it)  
This double compare failure (and other stuff) really should make us think if we really want to treat every data value as double instead of integral types for integral values.

PR #196, opened 2015-04-09
[HWKMETRICS-50] return distinct avails as determined by state change

jsanda:


jsanda:
This is not ready to be merged yet. I need to add an unknown availability state.

jsanda:
I have added an unknown availability type. This PR is ready for review.

mtho11:
Once this is merged can it be pushed to kettle as well, so the UI can use it?

stefannegrea:
@mtho11 we need to update kettle to use version 0.3.2-SNAPSHOT of Hawkular Metrics. I will take care of that.

PR #197, opened 2015-04-10
REST API - remake

stefannegrea:


api/metrics-api-jaxrs/README.adoc
```diff
@@ -170,7 +170,7 @@ GET /hawkular-metrics/{tenantId}/metrics
 *Example Request*
 
 ----
-curl http://localhost:8080/hawkular-metrics/{tenantId}/metrics?type=num
+curl http://localhost:8080/hawkular/metrics/{tenantId}/metrics?type=guage
```
burmanm: Weren't we supposed to leave these behind and use consistent naming everywhere? avail vs. availability and ?type parameter instead of /resource?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AvailabilityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+
+    @GET
+    @Path("/{tenantId}/availability/{id}/tags")
+    @ApiOperation(value = "Retrieve tags associated with the metric definition.", response = Map.class)
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Metric's tags were successfully retrieved."),
+            @ApiResponse(code = 204, message = "Query was successful, but no metrics were found."),
+            @ApiResponse(code = 500, message = "Unexpected error occurred while fetching metric's tags.",
+                response = ApiError.class) })
+    public void getAvailabilityMetricTags(@Suspended final AsyncResponse asyncResponse,
```
burmanm: This returns the Metric definition if tags are available, while it should probably just return the tags and not the rest of the definition (especially since it doesn't return the definition if there are no tags). See Google Docs comment on B10. Same applies to the getGaugeMetricTags

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AvailabilityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+            @Suspended AsyncResponse asyncResponse,
+            @PathParam("tenantId") String tenantId,
+            @PathParam("id") String id,
+            @ApiParam(value = "Defaults to now - 8 hours") @QueryParam("start") final Long start,
+            @ApiParam(value = "Defaults to now") @QueryParam("end") final Long end,
+            @ApiParam(value = "Total number of buckets") @QueryParam("buckets") Integer bucketsCount,
+            @ApiParam(value = "Bucket duration") @QueryParam("bucketDuration") Duration bucketDuration,
+            @ApiParam(value = "Set to true to return only distinct, contiguous values")
+                @QueryParam("distinct") Boolean distinct) {
+        executeAsync(
```
burmanm: This is a very long anonymous function, could it be refactored to something more readable (outside the method) ? Same applies to the Gauge version.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/CounterHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * @author Stefan Negrea
+ *
+ */
+@Path("/")
+@Consumes(APPLICATION_JSON)
+@Produces(APPLICATION_JSON)
+@Api(value = "", description = "Availability metrics interface")
+public class CounterHandler {
```
burmanm: Return codes are not mentioned for documentation purposes. Also, are these counters still hidden=true (they were hidden in the older versions because they weren't ready for use) ?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/GuageHandler.java
```diff
@@ -0,0 +1,407 @@
...
+            @PathParam("id") String id,
+            @ApiParam(value = "Defaults to now - 8 hours", required = false) @QueryParam("start") final Long start,
+            @ApiParam(value = "Defaults to now", required = false) @QueryParam("end") final Long end,
+            @ApiParam(value = "A threshold against which values are compared", required = true)
+                @QueryParam("threshold") double threshold,
+            @ApiParam(value = "A comparison operation to perform between values and the threshold."
+                    + " Supported operations include ge, gte, lt, lte, and eq", required = true)
+                @QueryParam("op") String operator) {
+
+        executeAsync(
```
burmanm: Very long anonymous function here also, and with some business logic also instead of only interface-related stuff.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AvailabilityHandler.java
jsanda: They are long functions; however, I tend to think it is somewhat outside the scope of this PR because that is how the functions already exist in master. I agree that they should be refactored, but I think it should be a separate PR.

burmanm: Yes, separate PR is fine, just wanted to log it here when the PR is about REST-API remake.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AvailabilityHandler.java
jsanda: There are at least a few things that fall into the category of metric meta data.

* tags
* data retention settings
* aggregation settings

Having separate endpoints that get only tags, only data retention settings, and only aggregation settings seems like overkill to me. I think something like either `GET /{tenantId}/availability/{id}/meta` or `GET /{tenantId}/availability/{id}` makes sense. I do think though that there should be different endpoints for updating each of the different meta data. Updating data retention will be an expensive, background operation. Updating aggregation settings might be as well. Updating tags is a simple map update.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/CounterHandler.java
jsanda: I think this class is just a place holder for now until we get an updated counters core Java API.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/GaugeHandler.java
```diff
@@ -0,0 +1,407 @@
...
+@Produces(APPLICATION_JSON)
+@Api(value = "", description = "Gauge metrics interface")
+public class GaugeHandler {
+    private static final long EIGHT_HOURS = MILLISECONDS.convert(8, HOURS);
+
+    @Inject
+    private MetricsService metricsService;
+
+    @POST
+    @Path("/{tenantId}/gauge")
```
jsanda: Shouldn't this be gauge**s** plural - `/{tenantId}/gauges`

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -847,7 +75,7 @@ public void onFailure(Throwable t) {
     public void findMetrics(
             @Suspended final AsyncResponse asyncResponse,
             @PathParam("tenantId") final String tenantId,
-        @ApiParam(value = "Queried metric type", required = true, allowableValues = "[num, avail, log]")
+        @ApiParam(value = "Queried metric type", required = true, allowableValues = "[gauge, avail, log]")
```
jsanda: For consistency I think the `type` parameter should accept the values,

* availability
* gauges
* counters

Since we have not implemented anything for events/logs, I think `log` should be removed.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
```diff
@@ -89,752 +58,11 @@
...
-            @Override
-            public void onFailure(Throwable t) {
-                asyncResponse.resume(t);
-            }
-        });
-    }
-
-    @GET
-    @Path("/{tenantId}/metrics")
+    @Path("/{tenantId}")
```
jsanda: I think that the previous endpoint, `GET /{tenantId}/metrics`, makes more sense because there might be other information we want to include under `/{tenantId}`. There are tenant-level data retention and aggregation settings for example. I would not be surprised if introduce additional tenant-level configuration settings.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/request/MixedMetricsRequest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.hawkular.metrics.core.api.Gauge;
+
+import com.google.common.base.Objects;
+import com.wordnik.swagger.annotations.ApiModel;
+
+/**
+ * @author Stefan Negrea
+ */
+@ApiModel
+public class MixedMetricsRequest {
```
jsanda: I am -1 on this class being in core-impl. It is specific to the REST API.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AvailabilityHandler.java
burmanm: This is fine solution also, as long as it is possible to request metadata somehow (which it currently isn't). The latter url sounds better to me (unless we want to provide datapoints from there).

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
burmanm: Is there a reason why this is QueryParam instead of PathParam? It would make more sense as /{tenantId}/gauges or the proposed /{tenantId}/metrics and the type would only filter (by default send everything) ?

clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
```diff
@@ -95,7 +94,7 @@
     @Before
     public void setUp() throws Exception {
         tenant = getRandomTenantId();
-        findNumericMetricsUrl = "http://" + BASE_URI + "/" + tenant + "/metrics?type=num";
+        findGuageMetricsUrl = "http://" + BASE_URI + "/" + tenant + "/metrics?type=guage";
```
tsegismont: Typo "guage" (is this the cause of Travis failures?) Note that there are other places in this file and in other files with the same typo.

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
```diff
@@ -40,7 +40,7 @@ public void closeResponse() {
     @Test
     public void testNotAllowedException() {
         response = target.clone()
-                .path("/test/metrics/numeric/test/tags")
+.path("/test/gauge/test/tags")
```
tsegismont: Can you restore indentation?

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
```diff
@@ -52,21 +52,21 @@ public void testNotAllowedException() {
     @Test
     public void testNotFoundException() {
         response = target.clone()
-                .path("/test/metricsssss/numeric/test/data")
+.path("/test/gaugesssss/test/data")
```
tsegismont: Same here

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
```diff
@@ -52,21 +52,21 @@ public void testNotAllowedException() {
...
-                + "/test/metricsssss/numeric/test/data?buckets=999",
+                + "/test/gaugesssss/test/data?buckets=999",
                 apiErrorJson.getErrorMsg());
     }
 
     @Test
     public void testNumberFormatException() {
         response = target.clone()
-                .path("/test/metrics/numeric/test/data")
+.path("/test/gauge/test/data")
```
tsegismont: Same here

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
```diff
@@ -79,7 +79,7 @@ public void testNumberFormatException() {
     @Test
     public void testNotAcceptableException() {
         response = target.clone()
-                .path("/test/metrics/numeric/test/data")
+.path("/test/gauge/test/data")
```
tsegismont: Same here

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
```diff
@@ -91,7 +91,7 @@ public void testNotAcceptableException() {
     @Test
     public void testNotSupportedException() {
         response = target.clone()
-                .path("/test/metrics/numeric/test/data")
+.path("/test/gauge/test/data")
```
tsegismont: Same here

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
tsegismont: +1

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
tsegismont: I'd prefer we simply get rid the query param option:
/{tenantId}/metrics -> returns all metrics, regardless of the type
/{tenantId}/gauges -> returns all gauges
/{tenantId}/counters -> returns all counters
/{tenantId}/avail -> returns all avail

burmanm: Except that not avail, instead availability. Lets keep them consistent.

rest-tests/src/test/java/org/hawkular/metrics/test/ErrorsITest.java
stefannegrea: Fixed all, I blame Eclipse! Thanks for the feedback.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/GaugeHandler.java
stefannegrea: The plural is extraneous, just an extra s. I think it is a fair assumption that a GET on /gauge/ will return a list of gauge metrics.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
stefannegrea: Agree with the comments. Updated the url to have the singular word "metric"  the url is /{tenantId}/metric to be consistent with gauge, counter, and availability.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/request/MixedMetricsRequest.java
stefannegrea: This is not the only stray class in that package, there is at least another request class. I would like to defer moving this into the JAXRS API package until the core and impl are refactored.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/GaugeHandler.java
jsanda: I disagree :) I do not think it is explicitly clear since gauge is singular and also think about other types like counters or histograms (when we introduce them). Are we going to have `/{tenantId}/counter` or `/{tenantId}/histogram` as opposed to  `/{tenantId}/counters` and `/{tenantId}/histograms`? I hope not .

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
tsegismont: Right

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/GaugeHandler.java
stefannegrea: I agree with your last comments. Will update everything to plural.

stefannegrea: But what should we do with availability? Should that be moved to availabilities?

jshaughn: I like Gauge, although gauges is OK.  I would not go with Availabilities, just Availability.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricHandler.java
mtho11: From a UI perspective (any client really), I just want to add parameters to get the things I need and not have to switch urls (and maintain). 
So all I want is:
 /{tenantId}/metrics -> returns all metrics, regardless of the type
with http param filters.

PR #199, opened 2015-04-12
Replace tagging async transformers with a single function class

burmanm:
Replaces duplicate code with a single reusable instance for both metricData types.


rhqci:

rhqci:

PR #201, opened 2015-04-21
Fix build so that we use the provided Dockerfile and not a generated one

mwringe:


rhqci:

rhqci:

PR #202, opened 2015-04-23
Move tenant id to request header param

stefannegrea:


pilhuhn:
As already discussed, I think this may be problematic for the Grafana integration via Influx adapter.
Can we evaluate first how to keep Grafana (and other simple clients) going?

tsegismont:
The Influx endpoint is not changed and still listens on `/db/{tenantId}/series`.

What I don't get is the motivation. Back in December 2014 we had a conversation on the mailing list about tenant information in path vs in headers, and it was agreed that putting it in the path was the way to go. What changed since then? Could the motivation be explained on `hawkular-dev` ?

pilhuhn:
On 24 Apr 2015, at 10:08, Thomas Segismont wrote:
> to go. What changed since then? Could the motivation be explained on 
> `hawkular-dev` ?

+1

PR #203, opened 2015-04-24
findMetrics did not follow consistent gauges naming

burmanm:
type should follow the same naming convention than rest of the REST-API (even if plural doesn't sound right). I still would prefer moving this to PathParam ..

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -87,7 +87,8 @@ public class MetricHandler {
                         metricType = MetricType.fromTextCode(type);
                     } catch (IllegalArgumentException e) {
                         return badRequest(
-                                new ApiError("[" + type + "] is not a valid type. Accepted values are gauge|avail|log")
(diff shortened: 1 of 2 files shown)


tsegismont:
>  I still would prefer moving this to PathParam ..

Agreed, this endpoint is more of RPC-style, whereas splitting it into `/{tenantId}/gauges`, ... etc conforms to resource oriented design, which was one of the motivations of the JAX-RS API remake.

Too bad [the comments in the corresponding PR](https://github.com/hawkular/hawkular-metrics/pull/197#discussion_r28882378) were not answered before it was merged.

tsegismont:
I agree with Micke. +1 for merging

jsanda:
I suppose it is not a big deal either way, but the plural makes sense to me and does seem more consistent.

stefannegrea:
We will need to rebase this on the latest master. 

@burmanm, can you please update the MetricType.fromTextCode to support parsing both single and plural for gauge? In my comment I was more concerned about the domain model not supporting singular. I am ok with the REST API using plural.

stefannegrea:
@burmanm the merge commit was really confusing. Do you mind just force rebasing on master. The primary changes are all good, but the diff for the second commit is really confusing.

burmanm:
Will redo later.

stefannegrea:
@burmanm thanks for the work. Sorry it did not get merged in time.

PR #205, opened 2015-05-01
Initial task scheduling support

jsanda:
I feel like this is at a point where we can start integrating. There are still several things to be done, but I think it makes sense to tackle those as we start integrating. There is initial support for retrying failed tasks but no support for rescheduling dependent tasks. There is no support for non-repeating tasks. What is here though is more than enough to get going with implementing pre-computed aggregates.


jsanda:
I forgot another thing that needs to be sorted out, schema installation. For now I have added the new tables to schema.cql in metrics-core-impl. That is not a long term solution because the new task-queue module should not have any dependencies on any other metrics modules. It is or can be generic enough that it be used outside of metrics.

jsanda:
I do not think it belongs under core because it is generic. I think that there is a good possibility that there might be use cases for this outside of metrics in which case it might even be better suited in a different repo altogether.

jsanda:
I can add some javadocs for Lease.java. I hadn't bothered yet since it is an internal, implementation class. The javaocs in TaskType.java do describe segment offsets.

jsanda:
No, this does not cover schema updates from previous installations. We do not yet have schema management tooling in place yet. Merging this to master will not be disruptive in that regard because nothing depends on the new task-queue module yet. It is only used in tests so far.

jsanda:
Good catch. I committed/pushed the change for using UTC.

stefannegrea:
Regarding the schema management comment, does it imply by the time the task queue tables are needed we will have some sort of schema management to upgrade/fix the existing schema smoothly?

jsanda:
No implications :) Just pointing out the current status with respect to schema change management.

jsanda:
I reduced visibility of some classes. Others cannot be reduced because they need to be accessible to TaskServiceBuilder. Hopefully though, the separation of client-facing API vs. internal implementation is pretty clear.

tsegismont:
Le 12/05/2015 23:09, jsanda a écrit :
> I do not think it belongs under core because it is generic. I think that
> there is a good possibility that there might be use cases for this
> outside of metrics in which case it might even be better suited in a
> different repo altogether.

+1, sooner or later another project will look at task scheduling.

tsegismont:
Le 12/05/2015 23:45, jsanda a écrit :
> Good catch. I committed/pushed the change for using UTC.

I've seen the commit you're talking about. Forcing the timezone for the 
whole JVM looks a bit odd to me. It will affect everything.

Besides, as it is implemented, it will happen only after the class is 
loaded, so it may happen that some code (maybe in the server or in any 
library) will call TimeZone#getDefault before the TZ is force to UTC.

I'm -1 on this and I believe we'd better manually set the timezone where 
needed.

While we're at it, is there any problem with storage in C* nodes?

Also, do we expect the clients to use UTC timestamps when sending data 
points? I guess we should and in this case, document it ASAP.

jsanda:
Just a brief follow up/clarification on the UTC change. ```DateTimeZone.setDefault(DateTimeZone.UTC)``` does **not** set the default time zone for the whole JVM. The method does not call java.util.TimeZone, and the JavaDocs point this out. This would apply only to Joda Time and for its class loader I believe.

tsegismont:
Good news, that makes the change much less pervasive. Still, is it possible to make the conversion in the scheduler utility class?
This change means that whenever we'll create an joda object without a timezone specified, it will be in UTC. In the server code, in the test code, or in the overlay(glue) code. That might lead to bad surprises.

jsanda:
I am kind of torn on this. I agree that there could be surprises. In RHQ, the surprises came from using the default time zone. See https://bugzilla.redhat.com/show_bug.cgi?id=1161806. Based on that experience, I am more inclined to think that UTC as a default is the safe way to go in terms of avoiding bad surprises.

jsanda:
For now I am leaving the UTC default. It is less error prone; however, I made the changes so that UTC is local to the task scheduling library. I pushed the changes to https://github.com/jsanda/hawkular-metrics/commit/dbea29a29de90a5ceaff45cbdd32c5e5c24c7193 so that we have them if we decide not to use a global default of UTC.

task-queue/src/main/java/org/hawkular/metrics/tasks/DateTimeService.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.metrics.tasks;
+
+import static org.joda.time.DateTime.now;
```
jpkrohling: Is there a reason to use Joda, instead of the similar new date APIs from Java 8? Are we going to support Java 7 and earlier?

jsanda: Right now we are basically reusing a lot date/time code from RHQ. Provided there is no major loss in functionality, the change makes sense. I just do not see it as a high priority right now.

jpkrohling: The migration from Joda to Java 8 Date API should be quite straightforward, as this new API was based on Joda's. In most cases, it's just a matter of changing the package where the class comes from.

PR #207, opened 2015-05-11
Don't push empty docs

Jiri-Kremser:
Check on the consumer side is made also here https://github.com/hawkular/hawkular.github.io/blob/pages/.travis.swagger.sh#L23

this PR is based on #206

### DIFF
--- a/.travis.swagger.sh
+++ b/.travis.swagger.sh
@@ -17,10 +17,14 @@
 #
 
 # generate the docs
-mvn clean compile -Pdocgen -DskipTests -Dcheckstyle.skip
+mvn clean compile -Pdocgen -DskipTests -Dcheckstyle.skip -Dlicense.skip
 
 FILE_NAME="rest-metrics.adoc"
 FILE_PATH="api/metrics-api-jaxrs/target/generated/$FILE_NAME"
+
+# don't push the empty docs
+[[ -s $FILE_PATH ]] || echo "$FILE_PATH is empty" && exit 1
+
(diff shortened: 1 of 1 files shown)


.travis.swagger.sh
```diff
@@ -17,10 +17,14 @@
 #
 
 # generate the docs
-mvn clean compile -Pdocgen -DskipTests -Dcheckstyle.skip
+mvn clean compile -Pdocgen -DskipTests -Dcheckstyle.skip -Dlicense.skip
```
tsegismont: Why do we need this ?

Jiri-Kremser: This is run in post success 'travis phase' so there is no need for this license check because it was already done before.

So it should speed up the doc generation process.

tsegismont: OK

tsegismont:
How come we sometimes generate empty docs?

Jiri-Kremser:
> How come we sometimes generate empty docs?

This happens if something goes wrong, for instance, it happened in metrics when it couldn't find the swagger template in the build tools, because it was still depending on the old version (without the template) or if anything goes wrong with the swagger generation process. It's better to have this check not to overwrite the real docs with obviously bad (empty) docs.

Further improvement would be to check the sha1 of newly generated docs and don't even perform the git push with curl in case there was no change. But GitHub handles this properly.. if there is a push with curl and there is no change in the file, no commit is performed.

PR #209, opened 2015-05-12
HWKMETRICS-84 Update ptrans to use tenant as a request header

tsegismont:
Ready for review.

### DIFF
--- a/clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
+++ b/clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
@@ -22,11 +22,6 @@ import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 
-import static org.hawkular.metrics.clients.ptrans.ConfigurationKey.BATCH_DELAY;
-import static org.hawkular.metrics.clients.ptrans.ConfigurationKey.BATCH_SIZE;
-import static org.hawkular.metrics.clients.ptrans.ConfigurationKey.REST_URL;
-import static org.hawkular.metrics.clients.ptrans.ConfigurationKey.SERVICES;
-import static org.hawkular.metrics.clients.ptrans.ConfigurationKey.TENANT;
 import static org.hawkular.metrics.clients.ptrans.util.ProcessUtil.kill;
 import static org.hawkular.metrics.clients.ptrans.util.TenantUtil.getRandomTenantId;
 import static org.junit.Assert.assertEquals;
@@ -51,6 +46,7 @@ import java.util.Locale;
 import java.util.Properties;
 import java.util.stream.Stream;
 
+import org.hawkular.metrics.clients.ptrans.ConfigurationKey;
 import org.hawkular.metrics.clients.ptrans.ExecutableITestBase;
 import org.hawkular.metrics.clients.ptrans.PrintOutputOnFailureWatcher;
 import org.hawkular.metrics.clients.ptrans.Service;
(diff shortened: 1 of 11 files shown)


clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
```diff
@@ -94,7 +96,7 @@
     @Before
     public void setUp() throws Exception {
         tenant = getRandomTenantId();
-        findGuageMetricsUrl = "http://" + BASE_URI + "/metrics?type=gauge&tenantId=" + tenant;
+        findGaugeMetricsUrl = "http://" + BASE_URI + "/metrics?type=gauge";
```
pilhuhn: No tenant needed here?

clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
```diff
@@ -242,12 +245,13 @@ private Point collectdLogToPoint(String line) {
     private List<Point> getServerData() throws Exception {
         ObjectMapper objectMapper = new ObjectMapper();
 
-        HttpURLConnection urlConnection = (HttpURLConnection) new URL(findGuageMetricsUrl).openConnection();
+        HttpURLConnection urlConnection = (HttpURLConnection) new URL(findGaugeMetricsUrl).openConnection();
+        urlConnection.setRequestProperty("tenantId", tenant);
```
pilhuhn: Also use `TENANT.getExternalForm()`?

clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
```diff
@@ -264,6 +268,7 @@ private Point collectdLogToPoint(String line) {
             String type = split[split.length - 1];
 
             urlConnection = (HttpURLConnection) new URL(findGaugeDataUrl(metricName)).openConnection();
+            urlConnection.setRequestProperty("tenantId", tenant);
```
pilhuhn: I think it is a bit confusing to me that you sometimes use `TENANT` and sometimes `tenantId`. While I see that the former is used for the config properties, it still confuses me. Perhaps not using a static import would help me here.

Btw: why did you introduce `getExternalForm()` and not override `toString()`?

pilhuhn:
Lgtm - my comment is more of a cosmetic nature

clients/ptranslator/src/test/java/org/hawkular/metrics/clients/ptrans/fullstack/CollectdITest.java
tsegismont: Le 12/05/2015 13:06, Heiko W. Rupp a écrit :
> I think it is a bit confusing to me that you sometimes use |TENANT| and
> sometimes |tenantId|. While I see that the former is used for the config
> properties, it still confuses me. Perhaps not using a static import
> would help me here.

I'll create another constant for the header name and won't use the 
static import for config param name.

>
> Btw: why did you introduce |getExternalForm()| and not override
> |toString()|?
>

Maybe I should have. This enum purpose is precisely to map config param 
names. So #toString may be more natural. Is that your thinking?

pilhuhn: On 12 May 2015, at 14:06, Thomas Segismont wrote:
> names. So #toString may be more natural. Is that your thinking?

Yes. And can even be omitted, as the implicit call to `toString()` would
then do the right thing already.

PR #210, opened 2015-05-13
Upgrade to parent 12

ppalaga:


rhqci:

rhqci:

stefannegrea:
Do not want to upgrade to Parent 12 because of the long and complex version numbers for build artifacts.

tsegismont:
I don't understand. The changes in parent POM should just add manifest entries.

ppalaga:
@stefannegrea , as @tsegismont said already, this change touches neither version numbers in pom.xml files nor artifact file names. It just adds git related entries into manifests of jars, wars and ears. That is acceptable, is it not?

stefannegrea:
@ppalaga thanks for the clarification. It all makes sense now, I will merge the change right away. 

@ppalaga, thank you the PR.

PR #211, opened 2015-05-13
HWKMETRICS-67 Influx endpoint: support pretty print query parameter

tsegismont:
PrettyFilter checks if the output should be formatted and PrettyInterceptor intercepts JSON response to format it.

I've tried to find how to interact with the RestEasy Jackson mappers directly but the solution would have been hard to maintain.

Here there's a little overhead due to parsing of the entity stream. But this will only be used by humans for a fraction of requests.


tsegismont:
Travis is happy but one of Filip's Jenkins slave is not configured (checked with him on IRC). That's why there's a red flag.

FilipB:
retest this please

FilipB:
retest this please

tsegismont:
@jsanda can you also review that one? It only applies the Influx endpoint. We'll see later the conclusion of the ML discussion about extending to all endpoints.

PR #216, opened 2015-05-14
HWKMETRICS-87 & HWKMETRICS-98

mwringe:
HWKMETRICS-87: the Cassandra containers are now able to be scaled down using a replication controller. This will properly decommission them from the cluster

HWKMETRICS-98: Store the Cassandra data to a volume instead of storing it within the container itself.


rhqci:

rhqci:

stefannegrea:
Thank you @mwringe

PR #218, opened 2015-05-19
HWKMETRICS-82 Refactor InfluxSeriesHandler

tsegismont:
Three itests fail on my box. The timestamps in the Influx response do not match expectations. Strangely enough, they are correctly sent to the server, but when the data is read back, the value is different.

Does that ring a bell @jsanda ?


tsegismont:
Looks like it was a data (from previous tests) conflict on my box. So PR is ready for review.

PR #219, opened 2015-05-19
HWKMETRICS-99

mwringe:
Wait until a Cassandra cluster is available instead of immediately failing.

WIP: Ideally we should be able to still keep the MetricServiceProducer as being eager. But, if the service waits during its construction, then you cannot terminate the server cleanly. This means you have to manually kill the server.


rhqci:

rhqci:

PR #220, opened 2015-05-19
Updates after Grafana integration testing

tsegismont:
- Catch query parse exception so that a proper response can be built
- Use plain text error messages as Influx does
- Fixed list series output bug
- Fix pretty interceptor/filter: InfluxHandler may change the media type (for errors)
- Support absolute moment operands in queries (select * from test where time > 1501560s and time < 4560546h)
- Configure InfluxObject explicitly to include empty points collection in JSON output

Also:
- make sure Influx tests never fail because of data conflict (user distinct tenantId/dbName)
- change ptrans log level in development config file


tsegismont:
I'll apply the changes to PR #218 after this PR is merged.

tsegismont:
Build failed on grunt execution. Restarted it.

PR #223, opened 2015-05-21
HWKMETRICS-99: allow Hawkular metrics to start and then wait for a Ca…

mwringe:
…ssandra cluster to become available rather than just failing.


rhqci:

rhqci:

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceProducer.java
```diff
@@ -70,14 +73,28 @@ void init() {
     @Produces
     public MetricsService getMetricsService() {
         if (metricsService == null) {
```
tsegismont: Could you move the code inside this conditional block to the #init method, thus replacing the `getMetricsService()` call? I believe this would make the intended behavior easier to understand.

tsegismont:
Could you add a method annotated with `@PreDestroy` to `MetricsServiceProducer` so that the resources are closed properly?

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/cassandra/MetricsServiceCassandra.java
```diff
@@ -145,13 +135,13 @@ public int hashCode() {
...
         System.getProperty(REQUEST_LIMIT, "30000")), 3, TimeUnit.MINUTES);
 
-    private Optional<Session> session;
-
     private DataAccess dataAccess;
 
     private final ListeningExecutorService metricsTasks = MoreExecutors
         .listeningDecorator(Executors.newFixedThreadPool(4, new MetricsThreadFactory()));
 
+    private boolean started = false;
```
tsegismont: This one should be `volatile`, I think

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceProducer.java
```diff
@@ -28,11 +28,14 @@
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 
+import com.datastax.driver.core.Session;
```
tsegismont: Minor: could you fix your IDE settings to follow the import ordering rules? There are some shared settings which you can import:
https://github.com/hawkular/hawkular-build-tools/tree/master/ide-configs

tsegismont:
In `HawkularMetricsRestApp`, there's a `MetricsService` injected:
```java
    @Inject
```
It used to be necessary because the `MetricsService` was created here (choosing between the (late) memory / C* backends). I'm not sure why it was kept when `MetricsServiceProducer` was created, but for sure now with the `EagerExtension` it's useless. Could you remove it?

tsegismont:
Overall, very nice PR @mwringe . Thanks!

I'll wait for your answers/updates before merging.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceProducer.java
```diff
@@ -70,14 +73,28 @@ void init() {
...
+
+            cassandraSessionBuilder.withInitializationCallback(new FutureCallback<Session>() {
+                @Override
+                public void onSuccess(Session session) {
+                    metricsService.startUp(session);
+                }
+
+                @Override
+                public void onFailure(Throwable t) {
+                    throw new RuntimeException("Error trying to get the Cassandra Session", t);
```
jsanda: You do not want to throw any exceptions in the callback methods. They will get swallowed and the thread running the callback may die silently. If I understand the initialization code, we should not reach the onFailure callback in normal circumstances because the async task blocks indefinitely until it establishes a connection. Maybe we should log an ERROR message here and either set some flag to indicate that server start up failed or shut down the server if possible.

jsanda:
It would be really nice to get some integration test written for this. I realize though that it might be somewhat difficult as we currently do not have anything in place for managing Cassandra lifecycle programmatically. We could use ccm like the DataStax driver does for its tests. The driver provides a Java wrapper for ccm which is available in Maven central. I think this could be done under a separate PR/ticket.

mwringe:
PR updated to fix the various concerns.

A separate jira task has been created to add integration tests to verify that we can handle the situations where Cassandra becomes unavailable: https://issues.jboss.org/browse/HWKMETRICS-102

tsegismont:
Thanks @mwringe !

PR #226, opened 2015-05-25
HWKMETRICS-96

mwringe:
Add a simple status page instead of displaying the default wildfly one.


rhqci:

rhqci:

tsegismont:
Looks good to me. @mwringe this is limited to the docker image. How about moving this page and its assets to the jaxrs web module? Then the welcome content page would redirect the http client to the `/hawkular/metrics` URI.

tsegismont:
Reopened. Sorry, I've closed it by accident. I just wanted to comment.

mwringe:
I added it separately in the welcome-content for two reasons:

1) I viewed this as being only really applicable to the containers. If you think this is something which could benefit the rest of the project, I have no problems moving it over to a more official location :)

2) Having it separate meant that even if the hawkular metrics webapp is still loading (or had an error starting) the user would at least see something graphical at the endpoint.  But, with the liveness checks, this probably wont be an issue as we would now check if hawkular metrics started properly or not.

So thinking about it now, it probably makes sense to move it over to the hawkular/metrics endpoint, and only have it displayed when the accept type is html.

tsegismont:
> 1) I viewed this as being only really applicable to the containers. If you think this is something which could benefit the rest of the project, I have no problems moving it over to a more official location :)

I believe it would benefit to all users, not only users of the container distribution

> 2) Having it separate meant that even if the hawkular metrics webapp is still loading (or had an error starting) the user would at least see something graphical at the endpoint. But, with the liveness checks, this probably wont be an issue as we would now check if hawkular metrics started properly or not.

True. But if the application doesn't load we'll get a 404 response so that could be checked by clients I guess.

> So thinking about it now, it probably makes sense to move it over to the hawkular/metrics endpoint, and only have it displayed when the accept type is html.

Perfect!

stefannegrea:
I prefer a simple REST endpoint structure that returns only JSON. So keeping the status page separate and only for docker is good for now.

tsegismont:
> I prefer a simple REST endpoint structure that returns only JSON. So keeping the status page separate and only for docker is good for now.

May I ask what is the risk you want to avoid?

mwringe:
Going through the code a bit on this, trying to expose web artifacts in the metrics-api-jaxrs war requires handling all of the artifacts (html, css, js) as rest resources. Which makes things a bit messy to handle.

pilhuhn:
For a status page it makes a lot of sense to also return it as human readable html that is returned if someone hits the page with a browser.
This does not need to be any fancy graphical page, but just plain good old html 1.0.

tsegismont:
@mwringe add this to the `web.xml` and you'll be fine:

```xml
  <servlet>
    <servlet-name>default</servlet-name>
    <servlet-class>io.undertow.servlet.handlers.DefaultServlet</servlet-class>
    <init-param>
      <param-name>resolve-against-context-root</param-name>
      <param-value>true</param-value>
    </init-param>
  </servlet>

  <servlet-mapping>
    <servlet-name>default</servlet-name>
    <url-pattern>/static/*</url-pattern>
    <url-pattern>/status.html</url-pattern>
  </servlet-mapping>
```

This works because exact matches and longer path matches have higher priority.

stefannegrea:
Again, I like boring code to go into the JAXRS implementation. That would allow for an easy port of that functionality to any other framework. The documentation is simple, and we could even go as far as on fly documentation.

So, let's keep the REST endpoint JSON only. Status in JSON is just fine if we want that. Mixing html and css and js and other things on an end-point if somebody specifies a different response time is a big deviation from simple, boring, almost identical code.

stefannegrea:
The PR as it was originally proposed is just perfect.

stefannegrea:
Per discussion with @mwringe, will merge this PR as is. Any improvements will be made post 0.3.4.

Thanks @mwringe !

tsegismont:
> Again, I like boring code to go into the JAXRS implementation. That would allow for an easy port of that functionality to any other framework.

I don't see any reason why moving an html page and a few static assets to a WAR file would prevent to port the code to another framework. I must be missing something. Let's say we'd want to implement Metrics with vertx-web. We'd route REST endpoints like this:

```java
Route route = router.route(HttpMethod.POST, "/catalogue/products/:productype/:productid/");

route.handler(routingContext -> {

  String productType = routingContext.request().getParam("producttype");
  String productID = routingContext.request().getParam("productid");

  // Do something with them...
});
```
And route the `/status` URI to another handler. So what's the matter?

> The documentation is simple, and we could even go as far as on fly documentation.

Sorry I don't understand this point.

> Mixing html and css and js and other things on an end-point if somebody specifies a different response time is a big deviation from simple, boring, almost identical code.

I don't understand this one either.

> The PR as it was originally proposed is just perfect.

It is a very good addition, indeed. And that's why I asked if it could be made so that it benefits to both docker and non dockers users. Looking at the thread again, I believe Matt and Heiko shared the feeling.

I have enough experience to understand that a good feature is better than a perfect feature. Now if "releasing today" is the reason to move forward I think the conversation could have happened here. We have nothing to hide, and "Open communication" is one of our core values.

Please in the future don't put an end to a conversation this way.

mwringe:
The branch has a few more commits to it:
1) the logo's are updated to use the newer ones
2) we display a json message when accessing /hawkular/metrics instead of throwing an error (it just display the name and version)

And the third new commit the the contentious one: https://github.com/mwringe/hawkular-metrics/commit/a8aab883b39caf80fd3c3d41eb5ade03261cd56e

The last commit moves the simple status page to the Hawkular Metrics war and makes it no longer container specific. 
We can at least see how that can be done in the code. Its not as messy when using the servlet-option that tsegismont suggested. But I also see how relying on that undertow specific servlet might makes things difficult if we were to try and use our war on another web server (is this even a real use case we are trying to support?)

mwringe:
FYI: I moved the issue over to two separate jiras since they are new feature requests. The updated simple json endpoint (HWKMETRICS-117) has already had a PR and been merged.

The html page which we don't have a consensus on have been moved it its own issue (HWKMETRICS-118) with its own PR (https://github.com/hawkular/hawkular-metrics/pull/239).

PR #225, opened 2015-05-25
HWKMETRICS-104

mwringe:
Create a status endpoint to display the state of things such as the MetricsService.

WIP: do not merge yet.

PR #229, opened 2015-05-26
HWKMETRICS-103

mwringe:
Add liveness probes and postStartup hooks to the hawkular metrics containers.
Update to use kubernetes v1beta03 api.

### DIFF
--- a/containers/hawkular-metrics-docker/src/main/docker/Dockerfile
+++ b/containers/hawkular-metrics-docker/src/main/docker/Dockerfile
@@ -33,13 +33,21 @@ RUN yum install -y java-1.8.0-openjdk java-1.8.0-openjdk-devel; \
 ENV JAVA_HOME /usr/lib/jvm/java-1.8.0
 USER jboss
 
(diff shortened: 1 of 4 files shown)


rhqci:

rhqci:

tsegismont:
> Please use the curl -H 'Hawkular-Tenant: status' instead and not the fallback

Well I'd say that we'd rather update the TenantFilter so that no tenant header is needed when sending requests to `status`

mwringe:
If HWKMETRICS-104 goes in, I wont be using the ping endpoint for this.

I view the /ping endpoint as being more of a check that the user can access their tenants endpoint (eg still requires a tenant id, user authentication token, etc). But that requires the security layer to be in place.

tsegismont:
Then shouldn't we finish HWKMETRICS-104 first? PR #225 says "do not merge yet"

mwringe:
OK, its been update to use the /status endpoint instead of /ping.

tsegismont:
Looks good to me. Thanks!

PR #230, opened 2015-05-26
Upgrade to hawkular-parent 14

ppalaga:


rhqci:

rhqci:

tsegismont:
Given that Mockito and Wildfly Maven Plugin are now declared in parent POM, shouldn't the Metrics POM files be updated, in order not to override parent definitions?

ppalaga:
@tsegismont yes, this is a good occasion, but I cannot promise to find the time to do that for metrics. Feel free you or anyone else to take 4c52582 , place the needed fixes on top of it and submit as a separate PR. But if there is nobody to do that now, I'd find it better to merge like this and create a followup Jira to address it later.

PR #232, opened 2015-05-27
[HWKMETRICS-105] Remove tenant id fallback mechanism. Only way to sen…

stefannegrea:
…d tenant id is via request header.


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/TenantFilter.java
```diff
@@ -63,14 +63,6 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
             return;
         }
 
-        tenant = uriInfo.getQueryParameters().getFirst(TENANT_QUERY_PARAM_NAME);
```
tsegismont: You should also update `MISSING_TENANT_MSG`  and remove `TENANT_QUERY_PARAM_NAME`

stefannegrea: Sorry, I just missed it ...

PR #234, opened 2015-05-28
Single definition fetching only works when tags are present

burmanm:
### https://issues.jboss.org/browse/HWKMETRICS-110

Do not merge yet (tests are failing on my machine for all tests, so I can't verify this yet)

* Change findMetrics to return Observable<Metric instead of Optional (doesn't make much sense with RxJava to use Optional)
* Add some tests and try to fix the fetching single definition

Do not merge yet, @jsanda could you comment if the approach to reading from metrics_idx is the right way?

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
@@ -123,8 +123,10 @@ public class AvailabilityHandler {
             @HeaderParam("tenantId") String tenantId, @PathParam("id") String id) {
 
         metricsService.findMetric(tenantId, MetricType.AVAILABILITY, new MetricId(id))
+                .map(m -> Response.ok(m).build())
+                .defaultIfEmpty(Response.noContent().build())
                 .subscribe(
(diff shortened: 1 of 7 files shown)


jsanda:
I agree that using Optional does not make much sense. +1 on that change. With these changes, can we get rid of the ```DataAccessImpl.findMetric``` query/field?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
```diff
@@ -123,8 +123,10 @@ public void getAvailabilityMetric(@Suspended final AsyncResponse asyncResponse,
             @HeaderParam("tenantId") String tenantId, @PathParam("id") String id) {
 
         metricsService.findMetric(tenantId, MetricType.AVAILABILITY, new MetricId(id))
+                .map(m -> Response.ok(m).build())
+                .defaultIfEmpty(Response.noContent().build())
```
burmanm: This defaultIfEmpty() is triggered even when there's rows to map(). Will have to wait until I'm back.. removing that one will return correctly the definition.

tsegismont:
@burmanm where do you stand with this one?

burmanm:
Replaced by PR #278

PR #236, opened 2015-06-01
HWKMETRICS-115 Create a Gatling load testing scenario

tsegismont:
Create a Gatling load testing scenario

### DIFF
--- a/load-tests/README.adoc
+++ b/load-tests/README.adoc
@@ -1,0 +1,23 @@
+= Hawkular Metrics Load Tests
+:type: article
+
+A simple load testing tool for Hawkular Metrics
+
+== Why a Maven project?
+
+Because it makes it easier for developers to fire a load test:
+there's no need to download, extract and configure PATH.
+
+== The scenario
+
+The simulation consists in sending metrics reports from a configurable number of clients (like a collector sending
+reports periodically). On top of the number of clients, many aspects of the simulation are configurable:
+number of loops, interval, number of metrics per report, ... etc.
+
+See the scenario file for the list of options.
+
+== Running
+
+ mvn gatling:execute -Dxxx=y ...
+
+Each scenario option can be set with a system property.
--- a/load-tests/README.adoc
+++ b/load-tests/README.adoc
@@ -21,3 +21,6 @@ See the scenario file for the list of options.
  mvn gatling:execute -Dxxx=y ...
 
 Each scenario option can be set with a system property.
+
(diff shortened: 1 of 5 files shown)


slandelle:
:beers:

tsegismont:
Thanks for your help @slandelle :wink:  I'll open the beer tonight, I promise! :beer:

tsegismont:
@slandelle by the way, I haven't checked if the intermediary stats can be sent to Hawkular Metrics. Maybe with our Influx compatibility layer it would work OTB. Eventually we could have a native Hawkular Metrics.

load-tests/README.adoc
```diff
@@ -0,0 +1,23 @@
...
+:type: article
+
+A simple load testing tool for Hawkular Metrics
+
+== Why a Maven project?
+
+Because it makes it easier for developers to fire a load test:
+there's no need to download, extract and configure PATH.
+
+== The scenario
```
jsanda: The README should indicate that you need to start up Cassandra and Hawkular separately.

tsegismont: Done. See commit 2fbdf97

jsanda:
I am ok with merging this. It is a good initial effort. @stefannegrea, @mwringe wdyt?

mwringe:
I am ok with merging as well. Is there any plans to include more than just the gauges as part of this?

jsanda:
> I am ok with merging as well. Is there any plans to include more than just the gauges as part of this?
I think this is just an initial PoC that demonstrates how to get up and running with Gatling. We definitely will test more than just gauges. 

I should also point out that I am not necessarily advocating for/against Gatling. Getting it into master increases the likelihood of others exploring it and providing more feedback.

tsegismont:
Le 02/06/2015 18:08, mwringe a écrit :
> I am ok with merging as well. Is there any plans to include more than
> just the gauges as part of this?

In the future we will make it possible to include availability in the 
mix of values.

tsegismont:
Le 02/06/2015 18:21, jsanda a écrit :
> I should also point out that I am not necessarily advocating for/against
> Gatling. Getting it into master increases the likelihood of others
> exploring it and providing more feedback.

To me, the ability to simulate a great number of clients from the same 
machine is a very good advantage.

Other than that, what's really missing when starting a load test is 
"metrics from Metrics" :)

I would like to include Codahale's library soon and start exposing 
performance indicators. Like metrics ingested/min.

PR #239, opened 2015-06-01
HWKMETRICS-118

mwringe:
Add a simple html page to the base hawkular metrics endpoint.


rhqci:

rhqci:

mwringe:
Note: there is no consensus on if this feature should go into Hawkular Metrics or not.

We should wait until consensus is reached before doing a PR

pilhuhn:
I believe very much in http/rest content negotiation and that a thing like a status page should be as well machine readable as well as human readable (and hey, if a user wants a png, we can even return the status as picture of a checkmark). This can be done like you do.
Or the code can just return html by "string concatenation"

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/BaseHandler.java
```diff
@@ -51,6 +56,17 @@ public Response baseJSON(@Context ServletContext context) {
         return Response.ok(hawkularMetrics).build();
     }
 
+    @GET
+    @Produces({APPLICATION_XHTML_XML, TEXT_HTML})
+    @ApiOperation(value = "Returns some basic information about the Hawkular Metrics service.",
```
tsegismont: I don't think it should be documented in Swagger

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/BaseHandler.java
```diff
@@ -51,6 +56,17 @@ public Response baseJSON(@Context ServletContext context) {
...
+                  response = String.class, responseContainer = "Map")
+    public void baseHTML(@Context ServletContext context) throws Exception {
+
+        HttpServletRequest request = ResteasyProviderFactory.getContextData(HttpServletRequest.class);
```
tsegismont: You should be able to declare `HttpServletRequest` and `HttpServletResponse` params annotated with `@Context`

Then the methods becomes as simple as:
```java
public void baseHTML(@Context HttpServletRequest request, @Context HttpServletResponse response) {
  request.getRequestDispatcher("/static/index.html").forward(request,response);
}
```

tsegismont: Or you could simply put the index.html page in the root dir and add `/index.html` as an url-mapping of the `staticContent` servlet.

api/metrics-api-jaxrs/src/main/webapp/static/index.html
```diff
@@ -0,0 +1,42 @@
...
+    <title>Hawkular Metrics</title>
+    <link rel="shortcut icon" href="metrics/static/favicon.ico" type="image/x-icon">
+    <link rel="StyleSheet" href="metrics/static/welcome.css" type="text/css">
+    <script src="metrics/static/status.js"></script>
+</head>
+
+<body>
+  <div class="content">
+    <div class="logo"/>
+      <img src="metrics/static/hawkular_logo.png" alt="Hawkular Logo"/>
```
tsegismont: Are you sure this `src` works?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/BaseHandler.java
mwringe: You can't use @Context here, @Context has to returns a proxy object and the RequestDispather.forward(request, response) will only accept the original objects. See https://issues.jboss.org/browse/RESTEASY-745

I could put index.html in the root, but the images and css would either have to stay into the /static directory or we would have to add more logic to the BaseHandler for each of those files. I believe it makes more sense to keep all of the static content under /static and use the servlet mapping to handle all the static files.

api/metrics-api-jaxrs/src/main/webapp/static/index.html
mwringe: It worked last time I checked it, are you not seeing the images or javascript working anymore?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/BaseHandler.java
tsegismont: Le 06/07/2015 15:27, mwringe a écrit :
> You can't use @Context <https://github.com/Context> here, @Context
> <https://github.com/Context> has to returns a proxy object and the
> RequestDispather.forward(request, response) will only accept the
> original objects. See https://issues.jboss.org/browse/RESTEASY-745
>

Ha. Bad. I didn't know.

> I could put index.html in the root, but the images and css would either
> have to stay into the /static directory or we would have to add more
> logic to the BaseHandler for each of those files. I believe it makes
> more sense to keep all of the static content under /static and use the
> servlet mapping to handle all the static files.
>

Sure, I'm not suggesting to move css, images and js out of /static. I 
was talking about removing everything non JSON from BaseHandler and then 
adding this to web.xml:

<servlet-mapping>
   <servlet-name>staticContent</servlet-name>
   <url-pattern>/index.html</url-pattern>
</servlet-mapping>

mwringe: I don't believe that will work. '/' and all its descendants are owned by the BaseHander, but we use the servlet-mapping to over ride it for anything under '/static/*'.

If we want to use just the servlet-mapping, then we will also need to create individual entries for all other types of static files (*.css, *.js, *.png, etc). 

I think we would still need to create an entry in BaseHandler anyways to forward '/' to '/index.html' only when the request type is html instead of json.

tsegismont: Le 06/07/2015 16:14, mwringe a écrit :
> I don't believe that will work. '/' and all its descendants are owned by
> the BaseHander, but we use the servlet-mapping to over ride it for

That's not true. /gauges/{id}/data is a descendant and it's not handled 
by the BaseHandler.

> anything under '/static/*'.
>
> If we want to use just the servlet-mapping, then we will also need to
> create individual entries for all other types of static files (*.css,
> *.js, *.png, etc).

No, /static/* is just fine.

>
> I think we would still need to create an entry in BaseHandler anyways to
> forward '/' to '/index.html' only when the request type is html instead
> of json.
>

No, it's not needed. Because servlet path matching rules, like JAX-RS 
ones, favor longer matches.

api/metrics-api-jaxrs/src/main/webapp/static/index.html
tsegismont: Le 06/07/2015 15:28, mwringe a écrit :
> It worked last time I checked it, are you not seeing the images or
> javascript working anymore?
>

No, I think it only works when you use the Docker image, because you 
load the content of this page from the root context with meta refresh.

But when you browse to /hawkular/metrics you can't see anything (no 
image, no status)

mwringe:
OK, I rebased to the latest and updated the commit.

The issue with the src specified images and css has been fixed (it partially worked before, but only on hawkular/metrics not hawkular/metrics/ nor hawkular/metrics/static/index.html).

I can't seem to be able to remove the logic from the BaseHandler and still have it work. Moving the index.html to the base directory (or specifying the welcome-page to specify the one in /static) didn't seem to work properly. It may have to do with the '/' already being owned by the rest app.
If you can get it working without modifying the BaseHandler, let me know

tsegismont:
@mwringe sorry I was wrong. I was focused on the idea that '/index.html' could be matched by the static servlet, which is true, but the problem is we're trying to reach it through '/' which will always be matched by the Resteasy servlet.

Thank you for fixing the image and css issues, and thank you for porting this nice feature to the application WAR.

PR #243, opened 2015-06-04
HWKMETRICS-119 Refactor configuration and session initialization code

tsegismont:
The PR has been divided into many small commits which can be reviewed in order. It makes the review easier, as each commit message explains the different steps).

Note that the `MetricsServiceLifecycle` bean checks if "hawkular-metrics.backend=embedded_cass" for backward compatibility purposes. When Hawkular build is updated to use the new ""hawkular.metrics.waitForService" flag, all remaining references to this system properties will be removed.

Also, after this PR is merged, I'll update the Metrics docs on the Hawkular website.


tsegismont:
The `haw-perf-baseline-50000` check is failing, but Travis is happy. So don't trust the red cross.

tsegismont:
Rebased on master

jsanda:
I think I am -1 on moving all of the lifecycle/initialization logic out of MetricsService. I think it needs to have start/init and shutdown methods for setting up and shutting down thread pools. I also think that it makes sense to keep schema initialization in MetricsService. Here's how I look at it. Suppose I want to run metrics-core in another, different environment other than our JAX-RS server. I am going to have to duplicate a lot of the initialization and set up code that is very specific to the core service.

tsegismont:
Le 05/06/2015 16:30, jsanda a écrit :
> I think I am -1 on moving all of the lifecycle/initialization logic out
> of MetricsService. I think it needs to have start/init and shutdown
> methods for setting up and shutting down thread pools. I also think that

MetricsServiceImpl still has a startUp method. You're right for the 
shutdown of the thread pool, it wasn't implemented and I missed that 
part. I'll push a new commit.

The important thing is that the MetricsService interface does not expose 
these implementation details.

> it makes sense to keep schema initialization in MetricsService. Here's
> how I look at it. Suppose I want to run metrics-core in another,
> different environment other than our JAX-RS server. I am going to have
> to duplicate a lot of the initialization and set up code that is very
> specific to the core service.

It's not a lot of code actually:
- 5 lines to create the Cluster and Session instances
- 1 line to drop the keyspace if required
- 2 lines to call the SchemaManager if required

I'm ok to move it to the core if you think it's better.

tsegismont:
> It's not a lot of code actually:
> - 5 lines to create the Cluster and Session instances
> - 1 line to drop the keyspace if required
> - 2 lines to call the SchemaManager if required
> I'm ok to move it to the core if you think it's better.

While looking at this I realized that we already have what's needed in the SchemaManager, which is in the core.

So I've cleaned MetricsServiceLifecycle a bit more which now does nothing else on startup than:
* creating a `Session` based on config (nodes+port)
* calling `SchemaManager`
* calling `MetricsServiceImpl#startup`
And on shutdown, calling `MetricsServiceImpl#shutdown`


See commit 27800c9

jsanda:
I do think the schema initialization/update belongs in the core. It is an implementation detail of the service. Other than that I think the changes overall look pretty good. I can do a more thorough review a bit later.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -0,0 +1,241 @@
...
+    public MetricsService getMetricsService() {
+        return metricsService;
+    }
+
+    @PreDestroy
+    void destroy() {
+        Future stopFuture = lifecycleExecutor.submit(this::stopMetricsService);
+        try {
+            Futures.get(stopFuture, 1, MINUTES, Exception.class);
+        } catch (Exception ignore) {
```
jsanda: If there is an exception, we should at least log it.

tsegismont: In general, logging the exception is the right thing to do. But 
sometimes it's really not necessary and I believe it's the case here.

Futures#get takes a checked exception as argument and wraps any Timeout, 
Interrupted or Execution exception with an instance of it. And the code 
which is scheduled for execution is very simple: call 
MetricsService#shutdown and close the C* Session.

It's very much like when you close JDBC result set and connection in a 
finally block, the methods throw a checked Exception but at this stage 
you don't really care.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -0,0 +1,241 @@
...
+    @PostConstruct
+    void init() {
+        lifecycleExecutor.submit(this::startMetricsService);
+        if (Boolean.parseBoolean(waitForService)
+            // "hawkular-metrics.backend" is not a real Metrics configuration parameter (there's a single
+            // MetricsService implementation, which is backed by Cassandra).
+            // But it's been used historically to wait for the service to be available before completing the deployment.
+            // Therefore, we still use it here for backward compatibililty.
+            // TODO remove when Hawkular build has been updated to use the eager startup flag
+            || "embedded_cass".equals(System.getProperty("hawkular-metrics.backend"))) {
```
jsanda: When we started embedding Cassandra as an EAR, one of the issues we faced was with start up and initialization. It was possible for metrics to start up before Cassandra was initialized and ready to accept client connections. Checking for the flag here was done to let us know that we should check to see if Cassandra has initialized. Because we now support hawkular starting before Cassandra, I think we need the initialization logic regardless of whether or not Cassandra is embedded. Can we just altogether remove this check?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -0,0 +1,241 @@
...
+        } catch (Throwable t) {
+            Throwable rootCause = Throwables.getRootCause(t);
+            LOG.warn("Could not connect to Cassandra cluster - assuming its not up yet", rootCause);
+            // cycle between original and more wait time - avoid waiting huge amounts of time
+            long delay = 1 + ((connectionAttempts - 1) % 4);
+            LOG.warn("[{}] Retrying connecting to Cassandra cluster in [{}]s...", connectionAttempts, delay);
+            lifecycleExecutor.schedule(this::startMetricsService, delay, SECONDS);
+            return;
+        }
+        SchemaManager schemaManager = new SchemaManager(session);
```
jsanda: As I mentioned previously, I think the schema initialization should be done in ```MetricsServiceImpl```.  It is an implementation detail of the service. Right now the the CQL statements for the task service are in the same cql file, but  they will be separated out at some point as the task service schema is independent of the rest of the metrics schema.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -0,0 +1,241 @@
...
+        Arrays.stream(nodes.split(",")).forEach(clusterBuilder::addContactPoint);
+
+        Cluster cluster = null;
+        Session session = null;
+        try {
+            cluster = clusterBuilder.build();
+            session = cluster.connect("system");
+            return session;
+        } finally {
+            if (session == null && cluster != null) {
```
jsanda: Overall I like the changes for managing the Session outside of MetricsService in large part because I think that ultimately we should be sharing the same Session object across services. This is a good, necessary step towards making that possible.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
tsegismont: I've pushed two new commits: ba1c3f5 and 44002ca

PR #244, opened 2015-06-05
The REST api of metrics has still base path '{host}/hawkular-metrics/…

Jiri-Kremser:
…'. It'd be better to change that to align with other components where the convention is '{host}/hawkular/{componentName}/' However, for now I am changing the path for the generated docs to reflect the reality.

### DIFF
--- a/api/metrics-api-jaxrs/pom.xml
+++ b/api/metrics-api-jaxrs/pom.xml
@@ -217,7 +217,7 @@
                 <apiSource>
                   <locations>org.hawkular.metrics.api.jaxrs</locations>
                   <apiVersion>1.0</apiVersion>
-                  <basePath>http://localhost:8080/hawkular/metrics/</basePath>
+                  <basePath>http://localhost:8080/hawkular-metrics/</basePath>
                   <outputTemplate>${project.build.directory}/dependency/hawkular-documentation/asciidoc.mustache</outputTemplate>
                   <swaggerDirectory>${project.build.directory}/generated/swagger-ui</swaggerDirectory>
                   <swaggerInternalFilter>org.hawkular.metrics.api.jaxrs.swagger.filter.JaxRsFilter</swaggerInternalFilter>


tsegismont:
@Jiri-Kremser I don't fully understand. In master, Metrics URIs start with `/hawkular/metrics`. That's the expectation so what's the problem?

stefannegrea:
I am -1 on this change. We are in the process to migrated Hawkular to Metrics 0.3.4. The change would no longer be needed once the migration is done.

Jiri-Kremser:
I was reacting on following chat on IRC:
```
<ming> I’m trying to create a tenant manually via API. Both "POST inventory/tenant" and "POST 
metrics/tenants" with "{'id': 'UUID'}" unfortunately fail. Can somebody suggest a proper API call?

<viet> ming:  curl -X POST -H "Content-Type: application/json" -d '{"id": "mytenant"}' 
http://209.132.179.82:18080/hawkular-metrics/tenants

<ming> viet, thanks! The documentation mislead me. http://www.hawkular.org/docs/rest/rest-metrics.html 
says metrics are at :/hawkular/metrics.
```


If metrics are really deployed on `/hawkular/metrics`, I probably got something wrong and this PR can be closed.

Jiri-Kremser:
the UI seems to call `hawkular-metrics/...` too, https://github.com/hawkular/hawkular-ui-services/blob/85f6f4d04c1e68e255bb4eba30ad0df49a86c3c9/src/rest/hawkRest-metric-factory.spec.js#L17 so how can metrics be deployed on `hawkular/metrics`

Is it because of the different version? Old one is on `hawkular-metrics/...` and the new one on `hawkular/metrics` ? I am quite confused here.

tsegismont:
Le 05/06/2015 23:21, Jirka Kremser a écrit :
> Is it because of the different version? Old one is on
> |hawkular-metrics/...| and the new one on |hawkular/metrics| ?

Yes

PR #245, opened 2015-06-05
[HWKMETRICS-120] Start Thrift Server in embedded Cassandra service.

metlos:
Please review.

The JIRA mentions problems with long shutdown times after this change. We should investigate that before merging this.


rhqci:

rhqci:

rhqci:

stefannegrea:
Thanks @metlos !

PR #247, opened 2015-06-09
ptrans docs update

tsegismont:
Minor updates and fixes to the ptrans docs

### DIFF
--- a/clients/ptranslator/README.adoc
+++ b/clients/ptranslator/README.adoc
@@ -1,4 +1,4 @@
-= ptrans - An Hawkular Metrics proxy
+= ptrans - A Hawkular Metrics proxy
 :type: article
 
 This is a proxy server taking in metrics data in several common formats and emitting
@@ -10,9 +10,11 @@ them as REST requests into Hawkular Metrics backend for storage.
 
 This module contains a simple proxy server that also listens on UDP port 5140
 for syslog messages. If a message is of the format:
+
 ----
 type=metric thread.count=7 thread.active=5 heap.permgen.size=20000000
 ----
+
 Then it is taken and the individual key-value-pairs are broken apart and
 sent as individual metrics to a hawkular-metrics rest server. The timestamp
 is the current time on the proxy when the udp packet was received.
@@ -21,6 +23,7 @@ is the current time on the proxy when the udp packet was received.
 
 You need to configure syslog to forward you the messages.
 For this you need to edit `/ec/syslog.conf` and add a line like this:
+
 ----
 *.info          @127.0.0.1:5140
 ----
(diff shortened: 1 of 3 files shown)


tsegismont:
Can anyone please review this small doc update?

clients/ptranslator/README.adoc
```diff
@@ -1,23 +1,18 @@
-= ptrans
+= ptrans - An Hawkular Metrics proxy
```
pilhuhn: A Hawkular ?

clients/ptranslator/README.adoc
```diff
@@ -58,7 +55,6 @@ to `gmond` udp mlticast packets.
 For this to work, the packets from Ganglia need to have a large enough TTL to be seen on the host where you run pTrans.
 
 Excerpt from gmond.conf:
-
```
pilhuhn: Generally there should be an empty line before such blocks (especially when they are denoted with `[source,...]` as otherwise the syntax coloring does not work.

clients/ptranslator/README.adoc
```diff
@@ -70,41 +66,54 @@ udp_send_channel {
...
+<Plugin network>
+        <Server "127.0.0.1" "25826">
+        </Server>
+</Plugin>
+----
 
+== Running ptrans for development
 
-== Running ptrans
+Don't forget to start an Hawkular Metrics web server as well.
```
pilhuhn: Make that a `TIP:` ?

clients/ptranslator/README.adoc
```diff
@@ -70,41 +66,54 @@ udp_send_channel {
...
  $ ./start.sh -c ptrans.conf
+----
 
-*With Maven*
-
-Once the project is built, simply run the following command:
+This method does not allow to configure logging and works on *nix systems only.
 
- mvn exec:java
+=== With Maven ===
```
pilhuhn: ADoc does not need the ending ===

clients/ptranslator/README.adoc
```diff
@@ -70,41 +66,54 @@ udp_send_channel {
...
+----
 
-== Room for improvement:
+By default, the log4j and ptrans.conf files at the project base directory will be used.
+If you want to use your own files:
+----
+mvn exec:java -Ddev.log4j.configuration=<path to log4j file> -Ddev.ptrans.conf=<path to ptrans.conf>
+----
 
-* add information about the remote to the key so make the key unique
```
pilhuhn: Why are you removing those? Do we have enhancement jiras instead?  Perhaps link to them?

tsegismont:
@pilhuhn thanks for the review, I just pushed a new commit

clients/ptranslator/README.adoc
tsegismont: Le 10/06/2015 08:35, Heiko W. Rupp a écrit :
>>
>> -* add information about the remote to the key so make the key unique
>
> Why are you removing those? Do we have enhancement jiras instead?
> Perhaps link to them?

A good part is already implemented. I'll create a JIRA for metric naming 
configuration. I'm not sure what is expected wrt to syslog tailer. Can 
you explain? Or create the JIRA yourself?

WRT linking the issues, I guess the goal is to attract contributions? In 
this case I believe creating a "Low Hanging Fruit" tag in JIRA is a 
better option. Then we can add that to all README files: search JIRA 
with the "Low Hanging Fruit" tag. We could even provide a link.

README files are not very often updated so having only general, non 
moving information is better, I think.

PR #249, opened 2015-06-09
Project README update

tsegismont:
An update to reflect the current state of the project and highlight goals.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -13,10 +13,12 @@ endif::[]
 Hawkular Metrics is the metric data store for the http://www.hawkular.org/[Hawkular] project.
 It can also be used independently.
 
-It uses https://cassandra.apache.org/[Apache Cassandra] as a backend and is comprised of:
+It relies on https://cassandra.apache.org/[Apache Cassandra] as a backend and is comprised of:
 
 - a core library
 - a REST/HTTP interface
+
+IMPORTANT: Cassandra 2.1.1 or later is required. It is recommended to use the latest 2.1.x release if possible.
 
 === The core library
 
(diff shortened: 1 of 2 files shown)


tsegismont:
Can anyone please review this doc update?

For the JAX-RS module README, I believe we should remove endpoint documentation, which is already published on the H* website, and keep simple JSON samples only (like POST metric data and read metric data).

README.adoc
```diff
@@ -8,66 +8,98 @@ image::https://travis-ci.org/hawkular/hawkular-metrics.svg?branch=master[Build S
...
-= Project to implement a MaaS and ChaaS
+== About
 
-* MaaS: Metrics as a Service
-* ChaaS: Charts as a Service
+Hawkular Metrics is the metric data store for the http://www.hawkular.org/[Hawkular] project.
+It can also be used independently.
 
-== Quickstart
+It uses https://cassandra.apache.org/[Apache Cassandra] as a backend and is comprised of:
```
jsanda: Might be good to mention here that we require Cassandra 2.1.2 or later.

tsegismont:
@jsanda thx for the review. I've pushed a commit to emphasize the dependency on C* 2.1.x at the beginning of the file ( b70cf99 ).

tsegismont:
Thanks for reviewing and merging. Any opinion on the JAX-RS module README question?

jsanda:
I think it looks good. I think it might make sense to eventually try to move most the JAX-RS docs to the hawkular site as well, particularly the request/response examples. We need to keep up with docs so it is good to see these changes. Thanks!

PR #254, opened 2015-06-11
HWKMETRICS-90

mwringe:
Add in container tests (enabled via profile) and fixes required for the tests.

### DIFF
--- a/containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
+++ b/containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2014-2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.containers;
+
+import static org.junit.Assert.assertEquals;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.fabric8.kubernetes.api.model.ReplicationController;
+import org.jboss.arquillian.junit.Arquillian;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @author mwringe
+ */
+@RunWith(Arquillian.class)
+public class TestCassandraScaling extends BaseContainerTests {
+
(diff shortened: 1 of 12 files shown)


containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
```diff
@@ -0,0 +1,77 @@
...
+
+        while(client.getPodsForService(CASSANDRA_NAME).size() < 2) {
+            Thread.sleep(1000);
+        }
+
+        testWriteData("2reps", "22.42", 2);
+
+        // Test scaling down to 1 replica
+        rc = client.getReplicationController(CASSANDRA_NAME);
+        rc.getSpec().setReplicas(1);
```
jsanda: Does this remove one of the Cassandra nodes? And if so, is it shutdown or permanently removed? (just curious)

containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
```diff
@@ -0,0 +1,77 @@
...
+    public void testScaling() throws Exception {
+        // Test with the intial setup of 1 replica
+        testWriteData("initial", "11.102", 1);
+        ReplicationController rc = client.getReplicationController(CASSANDRA_NAME);
+
+        // Test scaling up to 2 replicas
+        rc.getSpec().setReplicas(2);
+        client.updateReplicationController(CASSANDRA_NAME, rc);
+
+        while(client.getPodsForService(CASSANDRA_NAME).size() < 2) {
```
jsanda: What exactly does it mean when the size is two? Does it mean that kubernetes knows the pod is up and running? Or does it mean Cassandra is running? A nice future enhancement would be to wait until the new C* node has finished bootstrapping.

jsanda:
What are the prerequisite steps for running these tests?

containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
mwringe: When if the replica number is decreased, it will remove one of the Cassandra nodes (it gets decommissioned).

containers/container-tests/src/test/java/org/hawkular/containers/TestCassandraScaling.java
mwringe: When the size it two, it means that there there are 2 Cassandra pods running. It does not necessarily mean that the Cassandra node is fully started or not.
I have looked into making it so that a node stays in the 'pending' state until it finishes bootstrapping, but then it means we can no longer discover the other nodes in the cluster. Its something we need to fix.

mwringe:
There are a lot of prerequisites:
- have docker installed (& configured)
- have openshift installed (& configured)
- setup and configure the settings that the tests require.

Its not something fun to setup the first time, but once its configured its not that bad when you need to rerun the tests.

We need to have an external machine configured to run all of this for us, the Fabric8 Continuous Delivery suite will allow us to automate a lot of this: build the docker images, build the kubernetes app, test the docker images and kubernetes app, push the tested docker images out to docker hub, etc

jsanda:
Thanks for the info on the prerequisites. If there are not already any docs, this info would serve as a good starting point for some docs on how to get up and running.

PR #257, opened 2015-06-15
POM changes

tsegismont:
Use C* driver and Guava version from parent + mvn tidy:pom

### DIFF
--- a/containers/kubernetes/pom.xml
+++ b/containers/kubernetes/pom.xml
@@ -31,7 +31,7 @@
   <artifactId>hawkular-metrics-kubernetes-app</artifactId>
   <version>0.3.5-SNAPSHOT</version>
 
-  <name>Hawkular Metrics Kuberenetes Application</name>
+  <name>Hawkular Metrics Kubernetes Application</name>
 
   <properties>
     <!-- Properties for the fabric8-maven-plugin -->
--- a/containers/kubernetes/pom.xml
+++ b/containers/kubernetes/pom.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
 
     Copyright 2014-2015 Red Hat, Inc. and/or its affiliates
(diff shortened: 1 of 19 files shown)


containers/kubernetes/pom.xml
```diff
@@ -29,9 +29,10 @@
   </parent>
 
   <artifactId>hawkular-metrics-kubernetes-app</artifactId>
-  <name>Hawkular Metrics Kuberenetes Application</name>
   <version>0.3.5-SNAPSHOT</version>
 
+  <name>Hawkular Metrics Kuberenetes Application</name>
```
pilhuhn: Typo in K8s

pilhuhn:
Lmgtm  - perhaps @ppalaga to double check?

containers/kubernetes/pom.xml
tsegismont: Le 17/06/2015 10:35, Heiko W. Rupp a écrit :
> Typo in K8s

Fixed

ppalaga:
I do not see anything suspicious, but note that /me knows little about metrics internals.

tsegismont:
Thank you guys

jsanda:
We need to be careful about the version of Guava we use since it is also a dependency of the C* driver. We have bumped the driver version to 2.1.6, which still uses Guava 14.0.1. I have run into problems in the past by using a different version. Just need to be careful. The same would apply with other driver deps like Netty.

tsegismont:
Le 17/06/2015 13:21, jsanda a écrit :
> We need to be careful about the version of Guava we use since it is also
> a dependency of the C* driver. We have bumped the driver version to
> 2.1.6, which still uses Guava 14.0.1. I have run into problems in the
> past by using a different version. Just need to be careful. The same
> would apply with other driver deps like Netty.

Good point. Note that I haven't changed the effective POMs, only removed 
the Guava version declaration in Metrics, as it was the same as in parent.

PR #259, opened 2015-06-18
Update containers module C* version

tsegismont:


pilhuhn:
Why is that needed? What is the Jira? Does that have an impact on Hawkular?

burmanm:
2.1.x series is very buggy, so I say +1 on this.

jsanda:
FYI, I created [HAWKULAR-364)](https://issues.jboss.org/browse/HAWKULAR-364) yesterday.

tsegismont:
Le 18/06/2015 11:15, Heiko W. Rupp a écrit :
> Why is that needed?

Here's how it started: yesterday I saw some commits in Metrics master 
about embedded C* updates and a new requirement for version 2.1.5 or 
greater. So I've updated the website docs as well (you merged them).

Today, before publishing my blog post, Thomas asked me if I was 100% 
sure Metrics could be built and run on Windows. The only way to answer 
was to start my good old 2008R2 VM, clone the git repo, build and start.

While Maven was downloading the world, I noticed it spent quite some 
time on downloading Cassandra... 2.1.3. That was way too contradictory 
for a maniac like me.

So I found where this discrepancy was and realized the last version of 
C* 2.1.x available on Maven was 2.1.6, not 2.1.5. Probably because it 
was published early June, after we started the embedded C* upgrade.

Besides, 2.1.6 is the driver version we use (from parent POM)

> What is the Jira?

There's none from my side, sorry.

> Does that have an impact on Hawkular?

No. It has an impact on the C* containers built for Openshift.




All that said, I believe we should:

1. define a version property for org.apache.cassandra:cassandra-all in 
the parent POM, next to the com.datastax.cassandra:cassandra-driver-core 
version property

2. use the same version value in both properties

That might require us to wait for both artifacts to be published (the 
2.1.6 driver was released end of May)

3. Add some components in the parent to remind us that both versions 
should be changed at the same time, and that READMEs and website docs 
should be updated.

jsanda:
> 2. use the same version value in both properties

This might be ok sometimes, but I do think this is something we want to necessarily always do. The driver and Cassandra are independent projects with different release cycles. Here is a [compatibility guide](http://docs.datastax.com/en/developer/driver-matrix/doc/common/driverMatrix.html). When it makes sense to upgrade one whether it is for new features, bug fixes, etc., then we ought to look to do so regardless of the other.

tsegismont:
Le 18/06/2015 15:22, jsanda a écrit :
>      1. use the same version value in both properties

This wasn't my point #1, but #2, you cheater :p

I had a look at the version compatibility guide, they only say you 
should use version 2.1.x of the driver if you want to be able to use all 
features of C* 2.1.x.

I noticed that for every single 2.1.1x, there's a corresponding driver 
version. That's why I thought they were linked. But looking at the 
release dates, you're right, there could be two releases of the Driver 
before C* goes out. Or the opposite.

stefannegrea:
@mwringe can you please accept or reject this PR?

mwringe:
Looks good to me and the container tests still pass. Merged

PR #260, opened 2015-06-19
Upgrade to hawkular-parent 16 - upgarde to WildFly 9

ppalaga:


tsegismont:
Thanks @ppalaga . It doesn't work as is and the move to parent 16 is now tracked by HWKMETRICS-145. So I'm closing this PR without merging.

PR #263, opened 2015-06-19
HWKMETRICS-144 Support server-side gauge data aggregate fetch

jshaughn:
- override findGaugeData with a new signature that can apply one or
  more functions to the fetched data (uses a single fetch)
- provide some built in aggregate functions to be used with the new
  findGaugeData signature out-of-box.

### DIFF
--- a/core/metrics-core-api/pom.xml
+++ b/core/metrics-core-api/pom.xml
@@ -52,6 +52,11 @@
     </dependency>
 
     <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxjava-math</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>com.datastax.cassandra</groupId>
       <artifactId>cassandra-driver-core</artifactId>
     </dependency>
--- a/core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
+++ b/core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.metrics.core.api;
+
+import rx.Observable;
+import rx.functions.Func1;
+import rx.observables.MathObservable;
+
+/**
+ * Predefined aggregate functions usable with
+ * {@link MetricsService#findGaugeData(String, MetricId, Long, Long, Func1...).
+ *
+ * @author john sanda
+ * @author jay shaughnessy
+ */
+public interface Aggregate {
+    Func1<Observable<DataPoint<Double>>, Observable<Double>> Average = data -> {
+        return MathObservable.averageDouble(data.map(DataPoint::getValue));
+    };
+
+    Func1<Observable<DataPoint<Double>>, Observable<Double>> Max = data -> {
+        return MathObservable.max(data.map(DataPoint::getValue));
+    };
+
+    Func1<Observable<DataPoint<Double>>, Observable<Double>> Min = data -> {
+        return MathObservable.min(data.map(DataPoint::getValue));
+    };
+
+    Func1<Observable<DataPoint<Double>>, Observable<Double>> Sum = data -> {
+        return MathObservable.sumDouble(data.map(DataPoint::getValue));
+    };
+}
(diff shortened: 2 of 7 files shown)


rhqci:

rhqci:

jsanda:
I will start reviewing today.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
```diff
@@ -0,0 +1,46 @@
...
+import rx.observables.MathObservable;
+
+/**
+ * Predefined aggregate functions usable with
+ * {@link MetricsService#findGaugeData(String, MetricId, Long, Long, Func1...).
+ *
+ * @author john sanda
+ * @author jay shaughnessy
+ */
+public interface Aggregate {
```
tsegismont: For constants grouping, could you use a class with a private constructor instead of an interface? Or an enum.

core/metrics-core-api/pom.xml
```diff
@@ -52,6 +52,11 @@
     </dependency>
 
     <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxjava-math</artifactId>
```
jsanda: I know we need this in core-impl, but do we need it in core-api as well?

stefannegrea:
The new functionality is not exposed via the REST api. Any reasons for that?

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
jshaughn: I had it as an enum but jsanda encouraged a change. So now it is an interface.  I suggest not blocking the PR on this style point.  Please feel free to change it to your liking after the merge, I'm fine with a change.

core/metrics-core-api/pom.xml
jshaughn: We do because Aggregate is in the API and the the predefined aggregates use MathObservable.

jshaughn:
Three reasons.  I didn't need it in REST.  I wanted to make this task as small as possible.  I didn't want to work on REST only to see this get rejected or modified in a substantial way.  But mostly just reason #1.  I see REST as a follow-on enhancement.

jsanda:
> Three reasons. I didn't need it in REST. I wanted to make this task as small as possible.
+1

Any REST API changes can and should be addressed separately. People are much more opinionated about the REST APIs in general, so we probably should have some discussion to see how we want to proceed on that front. There was already some discussion on IRC before Jay started on this work.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
jsanda: I have started writing a reply on this and changed my thinking 3 times now. I do not know what makes the most sense or if it makes a difference, but I do agree that we do not need to hold up merging because of it.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/Aggregate.java
```diff
@@ -0,0 +1,46 @@
...
+    Func1<Observable<DataPoint<Double>>, Observable<Double>> Average = data -> {
+        return MathObservable.averageDouble(data.map(DataPoint::getValue));
+    };
+
```
jsanda: Minor nit-pick... These could be written more succinctly as,

```java
Func1<Observable<DataPoint<Double>>, Observable<Double>> Average = data ->
    MathObservable.averageDouble(data.map(DataPoint::getValue))
```
No need for the curly braces and ```return```

core/metrics-core-impl/pom.xml
```diff
@@ -66,6 +66,11 @@
     </dependency>
 
     <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxjava-math</artifactId>
+    </dependency>
```
jsanda: We don't need this here since it is already in core-api.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -490,6 +490,22 @@ void setDataAccess(DataAccess dataAccess) {
...
+        Observable<T> result = null;
+
+        for (Func1<Observable<DataPoint<Double>>, Observable<T>> func : funcs) {
+            result = (null == result) ?
+                    func.call(dataCache) :
+                    result.concatWith(func.call(dataCache));
+        }
+
+        return result;
+    }
```
jsanda: This method can be written more simply as,

```java
@Override
public <T> Observable<T> findGaugeData(String tenantId, MetricId id, Long start, Long end,
    Func1<Observable<DataPoint<Double>>, Observable<T>>... funcs) {

    Observable<DataPoint<Double>> dataCache = findGaugeData(tenantId, id, start, end).cache();
    return Observable.from(funcs).flatMap(fn -> fn.call(dataCache));
}
```
Aside from making the code more tidy, there is another good reason for the change. We can easily parallelize this with,

```java
return Observable.from(funcs).flatMap(fn -> fn.call(dataCache)).subscribeOn(Schedulers.compute());
```
The convention with a lot of the RxJava APIs is two have two versions of a method, one of which takes a Scheduler.

jshaughn: If you parallelize you may risk losing the ordering of returned data. Any change should be careful of the contract.

jsanda: Excellent point. Only ```flatMap``` call is parallel because of the ```subscribeOn``` call. This would have certainly have to documented in the version of the method that takes the Scheduler argument. The additional method is not necessary for the PR.

stefannegrea:
Thanks @jshaughn !

PR #264, opened 2015-06-19
HWKMETRICS-146 Add a startup signature that does not affect the schema

jshaughn:


rhqci:

rhqci:

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -142,12 +142,18 @@ public int hashCode() {
...
+
+    public void startUp(Session session, String keyspace, boolean resetDb, boolean createSchema) {
         SchemaManager schemaManager = new SchemaManager(session);
         if (resetDb) {
             schemaManager.dropKeyspace(keyspace);
         }
-        // This creates/updates the keyspace + tables if needed
-        schemaManager.createSchema(keyspace);
+        if (createSchema) {
+            // This creates/updates the keyspace + tables if needed
```
tsegismont: Why do we need this new boolean? I thought `#createSchema(String keyspace)` was able to determine if the schema actually needs to be created.

jshaughn: Mutliple session startups were somehow stepping on each other.  I'd be happy actually if this was re-written in a way that any client calling startUp could create the schema if needed.  Or even better, the schema should not be created by a client startup but perhaps an actual installer.  Anyway, this solved my startup issue and so I need it.  I do encourage you guys to make it better.

jshaughn: It may be that this change  is not required given other startup changes.  I don't really see how it hurts anything, though.

stefannegrea:
The long-term goal is to never have to access the internals of Hawkular Metrics directly. This change is good if it helps consuming projects in the short-term.

stefannegrea:
@jshaughn  thanks for the change!

jsanda:
> The long-term goal is to never have to access the internals of Hawkular Metrics directly

I respectfully and strongly disagree with this :) As we start implementing some of the "glue" code the REST API will not be involved. There are use cases for deploying Hawkular Metrics independent of the full, integrated Hawkular server. Some of those same use cases may extend to Hawkular Alerts as well. In all likelihood they would run in the same server in which case we won't be dealing with the REST API. In addition to those examples, I am not convinced that HTTP/REST is the best for really high write throughput scenarios. Maybe at some point we add support for other protocols. I could provide more examples, but I think you get the idea :)

stefannegrea:
@jsanda, I was just hoping that creating sessions from scratch would be avoided by the glue code. I agree with the rest of your assessment.

jsanda:
Ah, yeah I agree about session creation. I fear that is going to be a more difficult problem to solve with each of the services deployed in stand-alone WARs. Might be easier if they are in an EAR, but I am not 100% certain. That's a separate discussion though.

PR #265, opened 2015-06-19
Counter rates

jsanda:
I think this is ready for review. I will create separate tickets for incorporating counters into the REST API. I also need to create additional tickets for making the rates configurable. For example, the interval and window of the rate calculation task is currently fixed/hard coded.


stefannegrea:
Thanks @jsanda !

PR #270, opened 2015-06-22
List metrics endpoint fails with 500 Internal Server Error when "type" param is missing

tsegismont:
Ready for review

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -19,6 +19,7 @@ package org.hawkular.metrics.api.jaxrs.handler;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 import static org.hawkular.metrics.api.jaxrs.filter.TenantFilter.TENANT_HEADER_NAME;
+import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.badRequest;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.executeAsync;
 
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


burmanm:
A test that checks for 400 so that this bug isn't accidentially repeated later on?

tsegismont:
Le 22/06/2015 19:07, Michael Burman a écrit :
> A test that checks for 400 so that this bug isn't accidentially repeated
> later on?

Sure, but I won't be able to add it before the release. I can implement 
it tomorrow. Would you mind to file a JIRA?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -70,33 +71,32 @@
...
-    @ApiResponses(value = { @ApiResponse(code = 200, message = "Successfully retrieved at least one metric "
-            + "definition."),
-            @ApiResponse(code = 204, message = "No metrics found."),
-            @ApiResponse(code = 400, message = "Given type is not a valid type.", response = ApiError.class),
+                  response = List.class, responseContainer = "List")
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Successfully retrieved at least one metric definition.",
+                         response = ApiError.class),
+            @ApiResponse(code = 204, message = "No metrics found.",
+                         response = ApiError.class),
```
burmanm: 204 should not return any response actually. Is this addition for some weird swagger behavior?

jsanda: > 204 should not return any response actually.

Do you mean that there should be no response body?

jsanda:
The test would be nice, otherwise LGTM.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
tsegismont: Le 22/06/2015 19:15, Michael Burman a écrit :
> 204 should not return any response actually. Is this addition for some
> weird swagger behavior?

Right, I messed this up. It's fixed now.

tsegismont: Le 22/06/2015 19:21, jsanda a écrit :
> Do you mean that there should be no response body?

Yes. Have a look at `ApiUtils#noContent`.

Actually, it behaves like this since the beginning. I think it's the 
right thing to do, it's not an error, it's absence of result.

tsegismont:
Le 22/06/2015 19:40, jsanda a écrit :
> The test would be nice

Yep. Micke has filed (HWKMETRICS-155) Test against regression for metric 
definition list status code 400

Will do it tomorrow.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
jsanda: I assumed it meant no response body but wasn’t 100% certain. Thanks for the clarification.

tsegismont:
I've rebased the PR on master and added the itest (tracked by HWKMETRICS-155)

PR #272, opened 2015-06-23
Add endpoint for retrieving rates

jsanda:
This PR also integrates the task scheduling service into the jaxrs WAR so that rates get calculated and updated for counter metrics.


jsanda:
The tests were commented out while I was debugging. I have uncommented them and added some javadocs.

PR #274, opened 2015-06-26
HWKMETRICS-157 Implement PTrans on top of Vert.x3

tsegismont:
Please do not merge yet. I'm done with the code changes but still working on an enhanced version of MetricsSenderITest.

Reproducing the JIRA description here:

>PTrans is currently based on vanilla Netty.

>While using the raw toolkit gives a lot of freedom, it's not very easy to program correctly. Especially, it needs much thinking when passing around data from a server to the forwarder.
>Also, writing effective batcher and spool is hard. The easiest way to make it work with all the threads involved is to use the synchronized BoundMetricFifo, but this defeats the initial goal of relying on a high-performance, non blocking toolkit.

>With vertx.3 on the other hand, we can write servers as verticles and the sender as another verticle. They exchange data through the event bus. As every verticle is assigned a specific context, and thus a single event loop, we don't need to care about synchronization issues. So the MetricSender can work with a non blocking FIFO.

>The expectation with this change is to avoid making mistakes which you can easily do with raw Netty and make the PTrans programming model easier.


tsegismont:
The PR is now ready for review.

tsegismont:
Travis build failed due to OOM. Restarted the build

tsegismont:
Can someone please review this PR?

PR #275, opened 2015-06-29
HWKMETRICS-160 Fix Intermittent test failures in GraphiteITest

tsegismont:
HWKMETRICS-160 Fix Intermittent test failures in GraphiteITest

When EmbeddedJMXTrans is stopped, it collects and sends metrics one last time after the schedulers have
been shut down. So two data points can have very close timestamps. But Graphite only supports second
resolution. So it is possible that two points with the same timestamp (see #getTimestamp logic) are sent
to the server. Consequently, when building the expected data set, we need to check if one metric has two
points with the same timestamp, and, in this case, keep the older one only.


tsegismont:
Can someone please review this PR?

PR #277, opened 2015-06-30
Hwkmetrics 137

mwringe:
Add support for heapster to the kubernetes.json file and add in simple tests to make sure that heapster is able to write metrics into hawkular.

NOTE: this patch should be ready to be committed, except that it relies on a snapshot version of fabric8. Once fabric8 has its next release and the dependency updated to a non-snaphshot version, then this PR will be ready to be pulled.


mwringe:
Ok, this should be good now. If anyone wants to check and then merge the PR that would be greatly appreciated.

tsegismont:
Thanks Matt!

PR #280, opened 2015-07-07
[HWKMETRICS-20] [HWKMETRICS-171] Ability to fetch all the metric definitions, tags and type as filters

burmanm:
[HWKMETRICS-20] [HWKMETRICS-171] Allow / to fetch all the metric definitions, use tags and type as filters. Create new methods for fetching with tags.

Changes signature of fetching metric definition, missing type is no longer considered an error (and no code 400 is returned), instead all the metric definitions are returned for the tenant.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -17,6 +17,7 @@
 package org.hawkular.metrics.api.jaxrs.handler;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
 import static org.hawkular.metrics.api.jaxrs.filter.TenantFilter.TENANT_HEADER_NAME;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.badRequest;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.requestToAvailabilities;
@@ -43,9 +44,11 @@ import com.wordnik.swagger.annotations.ApiParam;
 import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 import org.hawkular.metrics.api.jaxrs.ApiError;
+import org.hawkular.metrics.api.jaxrs.param.Tags;
 import org.hawkular.metrics.api.jaxrs.request.MetricDefinition;
 import org.hawkular.metrics.api.jaxrs.request.MixedMetricsRequest;
 import org.hawkular.metrics.api.jaxrs.util.ApiUtils;
+import org.hawkular.metrics.core.api.Metric;
 import org.hawkular.metrics.core.api.MetricType;
 import org.hawkular.metrics.core.api.MetricsService;
 import rx.Observable;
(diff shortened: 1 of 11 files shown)


mtho11:
No 400, yes!

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -74,26 +77,31 @@
...
-                      required = true,
+                      required = false,
                       allowableValues = "[gauge, availability, counter]")
-            @QueryParam("type") MetricType metricType
-    ) {
-        if (metricType == null) {
-            asyncResponse.resume(badRequest(new ApiError("Missing type param")));
+            @QueryParam("type") MetricType metricType,
+            @ApiParam(value = "List of tags", required = false) @QueryParam ("tags") Tags tags) {
+        if(metricType != null && !MetricType.userTypes().stream().filter(t -> metricType == t)
```
tsegismont: How about `userTypes().contains(metricType)` ? This stream+filter+findAny construct looks a bit convoluted in this case.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricsService.java
```diff
@@ -84,6 +84,15 @@
 
     Observable<Metric> findMetrics(String tenantId, MetricType type);
```
tsegismont: Can you update the Javadoc to explain the new behavior?

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricsService.java
```diff
@@ -84,6 +84,15 @@
...
 
+    /**
+     *
+     * @param tenantId
+     * @param tags
+     * @param type Optional MetricType, if null is given all matching metrics are returned
```
tsegismont: Shouldn't this be "if null is given, all matching user type metrics are returned"?

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricType.java
```diff
@@ -89,4 +90,8 @@ public static MetricType fromTextCode(String textCode) {
         return type;
     }
 
+    public static List<MetricType> userTypes() {
+        return Arrays.asList(Arrays.copyOf(MetricType.values(), 3));
```
tsegismont: How about creating a static final `EnumSet` instead of creating a list on each call?

core/metrics-core-api/src/test/java/org/hawkular/metrics/core/api/MetricTypeTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import org.junit.Test;
+
+/**
+ * @author miburman
+ */
+public class MetricTypeTest {
```
tsegismont: :+1: Nice to have a test to check user type set is not broken

core/metrics-core-api/src/test/java/org/hawkular/metrics/core/api/MetricTypeTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    public void testUserValues() {
+        MetricType.userTypes().stream().filter(metricType -> metricType == MetricType.COUNTER_RATE).forEach
```
tsegismont: Same here, I find `assertFalse(MetricType.userTypes().contains(COUNTER_RATE))` more readable

tsegismont:
There's a test failure in `TagsITest.findDefinitionsWithTags`, can you check?
https://travis-ci.org/hawkular/hawkular-metrics/builds/69950074#L1871-L1874

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricsService.java
burmanm: No, as the tags can further filter the results.

tsegismont: Le 08/07/2015 11:55, Michael Burman a écrit :
> No, as the tags can further filter the results.

Understood, I had a closer look at MetricServiceImpl:
https://github.com/hawkular/hawkular-metrics/pull/280/files#diff-04e1ad2ad2cadc473893730a61b3e271R475

But then why return user types only when the tags query param is null? 
Or maybe I missed something?

burmanm: I'm not sure what you mean actually. If a null is given, all the userTypes are used (the filter doesn't stop anything) but if a certain type is given, that one is used (and only metrics with that type are allowed to pass). There's a bug in that sense that this allow setting non-user-defined type also if you can bypass the REST-interface..

tsegismont: My understanding is as follows.

If no type is set, then #findMetrics returns all metrics which type is a 
user type.

If no type is set, then #findMetricsWithTags returns all metrics having 
the specified tags, regardless of the type.

If I understood well, then I think that if no type is set, 
#findMetricsWithTags should return all metrics having the specified 
tags, and which type is a user type.

tsegismont:
Thanks @burmanm !

PR #281, opened 2015-07-07
[HWKMETRICS-172] Add checks for empty or missing payloads for various…

stefannegrea:
… types to allow users to omit them from the request. Also, added tests for various scenarios.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -19,6 +19,7 @@ package org.hawkular.metrics.api.jaxrs.handler;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.hawkular.metrics.api.jaxrs.filter.TenantFilter.TENANT_HEADER_NAME;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.badRequest;
+import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.emptyPayload;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.requestToAvailabilities;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.requestToCounters;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.requestToGauges;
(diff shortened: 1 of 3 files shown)


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -101,20 +107,34 @@ public void findMetrics(
     @ApiOperation(value = "Add data for multiple metrics in a single call.")
     @ApiResponses(value = {
             @ApiResponse(code = 200, message = "Adding data succeeded."),
+            @ApiResponse(code = 400, message = "Missing or invalid payload.", response = ApiError.class),
```
tsegismont: I couldn't find the code checking for an empty payload. Did I miss it?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/request/MixedMetricsRequest.java
```diff
@@ -42,15 +42,27 @@
...
     }
 
     public List<Counter> getCounters() {
-        return ImmutableList.copyOf(counters);
+        if (counters != null) {
+            return ImmutableList.copyOf(counters);
+        }
+
+        return null;
     }
```
tsegismont: Wondering if we shouldn't simply drop the `ImmutableList.copyOf` and thus the new null check, as this a mutable object on purpose (representation of the mixed metrics entity) and not used anywhere else than in the handler code.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/MetricsITest.groovy
```diff
@@ -150,4 +150,189 @@ class MetricsITest extends RESTTest {
...
+        path: "metrics/data",
+        headers: [(tenantHeaderName): tenantId],
+        body: [
+            gauges: [
+            ],
+            counters: [
+            ]
+        ]
+    )
+    assertEquals(200, response.status)
```
tsegismont: Shouldn't the status code be 400? Gauge handler replies with 400 when the list of metrics is empty.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -101,20 +107,34 @@ public void findMetrics(
...
             @Suspended final AsyncResponse asyncResponse,
             @ApiParam(value = "List of metrics", required = true) MixedMetricsRequest metricsRequest) {
 
-        Observable<Void> gaugeInserts = metricsService.addGaugeData(requestToGauges(tenantId,
-                metricsRequest.getGaugeMetrics()).subscribeOn(Schedulers.computation()));
-        Observable<Void> counterInserts = metricsService.addCounterData(requestToCounters(tenantId,
-                metricsRequest.getCounters()).subscribeOn(Schedulers.computation()));
-        Observable<Void> availabilityInserts = metricsService.addAvailabilityData(requestToAvailabilities(tenantId,
-                metricsRequest.getAvailabilityMetrics()).subscribeOn(Schedulers.computation()));
+        Collection<Observable<Void>> observables = new ArrayList<>();
```
burmanm: This isn't very reactive design and actually a bit more difficult to follow what's happening. Having empty observables is not an issue, they are just not processed (nothing to react to).

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
tsegismont: Now I understand what's going on. In `MetricsITest#addMixedDataEmptyRequestPayload`, you set the body to `[]`. But this is not an empty object, this is an empty array. So Jackson fails to deserialize from an array, throws an exception, converted by RestEasy into a 400 reply. 

What you need to do to test the empty object case is `body: new HashMap<>()`.

Also there's the null test case which is missing. And if you add it, you'll get a 500 reply, because of an NPE thrown in this method. You need to check if `metricsRequest` is null and return `emptyPayload` in this case.

tsegismont:
Stefan are your going to update all `@POST` endpoints in this PR? Or other PRs?

stefannegrea:
@tsegismont, I will create another Jira/PR to update the rest of the @POST methods. I want to keep this PR only for the newly updated method.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
stefannegrea: I updated the code and test cases to check for empty payload early and return 400 if no content.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
stefannegrea: Agree, I updated the code to preempt executing that code when there is no content.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/MetricsITest.groovy
stefannegrea: You are right, we should treat empty content consistently and return 400 even though the named arrays are there but empty. Updated the code to reflect this.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/request/MixedMetricsRequest.java
stefannegrea: I dropped the immutable portion out of the request object and closer to the metrics service calls. That should help checking for null and empty content without incurring the copy penalty.

tsegismont:
Ha! The build failed again because of the `CountersITest` intermittent failure:
```
Failed tests: 
  CountersITest.findRate:288 expected:<200> but was:<204>
```

I'll merge the PR anyway as the other tests are passing.

PR #283, opened 2015-07-08
HWKMETRICS-175 Put integration tests in profiles

tsegismont:
Don't run integration tests by default.

Of course they will still have to run on Travis.

Each project with integration tests has its own profile, so that one can build the sources and run selected integration tests:
- "rests-tests-integration-tests"
- "ptrans-integration-tests"
- "core-impl-tests-integration-tests"
- "task-queue-integration-tests"

All these profile are activated if "-Dintegration-tests" is added to the build command.

In rests-tests and ptrans modules, developers can still use the "-Drunning.service" flag to skip Wildfly start/deploy/stop cycle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_install:
 install:
 - mvn -version -B
 script:
-- mvn verify -Dwildfly.logging.console.level=DEBUG -B
+- mvn verify -Dintegration-tests -Dwildfly.logging.console.level=DEBUG -B
 after_failure: bash .travis.diagnose.sh
 after_success:
 - PROJECT_VERSION=`mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep -v '\['`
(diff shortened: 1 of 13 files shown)


tsegismont:
Recently the project build time has been growing a lot. While it's a good sign that more and more itests are added, it can be annoying as the only way to make a build faster is too disable all tests, unit tests included.
Also, putting integration tests in profiles removes the requirement of a running C* cluster to build the project (without skipping tests).
Note that developers who want to activate them all the time on their box can update their settings file as documented here:
https://maven.apache.org/examples/injecting-properties-via-settings.html

jsanda:
I can appreciate wanting to keep the build fast; however, the integration tests provide the bulk of the test coverage. Disabling them by default is basically the same as skipping tests IMO.

tsegismont:
There was an agreement today during the team meeting to keep running itests by default, but to put them in profiles so that a developer can deactivate them on demand. The second commit reflects that decision.
The PR has been rebased on master.

PR #284, opened 2015-07-09
HWKMETRICS-176 REST: make sure all POST requests with empty body return 400 Empty payload

tsegismont:
Updates handlers and added tests. Also made the behavior consistent across endpoints.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/exception/mappers/ReaderExceptionMapper.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/exception/mappers/ReaderExceptionMapper.java
@@ -20,7 +20,11 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 
+import org.hawkular.metrics.api.jaxrs.interceptor.EmptyPayloadException;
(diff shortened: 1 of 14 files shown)


burmanm:
Like discussed in the IRC, I would go for a filter that would reject all the "POST"-requests that have empty payload. Instead, if we want to set something empty, the semantics would be using "PUT" or "DELETE" instead.

tsegismont:
Yeah, I had started to write an interceptor and will resurrect it. It 
will check if body is null. I also had in mind to make sure that a List 
body or Map body is not empty. What do you think?

pilhuhn:
Couldn't that live in the handler code? Check for body == null or body.length==0 should not be too hard and it is much more obvious than in a filter or interceptor

tsegismont:
Le 10/07/2015 11:53, Heiko W. Rupp a écrit :
> Couldn't that live in the handler code? Check for body == null or
> body.length==0 should not be too hard and it is much more obvious than
> in a filter or interceptor

I started with the interceptor and did not push it for this reason. But 
it's true experience has shown it's wasn't always implemented or nor 
consistent across endpoints. Interceptor can help here.

tsegismont:
I've pushed two commits which introduce automated null and empty collection/map checking. Now everyone can get a feeling of what the code looks like.

tsegismont:
Rebased on master. @burmanm would you mind to finish the review (last two commits)?

PR #290, opened 2015-07-20
Some small fixes to serializable, switch and finalizers

burmanm:


tsegismont:
Why do we need those classes to be Serializable?

tsegismont:
@burmanm can we close this one? Or is there a good reason to make those Serializable?

PR #293, opened 2015-07-27
[HWKMETRICS-114] Refactor MetricId to include tenantId and MetricType

burmanm:
This PR moves the tenantId and MetricType definitions from Metric to MetricId. Metric's getTenantId()  and getType() are marked as deprecated and will return data from the assigned MetricId.


mwringe:
Looks good to me, merged.

PR #297, opened 2015-08-03
Task scheduler refactoring

jsanda:
There is still a bunch of work to do, some of which should be done in separate tickets that I will create. I want to start the review process because we need to move things along.


pom.xml
```diff
@@ -117,7 +117,7 @@
     <slf4j.version>1.7.7</slf4j.version>
     <log4j.version>1.2.17</log4j.version>
     <antlr.version>4.5</antlr.version>
-    <rxjava.version>1.0.10</rxjava.version>
+    <rxjava.version>1.0.12</rxjava.version>
```
burmanm: 1.0.13 is the newest, might be wise to go there directly (if we're upgrading anyway) ?

task-queue/src/main/java/org/hawkular/metrics/tasks/api/RepeatingTrigger.java
```diff
@@ -0,0 +1,185 @@
...
+        next.interval = interval;
+        next.delay = delay;
+        next.triggerTime = triggerTime + interval;
+        next.repeatCount = repeatCount;
+        next.executionCount = executionCount + 1;
+
+        return next;
+    }
+
+    private DateTime getExecutionDateTime() {
```
burmanm: Each Trigger implementation redefines this method some with static, some with non-static modifiers? Would interface default implementation be suitable here?

task-queue/src/main/java/org/hawkular/metrics/tasks/impl/TaskSchedulerImpl.java
```diff
@@ -0,0 +1,626 @@
...
+                        getQueue(lease)
+                                .observeOn(tasksScheduler)
+                                .groupBy(Task2Impl::getGroupKey)
+                                .flatMap(group -> group.flatMap(this::execute).map(this::rescheduleTask))
+                                .subscribe(
+                                        task -> logger.debug("Finished executing {}", task),
+                                        t -> logger.warn("There was an error observing tasks", t),
+                                        () -> {
+                                            Date timeSlice = new Date(lease.getTimeSlice());
+                                            // TODO We need error handling here
```
burmanm: Would retryWhen work here?

task-queue/src/main/java/org/hawkular/metrics/tasks/api/RepeatingTrigger.java
jsanda: I'm not sure about making it a default method because it is not part of the public API. It really is just an internal helper method. For now I have introduced an abstract base class and moved the method there to eliminate the duplication. I have also moved RepeatingTrigger.now into AbstractTrigger. It is used for testing with a virtual clock. 

The getExecutionTime method is used in several places. It was basically copied from metrics-core-impl. The task scheduler really should not depend on the core though. Maybe there needs to be some date utils module in hawkular-commons.

task-queue/src/main/java/org/hawkular/metrics/tasks/impl/TaskSchedulerImpl.java
jsanda: What specifically are you referring to with retryWhen? Note that I created [HWKMETRICS-201](https://issues.jboss.org/browse/HWKMETRICS-201) which will have a big impact on the implementation; however, I am not sure if it deals with what you are referring to specifically.

burmanm: Well, it wasn't about that blocking behavior, but instead of error handling. Documenting the potential cases where errors could happen would be important, eg. "we couldn't write to the Cassandra - add .." or "there's potential for large backpressure here" (which the HWKMETRICS-201 does sort of address but as a general rule of thumb).

Because this isn't in use yet, I don't have otherwise issues in merging it - tests that fail verify the blocking behavior would be nice though as they would help in refactoring and to verify that the order is correct always.

jsanda: > Well, it wasn't about that blocking behavior, but instead of error handling

Gotcha. Robust error handling is definitely needed. ```executeMultipleTasksFromMultipleGroupsInDifferentQueues``` in TaskSchedulerITest does verify that tasks execute in the correct order, but only for a single time slice. I can add another test that verifies execution order for multiple time slices. I can also include long running tasks that cause scheduling delays. 

>Because this isn't in use yet...

This is in use for counter rates. The counter rates REST integration test is disabled in this PR. The work I have done for [HWKMETRICS-199](https://issues.jboss.org/browse/HWKMETRICS-199) in the [hwkmetrics-199 branch](https://github.com/hawkular/hawkular-metrics/tree/hwkmetrics-199) re-enables the test with support for the virtual clock in REST API tests.

jsanda:
I added a new test, ```executeTasksFromMultipleGroupsMultipleTimes```, that verifies execution order across multiple time slices when there is a long running task. The test creates six tasks that are divided into two groups which are all scheduled to execute every minute. One of the tasks takes 3 minutes to complete. If the ```CountDownLatch.await()``` call is removed the test fails.

I think that the reason other tests might not have failed without the CountDownLatch is because of  bug in the ```TaskSchedulerImpl.findAvailableLeases``` method which is fixed in commit 4467d6f.

PR #299, opened 2015-08-04
RxJava migration of buckets computation code

tsegismont:
Since many metrics are now counters in Hawkular, counters endpoint need to support bucketed output.


tsegismont:
Not ready for merging. I'm just sharing the initial work: migration of the existing code (gauges and avail) to RxJava.
The approach I've taken is to create an operator which will be applied on the source observable with `#lift`. The implementation is inspired by `#buffer` methods.

@jsanda can you have a quick look ? I'd like to know if I should pursue tomorrow or completely different approach.

jsanda:
Sure, I will take a look. I have to be honest though. I am skeptical about using a custom operator. Custom operators are a very advanced feature that requires in depth understand and experience. I for one would not feel confident in trying to implement a custom operator. How does your operator deal with back pressure? How does it manage subscriptions? If you aren't sure how to answer questions like these, then a custom operator is not the approach :) The existing operators provide a lot of functionality. I only have an understanding of a handful at this point.

jsanda:
At a high level in terms of making it more reactive/functional, think about what operators you might use instead of the for and while loops. And instead of accepting a list of data points, consider take an Observable of data points instead.

burmanm:
There's no need to create a BucketedOutputOperator for example, it could be achieved by using scan/buffer(bufferClosingSelector)/skipWhile functions (no need for all of them, but different methods exists).

jsanda:
Do we need buffering operations for either mapper? Seems like they should be a reduce operation in each case.

tsegismont:
@jsanda @burmanm could you please have a look at the latest commit (6bb3d37280c977d0ad69af4da0a71452f0af8b95)?

Copying here the commit message.

*Gauges buckets computation without custom operator*

This is to show what the process could look like, code for avail has not
been updated, and code for counters not implemented.

In this attempt, no custom operator is used. Instead, N(=number of
buckets) requests for data are sent in parallel. Then the resulting
observables are flat mapped to form an observable of bucket points.

The bucket point observable emits just one value. The value is created
by processing data points as they arrive, instead of buffering
all the points for the bucket in an array.
This has a huge implication on stats computation. Some stats algos are
"storeless", which means you can get the exact result without buffering
the data. Examples are min, max, avg. But some are not, like percentile.

As a consequence, for median and 95th percentile, we need to use
an approximative algorithm, provided by
org.apache.commons.math3.stat.descriptive.rank.PSquarePercentile.
See http://www.cse.wustl.edu/~jain/papers/psqr.htm for a description of
the algorithm.
On the upside, this algorithm allows to compute stats when the number of
data points in the bucket is VERY large, with low memory consumption.
On the downside, the result might look a bit inaccurate when there VERY
few points in the buckets. It's also sensible to order of value
submission.

Note that the observable of bucket points will often be incorrectly
ordered, as there's no guarantee in the order the responses will be processed.

Side work:
- log throwable at TRACE level for 500 responses
- overloaded assertDoubleEquals so that caller can pass a failure
  message

tsegismont:
Following the different comments here and the conversations with @burmanm on IRC, I've come to a simple (no custom operator), readable (no observable creation), non buffering solution:

https://github.com/hawkular/hawkular-metrics/pull/299/files#diff-04e1ad2ad2cadc473893730a61b3e271R612

@jsanda @burmanm can you tell me if this is ok?

There could still be a slight change if @burmanm teaches me how to use `#window(closingSelector)` instead of `#groupBy`, because I failed to understand the mechanism.

If we agree on the solution, I'll apply the logic to avail and counters, and remove old mapper classes.

Guys, I'm looking forward to hearing from both of you on the "storeless" stat functions topic. Pasting again the issue here.

>The bucket point is created by processing data points as they arrive, instead of buffering
all the points for the bucket in an array.
This has a huge implication on stats computation. Some stats algos are "storeless", which means you can get the exact result without buffering the data. Examples are min, max, avg. But some are not. Like percentile.

>As a consequence, for median and 95th percentile, we need to use an approximating algorithm, provided by `org.apache.commons.math3.stat.descriptive.rank.PSquarePercentile`. See http://www.cse.wustl.edu/~jain/papers/psqr.htm for a description of the algorithm.

>On the upside, this algorithm allows to compute percentiles when the number of data points in the bucket is VERY large, with low memory consumption.
On the downside, the result might look a bit inaccurate when there are VERY few points in the buckets. It's also sensible to order of value submission.

burmanm:
To me using the stateless algorithms sounds good as it would support some of the ideas of streaming implementation in the future.

tsegismont:
>To me using the stateless algorithms sounds good as it would support some of the ideas of streaming implementation in the future.

Agreed. The drawback is that for testing, we can't apply simple formulas on a small data set.

jsanda:
I will spend time on this tomorrow. This is a good opportunity for us all to learn more about RxJava. I can start looking at it tonight if I can resolve some merge/rebase issues.

tsegismont:
Closed this PR and moved it to #299

PR #302, opened 2015-08-07
HWKMETRICS-206 Implementation of statistics computation based on RxJava

tsegismont:
Gauge and avail bucket computation code was still based on point buffering code. This PR migrates this piece of code to RxJava.
The changes should help when a *lot* of data is involved. For example, if a user requests monthly stats for a year of data, on a metric collected every second.


tsegismont:
@jsanda I've talked with @burmanm on IRC this morning, the `groupBy` solution is ok.

I wanted to add an itest to show that the algo was good for the use case presented in the PR description, but I got a weird error when posting many points in one request (something related to C* batches). I'll work on a test case for that separate issue next week and will file another JIRA.

jsanda:
I like the ```groupBy`` approach and using the DataPointCollector classes with ```collect```. Nicely done!

tsegismont:
Le 10/08/2015 21:12, jsanda a écrit :
> I like the |groupBy`` approach and using the DataPointCollector classes
> with|collect```. Nicely done!

Thanks

PR #303, opened 2015-08-08
[HWKMETRIC-207] clean up MetricsServiceImpl.createTenant method

jsanda:


core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -103,7 +100,7 @@
         private final MetricId metricId;
 
         public DataRetentionKey(String tenantId, MetricType type) {
-            metricId = new MetricId(tenantId, type, "[" + type.getText() + "]");
+            metricId = new MetricId(tenantId, type, "$" + type.getText());
```
burmanm: If $ is suddenly a magic key (repeated in several places), shouldn't it be a some sort of constant instead of magic-char in the strings?

jsanda: Yes it should. And we probably need a separate ticket along with updated documentation to formally introduce reserved characters.

core/metrics-core-impl/src/test/java/org/hawkular/metrics/core/impl/DelegatingDataAccess.java
```diff
@@ -20,18 +20,15 @@
 import java.util.Map;
 import java.util.Set;
 
+import com.datastax.driver.core.ResultSet;
```
burmanm: Another thing, update your hawkular-build-tools and reapply the settings.jar to get correct imports. They're messed up in every committed file.

jsanda:
I updated my IDE settings, and I added a function for wrapping a string with a reserved character. I created [HWKMETRICS-208](https://issues.jboss.org/browse/HWKMETRICS-208) for fully implemented reserved characters.

PR #304, opened 2015-08-11
HWKMETRICS-209

mwringe:
Remove the EC ciphers since OpenJDK doesn't currently support them properly.


stefannegrea:
Thanks @mwringe !

PR #307, opened 2015-08-13
rest-tests minor changes

tsegismont:
The main change is in the RESTClient failure handler implementation.

There used to be a stranged exception thrown when an error response was unexpected.

Now the status line and body (if present) are printed, and tests methods assertions are executed (like assert response status == 200)


stefannegrea:
Thanks @tsegismont !

PR #309, opened 2015-08-14
HWKMETRICS-215 IllegalStateException: Batch statement cannot contain more than 65535 statements

tsegismont:
C* batch size is limited to 0xFFFF. This fix lets the user submit a payload of arbitrary size.

### DIFF
--- a/core/metrics-core-impl/pom.xml
+++ b/core/metrics-core-impl/pom.xml
@@ -113,6 +113,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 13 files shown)


tsegismont:
I just pushed a commit to extract a `BatchStatementTransformer` class so that it can be re-used. I guess we should pay attention to that everywhere we use BatchStatement.

jsanda:
I think the changes look good. While you are at it, could you also refactor ```DataAccessImpl.updateMetricsIndexRx``` to use the batch transformer.

tsegismont:
I'm doing this right now :) In fact, I'm also doing it everywhere it makes sense.

tsegismont:
@jsanda done. Ready for final review.

I've also added a trivial regression test for the batcher and removed the unused, Futures-based `DataAccessImpl#updateMetricsIndex` method.

jsanda:
retest this please

PR #310, opened 2015-08-18
Use correct long argument form in containers README

spadgett:
Long arguments to OpenShift commands start with two dashes, `--`.

@mwringe


rhqci:

PR #311, opened 2015-08-18
[HWKMETRICS-204] Implicit tenant creation

jsanda:


jsanda:
I have provided a brief summary of the changes in the ticket [here](https://issues.jboss.org/browse/HWKMETRICS-204?focusedCommentId=13099563&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13099563). After the review is complete, I will create a squash commit prior to merging.

jsanda:
retest this please

stefannegrea:
retest this please

jsanda:
retest this please

jsanda:
retest this please

tsegismont:
retest this please

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -315,6 +352,7 @@ private void stopMetricsService() {
         state = State.STOPPING;
         metricsService.shutdown();
         taskScheduler.shutdown();
+        jobs.values().forEach(Subscription::unsubscribe);
```
tsegismont: Shouldn't this be ported to the JAX-RS 1.1 impl?

.travis.yml
```diff
@@ -43,7 +43,7 @@ before_install:
 install:
 - mvn -version -B
 script:
-- mvn verify -Dwildfly.logging.console.level=DEBUG -Djboss-as.logging.console.level=DEBUG -B
+- mvn verify -Dwildfly.logging.console.level=INFO -Djboss-as.logging.console.level=INFO -B
```
tsegismont: Out of curiosity, was there a problem on Travis? Or is there another reason for change?

tsegismont:
Noted while reviewing: after rebasing, squashing, you'll need to re-org the imports, otherwise the build will fail.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/CreateTenants.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public CreateTenants(TenantsService tenantsService, DataAccess dataAccess) {
+        this.tenantsService = tenantsService;
+        this.dataAccess = dataAccess;
+    }
+
+    @Override
+    public void call(Task2 task) {
+        long bucket = getBucket(task.getTrigger());
+        CountDownLatch latch = new CountDownLatch(1);
```
tsegismont: What's the reason for blocking the task executor thread until an error occurs or tenant bucket is deleted?

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/DateTimeService.java
```diff
@@ -29,12 +29,14 @@
  */
 public class DateTimeService {
 
+    public Supplier<DateTime> now = DateTime::now;
```
tsegismont: Constants should be static final

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/CreateTenants.java
tsegismont: Ok, found https://issues.jboss.org/browse/HWKMETRICS-214

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -227,6 +231,30 @@ void unloadDataRetentions() {
         dataRetentions.clear();
     }
 
+    private void initSystemTenant() {
+        CountDownLatch latch = new CountDownLatch(1);
+        Trigger trigger = new RepeatingTrigger.Builder().withInterval(30, MINUTES).withDelay(30, MINUTES).build();
+        dataAccess.insertTenant(new Tenant(SYSTEM_TENANT_ID))
+                .filter(ResultSet::wasApplied)
+                .map(row -> taskScheduler.scheduleTask(CreateTenants.TASK_NAME, SYSTEM_TENANT_ID, 100, emptyMap(),
```
tsegismont: Why do we need to schedule this task? Isn't the CreateTenants task only needed if the tenant did not exist?

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -227,6 +231,30 @@ void unloadDataRetentions() {
...
+                        trigger))
+                .subscribe(
+                        task -> logger.debug("Scheduled {}", task),
+                        t -> {
+                            logger.error("Failed to initialize system tenant", t);
```
tsegismont: error -> fatal ?

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -227,6 +231,30 @@ void unloadDataRetentions() {
...
+                            latch.countDown();
```
tsegismont: Shouldn't we stop the startup if the system tenant wasn't created?

tsegismont:
Maybe we should stop restarting the perf tests and start considering the impact of updating the tenant bucket on each POST? How about updating the tenant bucket "asynchronously", meaning, after the HTTP request was replied?

.travis.yml
tsegismont: Ok, just saw commit message "increase log level for travis to avoid 4 MB limit"

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
jsanda: Yes, good catch.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/DateTimeService.java
jsanda: It is not a constant. When testing with the virtual clock it gets set,

```java
dateTimeService.now = () -> new DateTime(virtualClock.now());
```

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
jsanda: Correct. We only need to perform the work done by CreateTenants if a tenant does not exist. If we do not have a scheduled, background job, when would we check? When inserting data points? We could do that but reads before writes, particularly in high write volume scenarios, is an anti-pattern. CreateTenants runs every 30 minutes. The tenants_by_time keeps track of any tenants having data inserted during the past 30 minutes.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
jsanda: Yeah, this should probably cause start up to fail. ```initiSystemTenant``` should probably return an Observable rather than void. I will create a ticket to address this.

jsanda:
>Maybe we should stop restarting the perf tests and start considering the impact of updating the tenant bucket on each POST?

Stefan and I were discussing this late last night. I want to hold off on merging until we do some more investigation. The impact of the tenants_by_time write should be small. It is done asynchronously along with all of the other writes in each of the addXXXData() methods. Of course, the response won't be sent until all writes finish, but they are done in parallel.

I spent time last night doing some stress testing. My results did not indicate a performance issue; however, I would not consider the testing thorough. I wasn't using a clean environment on each test run for example.

The issue might *not* be with the additional write to the tenants_by_time table. We need to better understand what exactly is happening with those perf tests. If for example it is creating 10,000 tenants, then 10,000 rate calculation jobs are getting created. That would introduce a lot more overhead with the task scheduler.

To avoid extra noise in this PR, I will branch off this PR and create a new PR just for testing purposes. I will spend some time investigating, but I do not want the the perf-stability-test job to completely hold up this PR.

jsanda:
retest this please

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
jsanda: I have created [HWKMETRICS-227](https://issues.jboss.org/browse/HWKMETRICS-227) for this.

jsanda:
I did some investigation into the haw-perf-stability-test job failure. The job runs 3 scenarios for writing gauge, availability, and counter data points. The test scenarios aim to test write performance in relative isolation. With this PR, the writes are no longer tested in isolation. With the implicit tenant creation, over 300 tenants are created during the test run. This in turn results in over 300 rate calculation jobs being created and scheduled to execute every minute. This results in a lot of reads being performed that previously were not done. I will work with Filip to update the test scenarios. I think we still want some scenarios to test writes in isolation, and we are also going to need new scenarios that put load on the system through the various background jobs and scheduler.

stefannegrea:
Thanks @jsanda !

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
tsegismont: Couldn't it run just once on startup ?

tsegismont:
I think it is ok to move on if we create a JIRA to evaluate the 
performance regression (low-medium impact).

We need to answers these questions, I think:

What's the impact of the additional C* requests? They are executed in 
parallel, but no response is sent until they are all executed.

How are the test implemented? How many tenants (thus rate computation 
jobs) are created?

Maybe the JIRA will be CLOSED WON'T FIX, because the system was just too 
simple before. But it needs to be created.

tsegismont:
Le 19/08/2015 18:53, jsanda a écrit :
> I think we still want some scenarios to test writes in isolation, and we
> are also going to need new scenarios that put load on the system through
> the various background jobs and scheduler.

Sorry I don't understand. Is it really possible to test writes in 
isolation ? What's the purpose? A real system would always have these 
rate calc jobs, correct?

jsanda:
Yes, a system will always have these jobs; however, the number of tenants has a direct correlation to the task scheduler overhead. More tenants --> more jobs --> more scheduler overhead.

We can and should have a scenario that writes the same amount of data points for the same number of metrics but only using a single tenant. I would expect the performance number to look different than what is being reported here.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
jsanda: It essentially does run only once. 

```java
dataAccess.insertTenant(new Tenant(SYSTEM_TENANT_ID))
    filter(ResultSet::wasApplied)
```

```wasApplied``` will only be true the first successful time. Subsequent attempts will basically be no-ops.

tsegismont: OK. I'm wondering if the system tenant shouldn't be created after schema 
setup and then we would simply fail the deployment if it no longer 
exists. After all, something very bad must have happened if the system 
tenant is no longer present.

PR #313, opened 2015-08-19
HWKMETRICS-212 Eliminate duplicate methods in MetricsService

tsegismont:
Replaced all #addXXXData and #findXXXData with generic #addDataPoints and #findDataPoints method

Note that #findDataPoints needs  a "Class<T> valueClass" argument otherwise the compiler may not be able to infer data points type.

Side work: removed deprecated method calls from core-impl


jsanda:
Can we do better than the approach I suggested in ticket? It has a couple of problems. First, we lose some compile-time safety when inserting data. It would be nice if we could come up with something that does not lose those compile-time checks. 

Secondly, querying data points requires specifying the value class. I understand why, but this is only a slight improvement over having multiple methods for each metric type.

My intuition says that there ought to be a way to parameterize based on MetricType, which would involve some refactoring because as it stands there is a disconnect between MetricType and DataPoint. In other words, it would be nice we could statically enforce that if the type is MetricType.GAUGE, then the return type is (or the equivalent of) DataPoint<Double>. I thought about this for a little while before creating the ticket and couldn't come up with a solution. I am hoping that someone else can :)

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -455,6 +457,94 @@ public void setTaskScheduler(TaskScheduler taskScheduler) {
...
+                        case COUNTER:
+                            @SuppressWarnings("unchecked")
+                            Metric<Long> counter = (Metric<Long>) metric;
+                            return dataAccess.insertCounterData(counter, getTTL(counter.getId()))
+                                    .doOnNext(counterInserts::mark);
+                        default:
+                            throw new IllegalArgumentException(metric.getId().getType().getText());
+                    }
+                });
+        Observable<Integer> indexUpdates = dataAccess.updateMetricsIndex(metrics);
```
burmanm: Behavior is not identical to previous version here. On purpose? Not to mention like I discussed with @jsanda we should get rid of the PublishSubject in this one (and the extra subscribe) as it resubscribes to the source observable.

tsegismont:
Le 19/08/2015 20:44, jsanda a écrit :
> Can we do better than the approach I suggested in ticket? It has a

We'll try :) I've started with the suggested approach so that we can see 
what it would like.

> couple of problems. First, we lose some compile-time safety when
> inserting data. It would be nice if we could come up with something that
> does not lose those compile-time checks.
>
> Secondly, querying data points requires specifying the value class. I
> understand why, but this is only a slight improvement over having
> multiple methods for each metric type.

Agreed

>
> My intuition says that there ought to be a way to parameterize based on
> MetricType, which would involve some refactoring because as it stands
> there is a disconnect between MetricType and DataPoint. In other words,

That's my finding as well, it's not possible with the current data model 
to enforce a data point value type given a metric type.

> it would be nice we could statically enforce that if the type is
> MetricType.GAUGE, then the return type is (or the equivalent of)
> DataPoint. I thought about this for a little while before creating the

I guess you meant DataPoint<Double>. In this case, yes, it would be nice.

> ticket and couldn't come up with a solution. I am hoping that someone
> else can :)

I believe we can, if we modify the data model.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
tsegismont: Le 19/08/2015 23:26, Michael Burman a écrit :
> Behavior is not identical to previous version here. On purpose? Not to

I've compiled the three methods into one (only #addAvailData filtered on 
empty data points).

And I had to move updates of the Meter instances along with the calls to 
DataAccess#insertXXXData

While it does not behave exactly the same internally, the result is 
comparable, isn't it ?


> mention like I discussed with @jsanda <https://github.com/jsanda> we
> should get rid of the PublishSubject in this one (and the extra
> subscribe) as it resubscribes to the source observable.

I wasn't aware of this discussion. There's no mention of it in JIRA or 
the devel list. Can you update the ticket?

jsanda:
>I believe we can, if we modify the data model.

Let's see what you can come up with.

tsegismont:
@jsanda See new commit (rebased on master).

I made MetricType and MetricId generic. As Enums cannot be generic,I had to translate MetricType to the old style type safe enum.

With these changes, the compiler will show an error if you try something like:
```java
Metric<Long> m = new Metric<>(new MetricId<>("tenantId", GAUGE, "G1");
```
You will get such an error:
```
Error:(622, 36) java: incompatible types: cannot infer type arguments for org.hawkular.metrics.core.api.Metric<>
    reason: inferred type does not conform to equality constraint(s)
      inferred: java.lang.Long
      equality constraints(s): java.lang.Long,java.lang.Double
```

jsanda:
I really like the changes with MetricId and MetricType. I would change the signature of ```addDataPoints``` to be,

```java
<T> Observable<Void> addDataPoints(MetricType<T> type, Observable<Metric<T>> metrics)
```
then you only have to branch on the type once.

tsegismont:
@jsanda Done (rebased the PR on master again, and added `MetricType<T>` to `#addDataPoints` signature).

tsegismont:
@jsanda I've rebased the PR on master again. Can you review the last commit? This PR impacts quite a lot of code, so I need to rebase it almost any time another one is merged :)

jsanda:
Overall I think the changes look good. I have a couple thoughts. I wonder if it might better to make the ```inserter``` functions member variables instead of recreating them on every invocation of ```addDataPoints```. 

We have ```findDataPoints```, that returns raw data points for the different metric types. Then we have various finder methods like,

```java
<T> Observable<T> findGaugeData(MetricId<Double> id, Long start, Long end, 
        Func1<Observable<DataPoint<Double>>, Observable<T>>... funcs);

Observable<DataPoint<AvailabilityType>> findAvailabilityData(MetricId<AvailabilityType> id, long start,
        long end, boolean distinct);

Observable<List<GaugeBucketPoint>> findGaugeStats(MetricId<Double> metricId, long start, long end,
        Buckets buckets)

Observable<List<AvailabilityBucketPoint>> findAvailabilityStats(MetricId<AvailabilityType> metricId
        long start, long end, Buckets buckets)

Observable<DataPoint<Double>> findRateData(MetricId<Long> id, long start, long end)
```

It would be nice if we could combine some of these. Just food for thought. I am not advocating that it be done for this ticket. I would rather see it done separately and try to keep PRs smaller.

@stefannegrea or @burmanm would one of you mind taking a look at the changes? If there are no other changes/suggestions, then I think it is ready for merge.

tsegismont:
Le 26/08/2015 04:43, jsanda a écrit :
> Overall I think the changes look good. I have a couple thoughts. I
> wonder if it might better to make the |inserter| functions member
> variables instead of recreating them on every invocation of
> |addDataPoints|.
>

Done.

> We have |findDataPoints|, that returns raw data points for the different
> metric types. Then we have various finder methods like,
>
> <T>  Observable<T>  findGaugeData(MetricId<Double>  id,Long  start,Long  end,
>          Func1<Observable<DataPoint<Double>>, Observable<T>>...  funcs);

This one is specific to gauges.

>
> Observable<DataPoint<AvailabilityType>>  findAvailabilityData(MetricId<AvailabilityType>  id,long  start,
>          long  end,boolean  distinct);
>

This one is specific to avail.

> Observable<List<GaugeBucketPoint>>  findGaugeStats(MetricId<Double>  metricId,long  start,long  end,
>          Buckets  buckets)
>
> Observable<List<AvailabilityBucketPoint>>  findAvailabilityStats(MetricId<AvailabilityType>  metricId
>          long  start,long  end,Buckets  buckets)
>
> Observable<DataPoint<Double>>  findRateData(MetricId<Long>  id,long  start,long  end)
>
> It would be nice if we could combine some of these. Just food for
> thought. I am not advocating that it be done for this ticket. I would
> rather see it done separately and try to keep PRs smaller.

The three above could me merged into one. Feel free to open the JIRA and 
assign it to me.

PR #315, opened 2015-08-20
make terminal event timeout in REST API tests configurable

jsanda:
Some REST API tests have been periodically failing on travis builds because of
the timeout value for TestSubscriber.awaitTerminalEvent() being too low in
VirtualClockHandler. This commit makes it configurable via a system property
and increases the timeout for travis builds to 25 seconds.


stefannegrea:
Thanks @jsanda !

PR #316, opened 2015-08-20
[HWKMETRICS-225] Fix CORS filters across JAX-RS implementations

stefannegrea:


api/metrics-api-jaxrs-1.1/pom.xml
```diff
@@ -124,6 +124,16 @@
   <build>
     <finalName>hawkular-metrics-api-rest-1.1</finalName>
     <plugins>
+      <plugin>
```
tsegismont: Why is this needed? Seems strange as the skip and port properties are not even defined here.

api/metrics-api-jaxrs-1.1/src/main/webapp/WEB-INF/web.xml
```diff
@@ -30,17 +30,19 @@
   </context-param>
 
   <context-param>
-    <param-name>resteasy.media.type.mappings</param-name>
```
tsegismont: Could we remove this from the other web.xml as well so that we keep differences in the 2 impl to a minimum?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/HawkularMetricsRestApp.java
```diff
@@ -110,7 +111,8 @@ public HawkularMetricsRestApp() {
         classes.add(EmptyPayloadFilter.class);
         classes.add(TenantFilter.class);
         classes.add(MetricsServiceStateFilter.class);
-        classes.add(CorsFilter.class);
```
tsegismont: Could you remove this from the 1.1 impl as well?

api/metrics-api-jaxrs-1.1/src/main/webapp/WEB-INF/web.xml
stefannegrea: Sure, removing now.

api/metrics-api-jaxrs-1.1/pom.xml
stefannegrea: The goal is to be able to do `mvn jboss-as:run` from this project and have the war running inside the container.  I will add the properties to the pom.

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CorsFilter.java
```diff
@@ -16,39 +16,77 @@
...
+import javax.servlet.http.HttpServletResponse;
 
 /**
  * @author Stefan Negrea
  *
  */
-@Provider
-@ServerInterceptor
-public class CorsFilter implements PostProcessInterceptor {
+public class CorsFilter implements Filter {
```
tsegismont: Is there a reason to use servlet filter instead of a ContainerRequestFilter and ContainerResponseFilter, just as in the 2.0 impl ? If not, then using those classes would be better, in order to minimize the differences between the implementations.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CorsResponseFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.container.PreMatching;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * @author Stefan Negrea
+ *
+ */
+@Provider
+@PreMatching
```
tsegismont: PreMatching is only for ContainerRequestFilter

api/metrics-api-jaxrs-1.1/pom.xml
stefannegrea: I will remove the skip property from the config, but leave the management port in case, users want to override that from the maven command line.

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CorsFilter.java
stefannegrea: I did not find an equivalent in RestEasy 2. Do you have some documentation to that effect?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CorsResponseFilter.java
stefannegrea: Removed the annotation.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/HawkularMetricsRestApp.java
stefannegrea: The original CorsFilter still exists in JAX-RS 1.1 implementation.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CORSITest.groovy
```diff
@@ -0,0 +1,79 @@
...
+        def testOrigin = "TestOrigin123";
+
+        def response = hawkularMetrics.options(path: "ping",
+         headers: [
+            (ACCESS_CONTROL_REQUEST_METHOD): "POST",
+            (ORIGIN): testOrigin,
+         ]);
+
+         //print headers
+         println("==== Request Headers = Start  ====");
```
tsegismont: Could you remove this? We've tried to avoid useless output so far. If you need to print it in case of an error, then you can build a string with all the headers, and add it as 1st param of assertXXX calls.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/HawkularMetricsRestApp.java
tsegismont: Le 20/08/2015 18:38, Stefan Negrea a écrit :
> The original CorsFilter still exists in JAX-RS 1.1 implementation.

Why?

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CorsFilter.java
tsegismont: Oh, my bad, I did a mistake while looking at those classes in IDEA. Ok for servlet, there's no other way.

rest-tests/src/test/groovy/org/hawkular/metrics/rest/CORSITest.groovy
stefannegrea: Removed, see commit below.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/HawkularMetricsRestApp.java
stefannegrea: I am confused. I think you answered this in your last comment regarding the CORS implementation for JAX-RS 1.1:

tsegismont: If in the JAX-RS 1.1 impl the feature is implemented as a servlet filter, then there's no need to declare it here. Only JAX-RS related classes should be added to `classes`.

stefannegrea: That makes sense. I will remove it from the JAX-RS 1.1 impl.

mwringe:
It looks good to me and I can verify that my UI poc can now display metrics again using the jaxrs implementation. Would merge except for the stability-test failing.

mwringe:
retest please

mwringe:
retest this please

PR #319, opened 2015-08-21
[HWKMETRICS-190] Fetch tenantIds from the metrics_idx table also

burmanm:
Fetches the tenant_ids from metrics_idx table as well as tenant table and then returns the tenants.

### DIFF
--- a/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
+++ b/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
@@ -400,15 +400,20 @@ public class MetricsServiceImpl implements MetricsService, TenantsService {
         return dataAccess.findAllTenantIds()
                 .flatMap(Observable::from)
                 .map(row -> row.getString(0))
-                .flatMap(dataAccess::findTenant)
-                .flatMap(Observable::from)
-                .map(Functions::getTenant);
+                .distinct()
+                .flatMap(id ->
+                                dataAccess.findTenant(id)
+                                        .flatMap(Observable::from)
+                                        .map(Functions::getTenant)
+                                        .switchIfEmpty(Observable.just(new Tenant(id)))
+                );
     }
 
     private List<String> loadTenantIds() {
         Iterable<String> tenantIds = dataAccess.findAllTenantIds()
                 .flatMap(Observable::from)
                 .map(row -> row.getString(0))
+                .distinct()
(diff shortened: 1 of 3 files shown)


core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
```diff
@@ -395,15 +395,20 @@ public void setDateTimeService(DateTimeService dateTimeService) {
         return dataAccess.findAllTenantIds()
                 .flatMap(Observable::from)
                 .map(row -> row.getString(0))
-                .flatMap(dataAccess::findTenant)
-                .flatMap(Observable::from)
-                .map(Functions::getTenant);
+                .distinct()
```
jsanda: Is this ```distinct()``` call doing anything useful? Isn't is operating on a com.datastax.driver.core.Row object? I believe com.datastax.driver.core.ArrayBackedRow is the only Row implementation, and it does not implement ```equals`` in which case every row object would be considered distinct, right?

stefannegrea:
Please do not merge this PR until 0.6.0 has been released.

core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
burmanm: No, it's calling equals on the String methods (couple lines above we do map(row -> row.getString(0)), which is the tenantId. There's a test in the MetricsServiceImpl that checks that no duplicate lines are returned.

jsanda:
The changes look good to me but checkstyle is not happy due to unused import in MetricsServiceImpl.

burmanm:
However, the unused import is actually used?

"/home/travis/build/hawkular/hawkular-metrics/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java:24:15: Unused import - org.hawkular.metrics.core.api.MetricType.COUNTER_RATE."

But line 654 in MetricsServiceImpl (findRateData):

MetricId rateId = new MetricId(id.getTenantId(), COUNTER_RATE, id.getName(), id.getInterval());

So it's there. Buggy checkstyle (this error does not happen on my machine when compiling) or buggy travis compiling wrong PR?

jsanda:
I am going to merge this even though the perf tests are failing. I have spent time discussing the perf tests with Filip and determined the causes. I will address it in a separate ticket/PR.

PR #320, opened 2015-08-21
Upgrade to hawkular-parent 20

ppalaga:


stefannegrea:
Thanks @ppalaga !

mwringe:
Part of this is my fault for not configuring the container tests as a PR check, but in the future it might be good to do a quick text search to make sure you are getting all references to a property which is being renamed.

PR #322, opened 2015-08-21
[HWKMETRICS-228] calculate rates at query time

jsanda:
This commit also removes most of the code around scheduled task execution,
which includes the tasks for generating creates and creating tenants. The
latter is no longer neeed with the work done in HWKMETRICS-190.

### DIFF
--- a/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -49,11 +49,9 @@ import org.hawkular.metrics.api.jaxrs.config.ConfigurationProperty;
 import org.hawkular.metrics.api.jaxrs.util.Eager;
 import org.hawkular.metrics.api.jaxrs.util.VirtualClock;
 import org.hawkular.metrics.core.api.MetricsService;
-import org.hawkular.metrics.core.impl.CreateTenants;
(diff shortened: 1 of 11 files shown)


burmanm:
Something to think:

Should we consider doing some of these rate calculations with the UDFs that were introduced in the Cassandra 2.2.0 line (after we can move to that one) ? Or are we happier in keeping the calculations on the client side with a slight performance drop?

tsegismont:
@jsanda can you remove the code instead of commenting out ?

tsegismont:
If rates are now computed on the fly, can we adopt an API similar to stats for gauges and avail? In particular, we might want to let the user choose the size of buckets (instead of fixed 1 minute).

tsegismont:
This PR changes the order of results when querying counter data. It should be announced so that existing clients (mainly UI team) can plan to adapt.

jsanda:
>Should we consider doing some of these rate calculations with the UDFs

Yes, I think we should investigate UDFs.

jsanda:
>If rates are now computed on the fly, can we adopt an API similar to stats for gauges and avail?

We may want to do that. We need to get 0.6.0 out, and I would like to get this work into the release. The UI team needs this so it is an opportunity for some feedback. The current implementation in master is getting removed, so I would rather not have feedback based on it. 

I would like to propose that any other changes are done in follow up tickets/PRs.

jsanda:
>This PR changes the order of results when querying counter data. It should be announced so that existing clients (mainly UI team) can plan to adapt.

Yes, it needs to be announced and documented.

stefannegrea:
Thanks @jsanda !

PR #326, opened 2015-08-26
Add type to the MetricDefinition

burmanm:
[HWKMETRICS-232] Adds MetricType information to the MetricDefinition (with json name "type") to separate two different MetricDefinitions with different types but with same name and also to make sense to the getMetrics which returns a list of all metrics, but currently without a type.


jsanda:
LGTM. This is next in the merge queue after [PR 313](https://github.com/hawkular/hawkular-metrics/pull/313).

jsanda:
@burmanm Would you mind rebasing? PR 313 introduced changes to MetricType, and it make impact your changes, particularly the JSON serialization.

burmanm:
Well, that should be done now..

PR #327, opened 2015-08-26
HWKMETRICS-128

stefannegrea:


mwringe:
screenshot from 2015-08-26 17 46 43

Looks good. I find the SHA1 hash to be a bit confusing on that page. Maybe make it a bit smaller and put it on a new line with a Git SHA1 prefix?

Accessing the page in json returns:
{"Implementation-Version":"0.7.0-SNAPSHOT","Built-From-Git-SHA1":"664f742231c634d1c5754757307797f4cfe21235","name":"Hawkular-Metrics"}

Is there any way you can make it so that the name is the first element being returned?

stefannegrea:
Unfortunately I do not see an easy way to sort the JSON output. I tried a few permutations with the Map implementations but the output is identical. I am not sure if the JSON mapper is the one that dictates the final output or the map itself.

stefannegrea:
I pushed a fix for the html output, the SHA1 is on a separate line and with smaller font. 
screen-status-page

burmanm:
@JsonPropertyOrder ? Although I fail to see any reason in sorting json

stefannegrea:
There is no mapping class right now with all the properties to use the annotation. I switched to a simple Map<String,String> since the status and default handlers just return a few information fields.

stefannegrea:
Please retest

stefannegrea:
Retest this please

mwringe:
Looks good to me. Merging

PR #331, opened 2015-08-31
Pre-commit hook sample for JAX-RS diffs

tsegismont:
Metrics dev should add this hook to their local repo. Whenever we introduce changes which impact the diff, the file will show up on GH. So it will help both the developer and the reviewer to make sure the impls do not diverge.

### DIFF
--- a/api/diff-pre-commit-hook.sh
+++ b/api/diff-pre-commit-hook.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+#
+# Copyright 2014-2015 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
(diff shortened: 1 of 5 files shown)


tsegismont:
I've checked the diff file and fixed something in `MetricDefinition` classes. Other differences look expected to me.

stefannegrea:
> Seulement dans api

The diff file that you committed has a few keywords in French, could you use/force English (or any language) so there is no churn in the file when somebody commits with a different system locale?

Excellent work! If we resolve the language issue, this PR is ready to be merged.

tsegismont:
@stefannegrea thanks, and nice catch!

I've rebased the PR on master and added a commit (2f8a146) to address the language issue.

stefannegrea:
Thanks @tsegismont !

PR #332, opened 2015-08-31
[HWKMETRICS-126] Implement timestamp validation for all metrics

stefannegrea:
A missing timestamp will result in rejecting the entire request before processing it.

Note: this is implement in handler code rather than using JAX-RS framework features beacuse of divergence between the two JAX-RS standards. If/when one of the standards is dropped this code could be revisited although this implementation is simpler and easier to extend.

### DIFF
--- a/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/Availability.java
+++ b/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/Availability.java
@@ -21,6 +21,10 @@ import static java.util.Collections.unmodifiableList;
 import java.util.List;
 import java.util.Objects;
 
+import org.hawkular.metrics.api.jaxrs.validation.Validate;
+import org.hawkular.metrics.api.jaxrs.validation.Validator;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.wordnik.swagger.annotations.ApiModel;
 
@@ -31,7 +35,7 @@ import com.wordnik.swagger.annotations.ApiModel;
  * @author jsanda
  */
 @ApiModel(description = "An availability metric with one or more data points")
-public class Availability {
+public class Availability implements Validator {
 
     @JsonProperty
     private String id;
@@ -65,5 +69,12 @@ public class Availability {
         return "Availability{" +
                 "id='" + id + '\'' +
                 '}';
+    }
+
+    @Override
+    @JsonIgnore
+    @org.codehaus.jackson.annotate.JsonIgnore
+    public boolean isValid() {
+        return Validate.validate(this.getData()).toBlocking().lastOrDefault(false);
     }
 }
--- a/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/AvailabilityDataPoint.java
+++ b/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/AvailabilityDataPoint.java
@@ -21,9 +21,11 @@ import static java.util.Collections.emptyMap;
 import java.util.Map;
 import java.util.Objects;
 
+import org.hawkular.metrics.api.jaxrs.validation.Validator;
 import org.hawkular.metrics.core.api.AvailabilityType;
 import org.hawkular.metrics.core.api.DataPoint;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableMap;
 import com.wordnik.swagger.annotations.ApiModel;
@@ -32,11 +34,11 @@ import com.wordnik.swagger.annotations.ApiModel;
  * @author jsanda
  */
 @ApiModel(description = "Consists of a timestamp and a value where supported values are \"up\" and \"down\"")
-public class AvailabilityDataPoint {
+public class AvailabilityDataPoint implements Validator {
 
     @JsonProperty
     @org.codehaus.jackson.annotate.JsonProperty
-    private long timestamp;
+    private Long timestamp;
 
     @JsonProperty
     @org.codehaus.jackson.annotate.JsonProperty
@@ -67,7 +69,7 @@ public class AvailabilityDataPoint {
         tags = dataPoint.getTags();
     }
 
-    public long getTimestamp() {
+    public Long getTimestamp() {
         return timestamp;
     }
 
@@ -100,5 +102,20 @@ public class AvailabilityDataPoint {
                 .add("value", value)
                 .add("tags", tags)
                 .toString();
+    }
+
+    @Override
+    @JsonIgnore
+    @org.codehaus.jackson.annotate.JsonIgnore
+    public boolean isValid() {
+        if (this.getValue() == null || this.getValue().trim().isEmpty()) {
+            return false;
+        }
+
+        if (this.getTimestamp() == null) {
+            return false;
+        }
+
+        return true;
     }
 }
--- a/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/Counter.java
+++ b/api/metrics-api-common/src/main/java/org/hawkular/metrics/api/jaxrs/model/Counter.java
@@ -21,6 +21,10 @@ import static java.util.Collections.unmodifiableList;
 import java.util.List;
 import java.util.Objects;
 
+import org.hawkular.metrics.api.jaxrs.validation.Validate;
+import org.hawkular.metrics.api.jaxrs.validation.Validator;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.wordnik.swagger.annotations.ApiModel;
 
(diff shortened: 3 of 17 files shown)


jsanda:
Do we really need to check that each data point has a timestamp? We basically have two clients that we support - the Hawkular agent and Heapster. Both include timestamps with the data point that they send to the server. Honestly, I feel like this is something can be addressed with documentation.

stefannegrea:
@jsanda are you afraid that this check will affect performance? 

I can see how any client could just make a mistake in the integration code and send invalid data. And probably the easiest mistake is just omitting the timestamp. 

Also, I've seen examples of REST interfaces where a missing timestamp means that the current server timestamp should be used. Keeping it at zero, because it happens to be the default Java/Jackson implementation is not a good idea; requiring a timestamp is the best scenario in current implementation.

jsanda:
>are you afraid that this check will affect performance?

I am not overly concerned about performance. I suppose it could be an issue though when you consider service lots of concurrent requests, each of which could contain thousands of data points, and we are checking every single one. 

>I can see how any client could just make a mistake in the integration code and send invalid data. And probably the easiest mistake is just omitting the timestamp.

More than the potential performance impact, I am just not convinced that this is really needed. With respect to writing data we are not dealing with *any* clients. We are dealing only with two - Heapster and Hawkular. I am not aware of either having problems with including timestamps. I think this is something we can leave to the clients and any future clients just as we leave it to the clients to make sure metric ids are unique. I agree that a default value of zero is bad, but how likely is that? For ad hoc query situations like submitting data from curl, I can see it happening easily enough, but that is not what we are dealing with. 

>Also, I've seen examples of REST interfaces where a missing timestamp means that the current server timestamp should be used.

Can we address this with documentation?

tsegismont:
Le 01/09/2015 05:43, jsanda a écrit :
>     are you afraid that this check will affect performance?
>
> I am not overly concerned about performance. I suppose it could be an
> issue though when you consider service lots of concurrent requests, each
> of which could contain thousands of data points, and we are checking
> every single one.

I believe there might be a more efficient (fail-fast) way to check if 
the timestamp is present.

We could have a custom deserializer declared on :
@JsonDeserialize(using = TimestampDeserializer.class)

And TimestampDeserializer would throw an exception if the value is absent.

>
>     I can see how any client could just make a mistake in the
>     integration code and send invalid data. And probably the easiest
>     mistake is just omitting the timestamp.
>
> More than the potential performance impact, I am just not convinced that
> this is really needed. With respect to writing data we are not dealing
> with /any/ clients. We are dealing only with two - Heapster and
> Hawkular. I am not aware of either having problems with including
> timestamps. I think this is something we can leave to the clients and
> any future clients just as we leave it to the clients to make sure
> metric ids are unique. I agree that a default value of zero is bad, but
> how likely is that? For ad hoc query situations like submitting data
> from curl, I can see it happening easily enough, but that is not what we
> are dealing with.

It is true that Heapster and Hawkular agent are unlikely to send data 
without timestamp. But supporting script-based client has always been on 
our list of goals. And this ticket is about providing a convenient way 
to send a single value.


>
>     Also, I've seen examples of REST interfaces where a missing
>     timestamp means that the current server timestamp should be used.
>
> Can we address this with documentation?

If eventually we agree on not supporting server side timestamp, I 
believe we should not only document the issue, but also fail requests 
when the timestamp is absent.

tsegismont:
Maybe we could add a new endpoint 'POST /gauges/<metric>/single' so that clients can send a single value, without timestamp?

jsanda:
>Maybe we could add a new endpoint 'POST /gauges/<metric>/single' so that clients can send a single value, without timestamp?

Interesting idea. Would the client care about the timestamp used, whether it is milliseconds or minutes past the time when the request was made?

stefannegrea:
I do not like the idea of introducing a single end-point with a special behavior. First we increase the number of end-points without providing and additional functionality, and second it will be confusing if this end-point has a different behavior.

stefannegrea:
> If eventually we agree on not supporting server side timestamp, I
believe we should not only document the issue, but also fail requests
when the timestamp is absent.

I agree with this, just documentation is not sufficient if the behavior is unexpected.

stefannegrea:
> I believe there might be a more efficient (fail-fast) way to check if
the timestamp is present.

>We could have a custom deserializer declared on :
>@JsonDeserialize(using = TimestampDeserializer.class)

>And TimestampDeserializer would throw an exception if the value is absent.

One constraint, the code needs to work for now on both Jackson 1 and 2. It looks like the custom deserializer is available in both. As a side note, Jackson 2 has a null check but we cannot use it yet.

But I am not sure if this is significantly faster than the code I already implemented.

tsegismont:
Le 01/09/2015 22:26, Stefan Negrea a écrit :
> I do not like the idea of introducing a single end-point with a special
> behavior. First we increase the number of end-points without providing
> and additional functionality, and second it will be confusing if this
> end-point has a different behavior.

If you add an endpoint and document its behavior with Swagger, I can't 
see how it can be confusing.

tsegismont:
Le 01/09/2015 22:58, Stefan Negrea a écrit :
> One constraint, the code needs to work for now on both Jackson 1 and 2.
> It looks like the custom deserializer is available in both. As a side
> note, Jackson 2 has a null check but we cannot use it yet.
>

It looks like there is an even simpler way:
For the JAX-RS 2 impl:
com.fasterxml.jackson.databind.DeserializationFeature#FAIL_ON_NULL_FOR_PRIMITIVES

For the JAX-RS 1.1 impl
org.codehaus.jackson.map.DeserializationConfig.Feature#FAIL_ON_NULL_FOR_PRIMITIVES

No need to write a custom deserializer.

I believe we should set those flags to true. If, in a DTO, a number may 
be absent, then it shouldn't be mapped as a primitive.

> But I am not sure if this is significantly faster than the code I
> already implemented.
>

The current proposal has the following drawbacks:
- the whole payload will be deserialized, whether it is valid or not
- when the payload is valid, it will be traversed twice
- the changes impact a lot of classes

On the proposal itself, the "Validate" class makes pointless use of 
"Observable". This forces the developer to add 
".toBlocking().lastOrDefault(false)" calls everywhere in the JAX-RS 1.1 
implementation, or ".lastOrDefault(false).subscribe(valid -> {" in the 
JAX-RS 2.0 implementation.

tsegismont:
Le 01/09/2015 19:29, jsanda a écrit :
> Interesting idea. Would the client care about the timestamp used,
> whether it is milliseconds or minutes past the time when the request was
> made?
>

Sorry I'm not sure what you mean. Can you elaborate?

jsanda:
+1 on use of ```FAIL_ON_NULL_FOR_PRIMITIVES```

stefannegrea:
>It looks like there is an even simpler way:
>For the JAX-RS 2 impl:
com.fasterxml.jackson.databind.DeserializationFeature#FAIL_ON_NULL_FOR_PRIMITIVES

>For the JAX-RS 1.1 impl
org.codehaus.jackson.map.DeserializationConfig.Feature#FAIL_ON_NULL_FOR_PRIMITIVES

This will not work because it is a generic deserializer setting, we need something specific to the timestamp properties. There might be other properties that allow empty request fields.

jsanda:
>There might be other properties that allow empty request fields.

I am not aware of any other primitive properties in the data point classes for which null input would be acceptable. We have the ```value``` and ```timestamp``` properties. That is it, right?

tsegismont:
Le 02/09/2015 16:53, Stefan Negrea a écrit :
> This will not work because it is a generic deserializer setting

This should work if our global policy is "don't map nullable values to 
primitive fields".

stefannegrea:
> We have the value and timestamp properties. That is it, right?

`value` is string for Availability metrics. Is that still considered a primitive?

tsegismont:
> `value` is string for Availability metrics. Is that still considered a primitive?

Shame on me. I hadn't thought about the avail data point case. Forget about null for primitive then.

This works on both JAX-RS implementations.

```java
    @JsonCreator(mode = Mode.PROPERTIES)
    @org.codehaus.jackson.annotate.JsonCreator
    @SuppressWarnings("unused")
    public AvailabilityDataPoint(
            @JsonProperty("timestamp") @org.codehaus.jackson.annotate.JsonProperty("timestamp")
            Long timestamp,
            @JsonProperty("value") @org.codehaus.jackson.annotate.JsonProperty("value")
            String value,
            @JsonProperty("tags") @org.codehaus.jackson.annotate.JsonProperty("tags")
            Map<String, String> tags
    ) {
        Preconditions.checkArgument(timestamp != null, "timestamp is null");
        Preconditions.checkArgument(value != null, "value is null");
        this.timestamp = timestamp;
        this.value = value;
        this.tags = tags == null ? Collections.emptyMap() : Collections.unmodifiableMap(tags);
    }
```

stefannegrea:
Unless there is strong opposition this last proposal, I will re-work the PR with that method of validation.

jsanda:
I am opposed.

tsegismont:
Le 02/09/2015 22:02, jsanda a écrit :
> I am opposed.

If you find the constructor approach too invasive, the JsonDeserializer 
for XXXDataPoint classes is better in this regard.

jsanda:
I am fine with the constructor approach.

tsegismont:
As agreed on IRC, I'll make the changes in another PR. This other PR will fix HWKMETRICS-127 at the same time. I am going to copy other the new itests.

Thanks @stefannegrea !

PR #335, opened 2015-09-03
Minor POM and doc changes.

tsegismont:
Some properties still present in ptrans itest config no longer exist in REST module.
By default, keep test server file logging to a minimum (like CONSOLE level).
Update README to precise on Wildfly version to download.


stefannegrea:
Thanks @tsegismont !

PR #339, opened 2015-09-07
Upgrade to hawkular-parent 21

ppalaga:


ppalaga:
@stefannegrea is this just the usual "fails sometimes" that only needs a CI restart?

tsegismont:
Not sure what happened. The server port offset got a strange value. 
Restarted the build.

tsegismont:
It looks like it's not an intermittent failure. I suppose you tried the build on your box?

ppalaga:
> you tried the build on your box?

No, I have not.

> The server port offset got a strange value.

There is a wildfly-maven-plugin upgrade from 1.1.0.Alpha2 to 1.1.0.Alpha3. Maybe they changed something in that area.

tsegismont:
> There is a wildfly-maven-plugin upgrade from 1.1.0.Alpha2 to
> 1.1.0.Alpha3. Maybe they changed something in that area.

Yes, see my email in the devel list.

tsegismont:
We can't upgrade parent POM until the issue is fixed in Wildfly Maven Plugin.

PR #340, opened 2015-09-07
HWKMETRICS-127 Add back support for specifying data retention with tenant creation

tsegismont:
This PR fixes the following issues:
- HWKMETRICS-127 Add back support for specifying data retention with tenant creation
- HWKMETRICS-126 Data points timestamp defaults to 0
- HWKMETRICS-244 Internal server error when sending malformed availability data points

Along with the fixes, it consolidates the models:
- all model classes (request and response) under the same package for clarity
- TenantParam renamed to TenantDefinition because it's both a request and response model
- shared MetricDefinition model class
- all model classes immutable
- use @JsonCreator for deserialization and basic, on-the-fly validation
- annotate getters for serialization when it makes sense (filter out empty collections/maps, serializer selection, custom field name)

Note that for HWKMETRICS-126, we won't provide a way to use server timestamp. If, in the future, we get demand for this, a solution would be to add a separate endpoint.


tsegismont:
@jsanda Can you please review? I will only add a new commit to fix unfortunate test name changes in `rest-tests` module

tsegismont:
@jsanda I've pushed a couple of commits to address your concerns. I've also updated the API doc to indicate data retention is stored in days.

tsegismont:
Travis failed in the scheduler module:
```
Failed tests: 

  TaskSchedulerITest.executeTaskThatRepeatsTwice:225 Number of onNext events differ; expected: 2, actual: 1
```

@jsanda is this intermittent failure known or should I open a new ticket?

jsanda:
I tried to make the task scheduler tests more predictable using a virtual clock, but there is still a lot of concurrency going on with them. I don't see a need for new ticket unless it becomes a more serious problem.

PR #341, opened 2015-09-07
Dropwizard/Codahale Metrics dependency management

tsegismont:
Opening this PR as I noticed recent changes in Dropwizard Metrics dependency management.

If we should use Codahale Metrics, in the same version as the C* driver, then we shouldn't:
- force a version in parent POM
- exclude/re-include it in core-impl POM

Am I wrong?


tsegismont:
PR #342 supersedes this one.

PR #342, opened 2015-09-08
POM cleaning and JAX-RS duplications removal

tsegismont:
POM cleanup

- removed redundant dependency declarations
- moved log4j and slf4j-over-log4j to test scope, otherwise they end up in the JAX-RS implementations WAR files
- removed codehaus jackson from JAX-RS 2.0 WAR file

Eliminate as much duplicated code as possible between JAX-RS implementations.


tsegismont:
The two most important things here are:
- reduced war file size
- restriction of JAX-RS differences to a minimum (handlers, exception mappers, filters, interceptors, Jackson config)

tsegismont:
@jsanda could you review with one as well? Thanks!

jsanda:
Looks like you need to rebase. After that, I think we are ready to merge.

tsegismont:
Le 09/09/2015 16:45, jsanda a écrit :
> Looks like you need to rebase. After that, I think we are ready to merge.

Done

PR #343, opened 2015-09-08
HWKMETRICS-250 Core-API does not compile on newest JDK 1.8

tsegismont:
HWKMETRICS-250 Core-API does not compile on newest JDK 1.8

Helped the compiler to infer generic type


tsegismont:
restest this please

tsegismont:
@FilipB can you have a look at the perf tests? The code I've changed isn't used in the test scenario AFAIK. So it's probably not the cause of the failure.

tsegismont:
retest this please

PR #347, opened 2015-09-11
HWKMETRICS-255 Upgrade to Swagger 1.5

tsegismont:
- Artifact and package names changed
- Dropped the Mustache template and generate document from JSON with Groovy
- Fixed doc issues here and there
- Removed annotations from JAX-RS 1.1 module (not used)

The doc is in better shape now :
 - no broken links
 - better representation of arrays and allowable values
 - new format column


tsegismont:
@burmanm would you mind to review?

burmanm:
Looks good.

PR #348, opened 2015-09-11
[HWKMETRICS-254] Remove m_tags field and usage from the data table.

stefannegrea:
The storage and queries are to be done via the metrics_idx table.


stefannegrea:
retest this please

jsanda:
retest this please

FilipB:
@stefannegrea @jsanda 
This PR is 4% faster than current baseline. If you think this is expected ping me please so we can coordinate merge with change of current baseline.

stefannegrea:
The code from this PR is faster because some data insertion code was removed and queries are now done from a different table. There is no change in user functionality, the changes are related to performance and data optimization.

@FilipB, we will need to update the baseline once this PR is merged.

PR #350, opened 2015-09-14
Exclude Jackson utility classes from CDI scans

tsegismont:
Exclude Jackson utility classes from CDI scans

 This is to avoid warnings (classes not available) on EAP6.4 during deployment


tsegismont:
Small update for the record: the warnings also show up on Wildfly. I could not see them at first because I had turned off the corresponding logging categories.

PR #352, opened 2015-09-15
HWKMETRICS-125 Add JBoss Logging as in all Hawkular components

tsegismont:
HWKMETRICS-125 Add JBoss Logging as in all Hawkular components

Conventions (Hawkular wide, gathered from mailing list threads):
- put logging stuff into `log` package
- start logging method with log level
- logger name is `log`
- only put messages which need an id (level INFO and above) in logger interfaces

Conventions (Metrics specific):
- group all messages in one class for maintainability
- use `XXXLogging` helper class to get a logger with specific category

Do not use primitive arguments in logging interfaces until we get rid of EAP6.4

Logback is now the logging backend for tests. It allows to set log level with a
system property while still having a default value:
```xml
<filter class="ch.qos.logback.classic.filter.ThresholdFilter">
  <level>${test.logging.console.level:-ERROR}</level>
</filter>
```
Developers can use `test.logging.console.level` and `test.logging.file.level` to set
the console and file test logging levels.


clients/ptranslator/src/main/java/org/hawkular/metrics/clients/ptrans/PTrans.java
```diff
@@ -60,7 +60,7 @@
  * @author Thomas Segismont
  */
 public class PTrans {
-    private static final Logger LOG = LoggerFactory.getLogger(PTrans.class);
+    private static final PTransLogger log = PTransLogging.getPTransLogger(PTrans.class);
```
pilhuhn: Just curious: why does this not follow `Logger log = Logger.get...` as in all the other cases? Is that due to limitations in the annotation processor?

tsegismont: Not sure what you mean.

The annotation processor just creates an implementation of the 
PTransLogger interface.

If you want the PTrans class to have access to PTransLogger methods, you 
have no choice but to declare the typed logger, not a generic JBoss 
Logging logger.

Does that answer your question?

pilhuhn: On 16 Sep 2015, at 13:11, Thomas Segismont wrote:

>> @@ -60,7 +60,7 @@
>> * @author Thomas Segismont
>> */
>> public class PTrans {
>> -    private static final Logger LOG = 
>> LoggerFactory.getLogger(PTrans.class);
>> +    private static final PTransLogger log = 
>> PTransLogging.getPTransLogger(PTrans.class);
>
> Not sure what you mean.

I was talking about the naming of that Interface class,
which is different from the pattern you use elsewhere.

-- 

tsegismont: I don't understand. How is it different from, for example, `RestLogger`, or `CoreLogger` ?

See https://github.com/hawkular/hawkular-metrics/pull/352/files#diff-9c93f73c3ccf924ff0d6ef4edf4219feR89

tsegismont:
Guys, sorry to insist, could you have a look sometime soon? As logging involves almost all modules, I'd rather not keep this PR open for too long.

jsanda:
I started looking at this this evening. I will continue reviewing tomorrow.

tsegismont:
Thanks

PR #354, opened 2015-09-15
HWKMETRICS-257

mwringe:
Fixes a Cassandra startup issue where it can show the container as ready when it should be pending.


jsanda:
I see that there is a test for scaling from one to two Cassandra nodes. What about starting with a multi-node cluster and scaling up from that?

mwringe:
The problem is that there is a lot of extra effort to get tests going for different configuration files. The one it is checking right now is the default one which is part of the build.

In order to perform tests on multiple initial configuration files, we would need to create new separate maven modules for each configuration files.

Even as it stands, the tests are very minimal. Trying to do more advanced tests in this space can be really difficult.

PR #357, opened 2015-09-16
More Grafana compatibility fixes and HWKMETRICS-65

tsegismont:
This PR implements "HWKMETRICS-65 Influx endpoint: add suport for timestamp precision query parameter" and fixes some issues for better Grafana compatibility:

- Avoid integer "overflow"

The parser used to store time amounts into int variables. This led to parsing errors if a time range operand was too far in the past.

- In where clause, absolute and relative time operands might have no time unit

The parser used to support explicit time units only, but the in the query language spec (aka the online manual) it is allowed to omit it.


tsegismont:
Would you mind to review @pilhuhn ? I recommend to look at individual commit diffs to get at better feeling of what happened.

PR #358, opened 2015-09-18
[HWKMETRICS-186] Add tagging support for counter metrics

stefannegrea:


jsanda:
Overall I think the changes look good. I have one comment and one question though. I would be in favor of pushing the tag validation down into MetricsService so that we only have to do it in one place and to keep the logic in the REST API to a minimum. As for my question, are there any REST API tests for the new endpoints? I did not see any but maybe I missed it.

stefannegrea:
I did not consider pushing the validation to the metrics service but that is a good idea. I update the code in the last commit.

stefannegrea:
All the tag related integration test now run on all the metric types. Please see this change:
https://github.com/hawkular/hawkular-metrics/commit/446cb00246821c18f93db865103f4962180ed778#diff-22240e21a40f348c547594cf77f1ee00R30

jsanda:
Ah, I missed the tags tests. Thanks for pointing it out.

PR #363, opened 2015-09-22
HWKMETRICS-182

mwringe:
Add in support to authenticate against the OpenShift server using subjectaccess reviews.


mwringe:
Some minor update discovered when testing with security enabled and with Heapster

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/BasicAuthentication.java
```diff
@@ -0,0 +1,144 @@
...
+            ".htpasswd");
+
+    public static final String MD5_PREFIX = "$apr1$";
+    public static final String SHA_PREFIX = "{SHA}";
+
+    private static final Logger logger = LoggerFactory.getLogger(BasicAuthentication.class);
+
+    private Map<String, String> users = new HashMap<>();
+
+    public BasicAuthentication() throws Exception {
```
burmanm: Lets not throw too generic exceptions (even just IOException is better), catching Exception catches RuntimeExceptions also.

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/OpenShiftTokenAuthentication.java
```diff
@@ -0,0 +1,180 @@
...
+        if (token == null || tenant == null) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST,
+                               "The '" + OpenShiftAuthenticationFilter.AUTHORIZATION_HEADER + "' and '"
+                                       + HAWKULAR_TENANT + "' headers" + " are required");
+            return;
+        }
+
+        if (isAuthorized(request.getMethod(), token, tenant)) {
+            filterChain.doFilter(request, response);
+            return;
```
burmanm: Pointless return

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/OpenShiftTokenAuthentication.java
```diff
@@ -0,0 +1,180 @@
...
+        } else {
+            response.sendError(HttpServletResponse.SC_FORBIDDEN);
+            return;
```
burmanm: Same here, pointless return.

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/OpenShiftTokenAuthentication.java
```diff
@@ -0,0 +1,180 @@
...
+            connection.setRequestProperty("Accept", "application/json");
+            connection.setRequestProperty("Authorization", token);
+
+            //Add the body
+            OutputStream outputStream = connection.getOutputStream();
+            for (byte b : generateSubjectAccessReview(projectId, verb).getBytes()) {
+                outputStream.write(b);
+            }
+            outputStream.flush();
+            outputStream.close();
```
burmanm: If IOException is thrown by the flush() or write(), this close() is never called. Refactor to use try-with or at least in the finally block.

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/OpenShiftTokenAuthentication.java
```diff
@@ -0,0 +1,180 @@
...
+            if (responseCode == 201) {
+                ObjectMapper objectMapper = new ObjectMapper();
+                JsonNode jsonNode = objectMapper.readTree(connection.getInputStream());
+
+                if (jsonNode.get("allowed").asText().equals("true")) {
+                    return true;
+                }
+
+            } else {
+                return false;
```
burmanm: Pointless return clause and if-clause.

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/OpenShiftTokenAuthentication.java
mwringe: OK, updated it to remove the extra returns. I like to put those returns in there to make sure that its the exit point for that method. We should always return after a sendError or doFilter, but in this case they are not needed.

containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/BasicAuthentication.java
mwringe: OK, update to just be IOException instead of Exception

mwringe:
Ok, made changes based on the PR comments.

stefannegrea:
Thanks @mwringe!

PR #365, opened 2015-09-23
NPE when app is deployed but Session is not created

tsegismont:
NPE when app is deployed but Session is not created

If Metrics has been started but Cassandra is not up, the Session object
is never created. Then, if you stop Metrics, you get an NPE.


stefannegrea:
Thanks @tsegismont!

PR #366, opened 2015-09-24
Upgrade to hawkular-parent 24

ppalaga:


tsegismont:
@ppalaga it seems that Wildfly Maven Plugin 1.1.0.Alpha4 has not been released yet. I'm afraid we can't upgrade to parent 24 until then. Unless Wildfly Maven Plugin version is overriden in Metrics parent.

ppalaga:
@tsegismont +1, I was about to write this. Just in case there is something interesting for Metrics in the new parent, there is still the option of overriding to `<version.wildfly-maven-plugin>1.1.0.Alpha2</version.wildfly-maven-plugin>` in metrics.

tsegismont:
OK, then would you mind to update your PR? I can merge afterwards.

ppalaga:
@tsegismont done

tsegismont:
Thanks Peter!

PR #368, opened 2015-09-25
HWKMETRICS-265 Counter "rate" returning same data as "data" endpoint

tsegismont:
HWKMETRICS-265 Counter "rate" returning same data as "data" endpoint

In the previous implementation, we used to:
- sort out counter data points in one minute buckets
- compute rate based on low and high value

The implementation had the following issues:
- if there's just one point per minute in the original counter, you get a rate metric similar to the original counter metric
- if there's less that one point per minute you get a rate metric with many `NaN` values

The new implementation has a basic approach:
 - buffer counter data observable into "previous/next" pairs
 - compute rate for each pair, where timestamp is the average of timestamps in the pair

Let's take an example. Consider the following counter data points:
```
{"timestamp":0,"value":0}
{"timestamp":30000,"value":50}
{"timestamp":60000,"value":80}
```

From these data points, we can create the following pairs:
```
{"timestamp":0,"value":0} / {"timestamp":30000,"value":50}
{"timestamp":30000,"value":50} / {"timestamp":60000,"value":80}
```

And the computed rate data points are:
```
{"timestamp":15000,"value":100.0}
{"timestamp":45000,"value":60.0}
```

### DIFF
--- a/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
+++ b/core/metrics-core-impl/src/main/java/org/hawkular/metrics/core/impl/MetricsServiceImpl.java
@@ -42,7 +42,6 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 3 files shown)


stefannegrea:
Thanks @tsegismont!

PR #373, opened 2015-09-28
HWKMETRICS-274 Missing JBoss logging method on EAP6.4

tsegismont:
HWKMETRICS-274 Missing JBoss logging method on EAP6.4

Do not use primitive arguments in logging methods calls.


tsegismont:
Closing, created on a wrong branch

PR #374, opened 2015-09-28
HWKMETRICS-274 Missing JBoss logging method on EAP6.4

tsegismont:
Do not use primitive arguments in logging methods calls.


stefannegrea:
Thanks @tsegismont!

PR #378, opened 2015-09-29
HWKMETRICS-252 Warning about batch size threshold in the logs

tsegismont:
HWKMETRICS-252 Warning about batch size threshold in the logs

Use more conservative value since warnings are still present on Openshift.


jsanda:
The purpose of the change is to eliminate the warnings that are literally filling up the Cassandra logs. Without some investigation, testing, and analysis, we can only speculate about appropriate batch sizes. Until we do that, I think it makes sense to do what we can to eliminate the warnings.

jsanda:
https://goo.gl/ySgpwB is worth a read as it contains some additional thoughts on the subject of batch statements.

mwringe:
I have been running this on my openshift instance and I no longer see the warning once I enable this patch. It looks good to me.

PR #381, opened 2015-09-30
HWKMETRICS-291 Provide different strategies for C* statement grouping

tsegismont:
HWKMETRICS-291 Provide different strategies for C* statement grouping

This PR introduces a contract for batching strategies.

So far, two strategies are implemented:
- use batch statements
- use single statement

We can add variations of the first strategy if needed.

The user can choose the batching strategy with configuration. If batch statement strategy is chosen, batch statement size can be configured.


tsegismont:
**Please do not merge until v0.7 is released.**

tsegismont:
Can anyone have a look now that v0.7 is released?

jsanda:
I have to wrap up work for HWKMETRICS-292 and then I can spend some time on this.

jsanda:
From the first commit,

>we should never use LOGGED batches

We currently do not use logged (i.e., atomic) batches, but that does not mean that they should never be used. Cassandra guarantees that if part of an atomic batch succeeds, then all of it will. It should be noted that this does not imply that all part of the batch will succeed at the same time. This guarantee comes with a performance trade off, which might be acceptable and some situations. We have various places where we perform updates across multiple partitions/tables, and atomic batches might be appropriate in some of the places.

jsanda:
I like the idea of introducing a batching strategy, but I think this requires some more thinking / discussion. For example, in terms of the batch size we need to look at the physical size in terms of bytes, not the number of statements. We also need to think about cluster topology. What works well for one node may or may not work well for 3, 5, etc., nodes. We also should think about what to do with a batch that contains mutations across multiple partitions. Do we still create an unlogged batch and just log a warning message? Do we fail fast? Do we use an atomic batch in that situation?

tsegismont:
OK. I'll amend the commit message.

tsegismont:
> I like the idea of introducing a batching strategy, but I think this requires some more thinking / discussion. For example, in terms of the batch size we need to look at the physical size in terms of bytes, not the number of statements. We also need to think about cluster topology. What works well for one node may or may not work well for 3, 5, etc., nodes. We also should think about what to do with a batch that contains mutations across multiple partitions. Do we still create an unlogged batch and just log a warning message? Do we fail fast? Do we use an atomic batch in that situation?

I couldn't find a way to evaluate a `BatchStatement` size in bytes. If there is one, I can make the changes.

If we need more strategies or more options in `BatchStatementStrategy`, then I agree we should add them. But I'd rather not start implementing smart, self-tuning switches right now. The purpose of this PR is to be able to evaluate the performance of the strategies with different cluster sizes. When we'll have the performance tests results, we'll be able to answer all of your questions, and to implement such switches.

tsegismont:
@jsanda I've added a `groupByReplica` option to `BatchStatementStrategy`

tsegismont:
Note that `groupByReplica` is currently not very useful, because `DataAccessImpl` always uses `BatchStatement` for mutations within a single partition.

tsegismont:
OK to merge @jsanda ?

jsanda:
Sorry, I did not get to finish reviewing the last couple commits. I will continue looking at it tomorrow.

jsanda:
I am not really comfortable with merging. There is a non-trivial amount of code in this PR, and I do not see much in terms of functionality. We reduced the max batch size in HWKMETRICS-252 to help reduce the warnings that were filling up the Cassandra log. Cassandra will log a warning or fail to execute a batch if its sizes exceeds thresholds specified in cassandra.yaml (which are also configurable over JMX). The size is **not** the number of statements but rather the size in terms of bytes, which means it is entirely possible for Cassandra to log the warning for a batch that contains only a single statement. We need to investigate how we could calculate the size or an approximation without introducing too much performance overhead. If we want to make the batch statement count configurable, then couldn't we more easily via a system property for example?

As for the group by replica changes, as you mentioned we only use unlogged batches for mutations within a single partition. That is by design. If we start trying to create unlogged batches across multiple partitions, then I would consider that a problem with our code that has to be fixed. Logged or atomic batches across multiple partitions is a different story. If we need to write to multiple tables and if we need to make sure either all or none of those updates fail, then we should consider whether or not the performance overhead of an atomic batch is acceptable.

FilipB:
The performance drop caused by this PR is between 2-3% and it seems to be real (not caused by perftest instability). 3 runs showed the same results - we can try it again later just to be sure.

tsegismont:
Le 09/10/2015 15:26, Filip Brychta a écrit :
> The performance drop caused by this PR is between 2-3% and it seems to
> be real (not caused by perftest instability). 3 runs showed the same
> results - we can try it again later just to be sure.
>

It seems to report the same for another, doc changes only pull request:
http://209.132.178.156/report370.html

tsegismont:
John, sorry, I had no time yet to reply. I will do this week. I've been 
working on tickets with higher priority.

FilipB:
@tsegismont the report you pointed out is +2% (I upgraded cassandra to 2.2.1 today so it's probably faster.. still investigating) report from this PR is -3% (I will run it again when I set new baselines for cassandra 2.2.1)

FilipB:
@tsegismont I tried a few more runs and the drop -2/-3% is still visible. Other PRs are ok. So it will be necessary to change baselines before merging this to master or make it 3% faster:)

FilipB:
Retest this please

jsanda:
This PR has been open for a while, and I should have gotten back to it sooner. I am going to close this based on my last comment. I definitely think that this is something we should revisit though.

PR #384, opened 2015-10-01
HWKMETRICS-293 REST API: Extract time ranges and buckets param classes to simplify handler code

tsegismont:
Introduced TimeRange and BucketParams classes in 'param' package.

Handlers use these classes to validate parameters and create core library entities.

### DIFF
--- a/api/diff.txt
+++ b/api/diff.txt
@@ -466,27 +466,27 @@ diff -r '--exclude-from=api/diff-excludes' api/metrics-api-jaxrs/src/main/java/o
 >                 .build();
 >         return ServerResponse.copyIfNotServerResponse(response);
 diff -r '--exclude-from=api/diff-excludes' api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
-28d27
+25d24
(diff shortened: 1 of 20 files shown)


tsegismont:
retest this please

tsegismont:
It seems the perf test job failed because it couldn't build Hawkular.

tsegismont:
Can someone have a look? Thanks!

jsanda:
I have to wrap up work for HWKMETRICS-292 and then I will spend some time on this as well.

jsanda:
Overall I like the changes, but I have question around parameter validation. Would it be better to do the validation in the core service layer? Having the logic in the core service layer makes it easier to test/debug. Having it in the REST API means we have to deal with more JAX-RS / container specific stuff which sometimes can be more cumbersome than working with the core server layer. When we are dealign with HTTP / REST specific things, I absolutely think we should do validations, conversions, etc. in the REST API, but I do not think this is one of those situations.

There is another benefit of pushing more logic down into the core service layer. If someone wants to provide another REST implementation or a non-REST/HTTP implementation, then he won't have to reimplement that validation logic.

tsegismont:
In this PR, there are two objects involved.

- BucketConfig

In the BucketConfig class, we validate JAX-RS params and eventually try 
to build a org.hawkular.metrics.core.api.Buckets instance. Model 
validation happens in the Buckets class. So we're all good.

- TimeRange

The TimeRange class only checks that the end bound is strictly bigger 
than the start bound. I will extract this check in a utility method in 
the core library, and then re-use the utility method in MetricsService.

PR #385, opened 2015-10-02
Close C* cluster only when session creation fails

tsegismont:
While working on another driver issue today,
I realized that we do not need to call cluster#close if cluster init fails.

So I've changed #createSession method a bit to call cluster#close only when necessary.


stefannegrea:
Thanks @tsegismont!

PR #387, opened 2015-10-04
HWKMETRICS-292

jsanda:


jsanda:
There is one thing I want to point in. In the new method in GaugeHandler, I added ```@DefaultValue("")``` to the tags parameter. The tags parameter is required. If the request omits the tags parameter and if ```@DefaultValue("")``` is not supplied, the server returns a 500 status due a NPE being thrown in one of RestEasy's inject classes. I think we need to handle missing parameters. I looked briefly to see if there is a way to handle this in TagsConverter, but I think the NPE is thrown before TagsConverter gets called. I think the NPE is thrown from ```StringParameterInjector.extractValues(List<String> values)``` at this line,

```java
for (int i = 0; i < values.size(); i++) Array.set(vals, i, extractValue(values.get(i)));
```
I am not sure if there is a better way to handle it than how I did.

stefannegrea:
@tsegismont has the #384 that extracts the bucketing and time range validation to common. Should we wait for that PR and then rebase?

jsanda:
I think #384 is good to go, so merging that and rebasing makes sense.

jsanda:
>Can we add support for a list of metric ids too? 

Do you mean additionally searching by ids?

stefannegrea:
Right now you only have support for get the list of metrics for aggregation based on tag querying. The endpoint should also support a list of metric ids. The question is what happens if the user provides both.

jsanda:
>The question is what happens if the user provides both.

I have been wondering the same thing. If we can figure out quickly, I am fine with trying to make the changes in this PR. If not, I would like to considering the additional changes as a future PR because the current changes are needed ASAP by the Heapster sink.

stefannegrea:
We can do one or the other; if both are provided, then return an error. If this feature does not make it in this PR, I will implement it part of JIRA 262 & 263.

jsanda:
Doing one or the other sounds good to me. I can try to do it in this PR.

tsegismont:
Can you double check? Your usage of tags as a query param does not look 
different than its usage in MetricHandler#findMetrics. And I just 
checked, it works when you don't supply the query param.

jsanda:
I did double check and I do indeed get the same behavior that I previously cited. That is really strange since we do not see the same behavior in findMetrics. I will investigate a bit more. Maybe there is something else going on causing the NPE that I am missing.

stefannegrea:
@jsanda, the changes look good. Should we merge this PR as is and add metric id queries later?

jsanda:
I am in the process of adding support for ids.

stefannegrea:
Perfect! Thanks @jsanda  for the clarification. Will watch for the changes.

jsanda:
The PR is ready again for review. I have added support for querying by metric names. I added an overloaded version of ```MetricsService.findGaugeStats``` that takes a list of metric names instead of tags. In the future it might be nice to have single method that take a search/query criteria object instead the additional methods in MetricsService.

burmanm:
I hadn't had a chance to review this earlier, but what I wished for metricIds was a filter that would be a combination of tagsFilter & metricIds (filtering only by regexp metricId is not possible). This is different use-case than the one provided by metricNames. Hopefully this metricName-list isn't a stopper for that feature.

tsegismont:
Thanks. But this is not blocking, I think. So if you can't figure out 
quickly, let's file a JIRA and come back to this later.

stefannegrea:
Let's open a JIRA for a more complicated filtering scheme. The current one implemented by @jsanda should be good for this release. 

I will merge this PR because it is the foundation for two other JIRAs.

PR #388, opened 2015-10-05
Change percentile95th -> percentile and add percentile query parameter

burmanm:
This allows requesting different percentile than 95th to be calculated when requesting buckets. The REST API changes from `percentile95th` -> `percentile` parameter and the new query parameter is called `percentile` (with a default value of 95.00 for old use-cases).

This is necessary for the Initial Resources implementation (it needs configurable percentile), although I'd almost want to see following:

`?buckets=1&percentiles=90,99,99.9` returning:

```javascript
"percentiles": {
  "90": 0.0
  "99": 0.0
  "99.9": 0.0
}
```

### DIFF
--- a/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
+++ b/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
@@ -25,6 +25,7 @@ import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.serverError;
 import static org.hawkular.metrics.core.api.MetricType.COUNTER;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -50,6 +51,7 @@ import org.hawkular.metrics.api.jaxrs.model.GaugeDataPoint;
 import org.hawkular.metrics.api.jaxrs.model.MetricDefinition;
 import org.hawkular.metrics.api.jaxrs.param.BucketConfig;
 import org.hawkular.metrics.api.jaxrs.param.Duration;
+import org.hawkular.metrics.api.jaxrs.param.Percentiles;
 import org.hawkular.metrics.api.jaxrs.param.Tags;
 import org.hawkular.metrics.api.jaxrs.param.TimeRange;
 import org.hawkular.metrics.api.jaxrs.util.ApiUtils;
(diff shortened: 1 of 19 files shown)


jsanda:
The changes look fine to me. How come you did not add support for multiple percentiles?

tsegismont:
- Is there a JIRA for this PR ?

- The UI team needs to be warned about this change: the result object is changed and they need to use the new query parameter

tsegismont:
@burmanm how about keeping the hardcoded 95thpercentile and adding the multivalued paramater as you suggest in the description? That would avoid breaking `hawkular-charts` compatibility.

stefannegrea:
We cannot introduce charting breaking APIs in 0.8.0.

mtho11:
I agree with @stefannegrea that we can't introduce breaking changes. Maybe add the other percentiles and mark it deprecated in the notes for 0.8.0 and then we can remove it in 0.9.0. 
This gives UI and others time to update their code without things breaking.

burmanm:
Will do those changes, I'll create the JIRA to remove it after 0.8.0 release.

burmanm:
Rebased

tsegismont:
@burmanm hehe, Travis is back. Maybe it heard you rant about rebasing :stuck_out_tongue:

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
```diff
@@ -217,12 +220,17 @@ public Response findCounterData(
                         .toBlocking()
                         .lastOrDefault(null);
             } else {
+                if(percentiles == null) {
```
jsanda: You do not need the null check here. JAX-RS will default to an empty list if the ```percentiles`` param is not in the query string.

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
```diff
@@ -258,8 +267,13 @@ public Response findRate(
...
```
jsanda: No need for the null check.

jsanda:
No need for the null check on the ```percentiles`` query param. JAX-RS will default to an empty list if it is not in the query string.

burmanm:
Percentiles is not a list, but an object. JAX-RS will default to null in that case (the list is inside the Percentiles object).

jsanda:
>Percentiles is not a list, but an object

My mistake. Yeah, the null checks makes sense then.

jsanda:
I tested manually against Hawkular master to make sure that the API changes do not break the UI. The changes look good to go.

PR #390, opened 2015-10-08
HWKMETRICS-263

stefannegrea:


jsanda:
Overall, I think the change look good. I have a couple minor nits around naming. I am not a fan of the findSimpleGaugeStats or findSumGaugeStats method names or the SumNumericBucketPointCollector class name. With that said, I have been struggling to come up with names. I think there needs to be a larger discussion around terminology both in terms of docs and in the code. The naming isn't anything that should hold up merging though.

From an API perspective, I am not so sure about the ```downsample``` query param. I find it confusing. It does not really describe what is going on. By default this endpoint fetches data points from any number of metrics that match the search criteria (which is either a list of metric names or a set of tag filters). We then perform our "bucketing" which is more commonly referred to as down sampling and interpolation. When data points do not align, i.e., have the same timestamps, we interpolation to align them. We then perform down sampling, which reduces the number of data points at each timestamp using one or more aggregation functions.

With your changes, we perform the interpolation and down sampling for each metric separately. When that is done, we merge all of the streams / series or in effect perform another down sampling. Because whether we have 1 or N matching metrics, we still are guaranteed to get only one value per timestamp.
I am afraid that I do not have a good suggestion at the moment, but maybe we can think of a different term that might better differentiate the behaviors.

stefannegrea:
I think as you pointed out the biggest concern is `downsample` query param because is public facing. I wanted something simple that relates to the task being done in the end-point. I thought about `downsamplestrategy` but that was too long. So I shortened to `downsample`. But then I ran out of ideas. We could switch to `strategy`. Could that be a better name for the query param?

jsanda:
LGTM. I will merge as soon as we get the green light from travis. Thanks!

PR #395, opened 2015-10-09
HWKMETRICS-264

stefannegrea:


jsanda:
It seems there is a good bit of code in the two versions of ```MetricsServiceImpl.findNumericStats``` that is duplicated. We ought to see if we can refactor that; however, I am fine with doing it post-merge. The things I would like to see addressed prior to merge are in the tests.

stefannegrea:
@jsanda, I tried to reduce the duplication in those two methods by introducing a third method that takes a set of metrics and does the work. But got a bit difficult due to generics usage. Let's revisit this post 0.8.0.

jsanda:
Refactoring the duplication after 0.8.0 sounds good to me. I was more interested in the test code. Thank you for the changes. Part of the reason I suggested using canned, i.e., hard coded values is that it can simplify test code, reducing the chance of bugs in the test code and also making the test code easier to read. For example, here is the code to compute the expected min for a bucket,

```java
T expectedMin = combinedData.stream()
        .min((x, y) -> Double.compare(x.getValue().doubleValue(), y.getValue().doubleValue()))
        .get()
        .getValue();
```

If we use hard coded values, it is just be a matter of a simple assignment. I realize however, this is definitely a subjective area and the most important thing is having test coverage in place, which we have. Thanks again!

PR #396, opened 2015-10-12
Use CLI commands to configure test server

tsegismont:
Use CLI commands to configure test server instead of copying files from source tree


pilhuhn:
Doesn't that require a running server (at least in admin-only mode) as opposed to a not (yet) started server?

tsegismont:
Le 12/10/2015 17:39, Heiko W. Rupp a écrit :
> Doesn't that require a running server (at least in admin-only mode) as
> opposed to a not (yet) started server?

Yes it does. That's why the execution happens after the plugin starts 
the server.

On one hand, it prevents us from setting the level when the logging 
subsystem starts. On the other hands, it's a lot of content removed from 
the repo in favor or 3 CLI commands.

It's not a big problem if we get logging defaults when the server is 
started. At this point it does not log too much.

stefannegrea:
This is a massive reduction in configuration files and lines of code. Thanks @tsegismont!

stefannegrea:
@pilhuhn, if the application  server is not running then the tests will fail too. I do not see requiring a running server a problem.

jsanda:
LGTM. nicely done!

stefannegrea:
Thanks @tsegismont!

PR #398, opened 2015-10-15
Hawkular bus integration fixes

tsegismont:
HWKMETRICS-315 Bus integration: events randomly or never posted

When Metrics was started after the Pinger, the Pinger would accumulate data to post before sending it all at once eventually.
Because there were not enough threads to handle the peak load, RxJava tried to apply backpressure but PublishSubject does not support it by nature.

The fix consists in:
- buffering inserted metric to avoid submitting too much to the #io scheduler
- adding a buffer-on-backpressure strategy

Note that the problem was emphasized before as we used to schedule sending on a wrapped ManagedExecutorService. Now we use the #io scheduler (change already in master).

I've successfully tested the changes on Hawkular server.

### DIFF
--- a/hawkular-component/src/main/java/org/hawkular/metrics/component/publish/AvailDataPublisher.java
+++ b/hawkular-component/src/main/java/org/hawkular/metrics/component/publish/AvailDataPublisher.java
@@ -88,7 +88,7 @@ public class AvailDataPublisher {
         List<AvailDataMessage.SingleAvail> availList = avail.getDataPoints().stream()
                 .map(dataPoint -> new AvailDataMessage.SingleAvail(availId.getTenantId(), availId.getName(),
                         dataPoint.getTimestamp(),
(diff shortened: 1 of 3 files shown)


tsegismont:
I've been trying to implement Metrics publishers with raw ActiveMQ API. The goal was to put the connection in async sending mode.

I got the implementation working, in the sense it passed testing without errors. But I'm not sure it actually sends things asynchronously. When you get a connection object from the resource adapater, it's a "managed" connection object. So I the `#setUseAsyncSend` calls is effective.

I could spend more time digging into this, but has the changes in the PR already solve the issue, I'd like to have them merged to move forward on "HAWKULAR-698".

jsanda:
I will open a ticket against Hawkular to track the changes discussed on the dev list with repsect to switching to the default JMS provider in WildFly which implements JMS 2.0. JMS 2.0 provides async sending which I think is necessary for scalability.

PR #399, opened 2015-10-17
[HWKMETRICS-313] Remove the hardcoded 95th percentile calculation

burmanm:
Removes the old 95th percentile calculation, use the percentiles instead to calculate the necessary percentiles.


burmanm:
There are as part of the original percentiles calculation, testBucketPercentiles() in MetricsServiceITest and CountersITest (JAX-RS test) percentileParameter, but I'll add some of those previous tests back with new syntax.

stefannegrea:
Thanks @burmanm!

PR #403, opened 2015-10-22
[HWKMETRICS-319] use non-persistent delivery mode for sending data to…

jsanda:
… bus


tsegismont:
There's an issue in the code: the sending call for numeric data is commented out. This explains why the tests failed on Travis.

Besides, the changes are not really needed, because the broker creates non persistent destinations:
- http://git.io/vWrna
- http://git.io/vWrnD

So I would prefer not to merge this one.

jsanda:
I should have noted that I intentionally commented out the JMS call for numeric data. I wanted to make sure that the issue was in fact the JMS call and not something happening on the client side.

tsegismont:
Ha, sorry, I didn't notice it was added for debugging in a second commit

PR #404, opened 2015-10-27
Remove TTL from data table and use partitioning with dpart

burmanm:
This solves:
 * HWKMETRICS-189
 * HWKMETRICS-191
 * HWKMETRICS-192
 * HWKMETRICS-193

Use dparts in the data table for partitioning purposes (no more DPART = 0). data_retention column is removed from the data table. Default value for buckets is 1 day and each metric has new property bucket_size that can override the default value.


jsanda:
I am a little late to start looking at this, but there are a couple things I want to point out. First with this change purging data now needs to be done manually. We need a scheduled, periodic background job to handle this. Secondly, since H-Metrics 0.8.0 went into OpenShift, we need to consider how these changes impact 0.8.0 installations. Some changes may be needed to handle that.

jsanda:
About the second commit and date/time zone usage in general... There have been issues/discussions in the past about time zones. Maybe we should be defaulting to UTC. @stefannegrea what do you think?

stefannegrea:
For any releases subsequent to 0.8.0 we will need to add data migration. Release 0.8.0 marks a first release with a major integration and we will need to be careful with regards to schema changes going forward.

stefannegrea:
We need to be extra careful with timezones (there is a history on this). At the server level, we should never use anything but UTC; but interpreting of the actual time should be left to the user.

burmanm:
@stefannegrea There's no timezone control in current releases (that's why there's a test change because the tests with exact timestamps behave differently on different machines). I created HWKMETRICS-324 for UTC change, but I don't think this PR should take care of that.

jsanda:
Thanks for creating HWMETRICS-324. I agree that the time zone issue should not hold up the PR given that it is consistent with current behavior.

jsanda:
Since I was the one who chose the column name ```dpart```, I will free feel to say it is a crappy, non-descriptive name. Buckets is a more familiar term in the context of time series databases. On that note, I suggest we change the method name ```DateTimeService.getDparts``` to ```getBuckets``` or something along those lines. We should rename the ```dpart``` column as well, but I am ok with doing that separately once we get our schema upgrade/migration process ironed out.

jsanda:
I noticed that the bucket size is specified in seconds. Data retention is specified in days. I think it makes things less confusing for users to keep the units consistent. That raises the question of what is the most appropriate unit. To be honest, I would be surprised if we need anything smaller than days for either bucket size or data retention; however, using seconds does provide flexibility which can be useful for performance/stress testing.

jsanda:
The bucket size (i.e., duration) is a configuration setting like data retention; consequently, I think it should be treated similarly. While it is not well defined, we have three levels of configuration - system-wide, per tenant, and per metric. We should adopt the same patter for bucket size. When a user deploys H-Metrics for the first time, he should be able to specify the system-wide defaults for data retention and bucket size.  Similarly with tenant creation, clients should be able to specify both data retention and bucket size.

I realize that there is going to be a follow up PR to introduce caching for the bucket setting, but I am not comfortable with doing a read on every write to get the bucket size. I would prefer to do something similar to what we do with data retentions in which we populate a cache on start up and update it on tenant/metric creation.

burmanm:
That method was my planning (for caching) to update it on metric creations, updates and potentially start-up (although it would populate itself for often used metrics pretty fast so I'm not sure if loading them from the Cassandra is necessary..)

FilipB:
Retest this please

FilipB:
This PR introduces performance drop ~13%. Master baseline for gauges writes is 16950 req/sec - this PR has 14685 req/sec

jsanda:
I am going to close this PR for now since it is pretty out of sync with master. These changes are dependent on [HWKMETRICS-194](https://issues.jboss.org/browse/HWKMETRICS-194), and we really need to figure that out before we put these changes into master. I do not think we should put these changes into master until we have the capability to run a scheduled job to perform the deletions.

PR #405, opened 2015-10-27
HWKMETRICS-322 InfluxDB API: groupby should accept milliseconds unit

tsegismont:
HWKMETRICS-322 InfluxDB API: groupby should accept milliseconds unit

Updated the grammar and tests. I thought ms were not accepted in groupby clause, as indicated in the documentation:
https://influxdb.com/docs/v0.8/api/query_language.html#group-by

But running the query on a real Influx server shows that it is supported.


stefannegrea:
Thanks @tsegismont!

PR #406, opened 2015-10-28
HWKMETRICS-322 InfluxDB API: groupby should accept milliseconds unit

tsegismont:
HWKMETRICS-322 InfluxDB API: groupby should accept milliseconds unit

This is the second part for fixing HWKMETRICS-322.

Add validation rule to make sure bucket size in groupby is bigger than metric resolution
Support subsecond bucket size in #applyMapping
Add tests


stefannegrea:
Thanks @tsegismont!

PR #407, opened 2015-11-06
[HWKMETRICS-320] move common model classes into new module

jsanda:
This will make it easier to reuse the REST API model classes.


stefannegrea:
Thanks @jsanda!

PR #408, opened 2015-11-09
HWKMETRICS-66 Influx endpoint: use series name prefix to match a counter or a gauge

tsegismont:
* Updated the "list series" handler so that it returns gauges with the "gauge." prefix and counters with the "counter." prefix.
* Updated the query handler so that it looks at the series name for a known prefix ("gauge." or "counter."); if there's no prefix, it will assume it's a gauge
* Updated the POST handler as well
* Removed old (and buggy) method: DataAccess#findAllGaugeMetrics
* Introduced MetricTypeFilter, to use with Observable#compose


tsegismont:
haw-perf-stability-test job failed for unrelated reason (errors in inventory)

stefannegrea:
Thanks @tsegismont!

PR #410, opened 2015-11-17
Fix Influx handler log issue (%d used with double value)

tsegismont:
%d cannot be used with double values, the call ends with a runtime exception thrown by String#format


stefannegrea:
Thanks @tsegismont!

PR #412, opened 2015-11-20
HWKMETRICS-326

stefannegrea:


core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/codehause/jackson/MetricTypeKeySerializer.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.hawkular.metrics.api.jaxrs.codehaus.jackson;
+package org.hawkular.metrics.core.api.codehause.jackson;
```
tsegismont: Typo in the package name: "codehause"  -> "codehaus"

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricRequest.java
```diff
@@ -46,19 +40,19 @@
  * @author John Sanda
  */
 @ApiModel(description = "A gauge metric with one or more data points")
```
tsegismont: Documentation needs to be fixed

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricRequest.java
```diff
@@ -46,19 +40,19 @@
...
-public class Gauge {
+public class MetricRequest<S, T extends DataPoint<S>> {
```
tsegismont: Could we keep that simple and declare `MetricRequest<T>`, and below `private final List<DataPoint<T>> data` ?

tsegismont: That would make code using MetricRequest much more concise.

core/metrics-core-api/src/main/java/org/hawkular/metrics/core/api/MetricRequest.java
```diff
@@ -46,19 +40,19 @@
...
+    public MetricRequest(
             @JsonProperty("id")
             @org.codehaus.jackson.annotate.JsonProperty("id")
             String id,
             @JsonProperty("data")
             @org.codehaus.jackson.annotate.JsonProperty("data")
-            List<GaugeDataPoint> data
+            List<T> data
     ) {
         checkArgument(id != null, "Gauge id is null");
```
tsegismont: Message is wrong

tsegismont:
Wasn't the plan to move all model classes to a separate project, so that consumers (like a Java client) can use it without depending on core-api?

tsegismont:
Can you have a look at the generated doc ? 

```
cd api/metrics-api-jaxrs
mvn clean install -Pdocgen
cd target/generated
asciidoctor rest-metrics.adoc 
xdg-open rest-metrics.html 
```

It seems that the changes lead to very long name where `MetricRequest` is used. I believe it could be made shorter by simplifying the generic declaration as commented earlier.

Also, there seem to be some duplication:
rest-metrics.html#DataPointDouble
rest-metrics.html#GaugeDataPoint

The API should use one or the other, not both. Given that the goal of this PR is to reuse core library objects as much as possible, I guess GaugeDataPoint should go away.

stefannegrea:
It is not possible to simplify the MetricRequest because Jackson no longer knows how to deserialize the request. And the only solution would to introduce concrete classes for each request type, which then balloons the number of classes. I prefer a more verbose class  declaration than 1 base class with 3 subclasses.


With the following changes, I get the error bellow:

Changes:
```
public class MetricRequest<S>
```
```java
@POST
@Path("/data")
public Response addGaugeData{
     List<MetricRequest<Double>> gauges
```
Error:
```bash
{"errorMsg":"No suitable constructor found for type [simple type, class org.hawkular.metrics.core.api.DataPoint<java.lang.Double>]: can not instantiate from JSON object (missing default constructor or creator, or perhaps need to add/enable type information?)\n at [Source: io.undertow.servlet.spec.ServletInputStreamImpl@7265bf74; line: 1, column: 24] (through reference chain: java.util.ArrayList[0]->java.util.ArrayList[0])"}
```

stefannegrea:
> Also, there seem to be some duplication:
>rest-metrics.html#DataPointDouble
>rest-metrics.html#GaugeDataPoint

The duplication is because of JSON deserialization and deserialization. In this case, a concrete class is needed rather than a single core class because of specific string representation.

stefannegrea:
@tsegismont, the plan was to have one single package for model classes. At this point I think that is what core-api is. The only exception is the MetricsService interface, which I could see getting moved up to core-impl.

Is there anything that should be moved out of the core-api that does belong there?

tsegismont:
I was opposed to the changes (see comments HWKMETRICS-326 on JIRA) for the following reasons:
- transfer objects give you flexibility because you can change the core model without breaking clients
- transfer objects play nicely with serialization and documentation tools
- its has no been proven that using transfer objects is a performance issue in Metrics

Then during the last Metrics meetings (November, 18) it was argued that doing it would bring simplicity, for the users and the team. I don't believe the goal has been achieved:
- most of the transfer objects have been kept, moved up to the core library
- portions of code which involve using the new MetricRequest class are harder to read
- there are objects providing the same functionality (e.g. DataPoint<Double> , GaugeDataPoint); this will cause confusion for users, whether they want to use these objects in a Java client or they are looking at the documentation
- the generated documentation has unreadable data objects names (e.g. MetricRequestAvailabilityTypeAvailabilityDataPoint)

Consequently, I'm :-1: for merging.

stefannegrea:
@tsegismont, the last two commits should address all the points from your last comment. There is only DataPoint<T> left, and MetricRequest only needs one type parameter. Initially I overlooked the custom type serializers/deserializers. The project was actually missing the deserializer for AvaiabilityType and registering the two globally.

Please let me know if there any other problems after the last update.

tsegismont:
The last two commits make things better, but the original promise is not fulfilled: the code still has multiple representations of the same entities (e.g. MetricDefinition/MetricRequest/Metric)

So still :-1: on this one.

stefannegrea:
@tsegismont, I can further unify the three classes you mention in your last message by instantiating a generic MetricId with a system tenantId and an "Undefined" metric type. That is exactly what MetricRequest is today; and MetricRequest is later transformed into a Metric by adding additional info inside the JAX-RS method. And MetricDefinition is somewhat analogous to MetricRequest although it serves a little different role.

Any feedback for this proposal before I start refactoring?

stefannegrea:
@tsegismont, the last set of 5 commits should address the class duplication you mention in your last comment. Also the documentation looks really clean to me after this last set of updates.

stefannegrea:
I just rebased on master to make sure the code changes are good with the lastest code. Everything looks good.

vprusa:
retest this please

vprusa:
test this please

stefannegrea:
I had to do one more rebase on master due to merge conflicts.

jsanda:
Why were the generics for MetricType removed? Without the type params I can do,

```java
Observable<DataPoint<String>> observable = metricsService.findDataPoints(
    new MetricId<>(tenantId, GAUGE, "m1"), start.getMillis(), end.getMillis());
```

and it compiles just fine, which is bad because we lose some type safety. With the generics in place this code will produce the following compiler error,

```
[ERROR] incompatible types: inferred type does not conform to equality constraint(s)
[ERROR] inferred: java.lang.String
[ERROR] equality constraints(s): java.lang.String,java.lang.Double
```

jsanda:
I think the source code organization is a bit inconsistent and confusing. We have modules under hawkular-metrics/core and hawkular-metrics/api. Whether we have a DTO layer or not, I would like to see a module that just contains the model classes, e.g., hawkular-metrics-model or hawkular-metrics-api-model. Having that makes it more clear as to what the supported data model and schema is, and it also makes it easier for others to consume.

I do not think MetricsService belongs in the same package or maven module as the model classes. It is not part of the data model. Same for Aggregate.java. That should live along side MetricsService. And if we move the ```Metric.toObservable``` method as well, then we can remove the rxjava dependency on hawkular-metrics-core-api or wherever the model classes wind up. This will make it easier to consume.

stefannegrea:
The generic type was never used inside the MetricType class, and in my opinion that is a confusing use of generics. The code looks much simpler and cleaner now. 

However, it can be added back to the class but then we need to live with untyped instance for the Undefined variant of MetricType.

stefannegrea:
You are right, MetricsService is the only class that is not a model class in that package. I propose rolling `MetricsService` into hawkular-metrics-core-impl and also moving `Metric.toObservable` in the same module. Is this a good plan?

jsanda:
>The generic type was never used inside the MetricType class, and in my opinion that is a confusing use of generics. The code looks much simpler and cleaner now.

Its usage might be unclear, but it adds compile time type safety. Compile my example above with and without the generics to see for yourself.

Moving MetricsService and Metric.toObservable into hawkular-metrics-core-impl sounds good.

stefannegrea:
@jsanda, I addressed the last comment with the previous three commits. 

As soon as this PR is merged, I will open another one to rename the core-api and core-impl modules. But for this PR, I would like to stop short of that because there is already too much churn for a single PR.

jsanda:
I think the changes look good, and doing module name changes in a separate PR sounds good. I know the @tsegismont is on PTO (and congrats btw), but it would be nice if he is able to chime in since he had some strong objections early on.

vprusa:
retest this please

vprusa:
retest this please

tsegismont:
I don't like the new `MetricType.UNDEFINED`, which was introduced to work around the fact `Metric` entities require a non null type field. To me, it is confusing. Especially when you look at the handler code for metric creation. Hopefully this can be resolved in the future.

Other than that, Stefan has addressed all the blocking points.

PR #417, opened 2015-12-14
Update core module names, location, and Java package names

stefannegrea:


pilhuhn:
Renaming 'api' to 'model' sounds odd to me. To me 'model' are more the data classes that are moved around with the methods of the (public) api.

jsanda:
> To me 'model' are more the data classes that are moved around with the methods of the (public) api.

@pilhuhn That is exactly the reason for the renaming. The maven module now consist only of the model classes used in the public API.

jsanda:
Two small nits. I think "models" should be made singular which would also be consistent with the package name; so, I would rename the module directory to metrics-model and rename the artifactId to hawkular-metrics-model. Secondly, I would move the module up a directory since it is no longer a core, internal set of model classes but instead the public API model classes.

pilhuhn:
On 16 Dec 2015, at 15:34, jsanda wrote:
> @pilhuhn That is exactly the reason for the renaming. The maven module 
> now consist only of the model classes used in the public API.

Perfect then :)

stefannegrea:
@jsanda, I renamed the modules to singular; that is a good suggestion to make things less confusing for users.

However, I would like to keep the code for module itself in the core folder. The folder does not have a pom.xml, so when the module gets published to maven it will be part of the same flat structure. But for internal purposes until we trim down the number of modules (eg. get rid of JAX-RS 1.1 stuff) I like the current arrangement where we have four major areas: core, api, testing and containers. I would like to keep models either under core or api. But since JAX-RS modules do not depend directly on it, only indirectly via the metrics-core-service module, the best place is core.

jsanda:
> But since JAX-RS modules do not depend directly on it, only indirectly via the metrics-core-service module, the best place is core.

Actually the JAX-RS modules do depend directly on it. The model module is not just a transitive dependency as the model class are directly used in the JAX-RS modules. It just so happens that the JAX-RS modules pick up the dependency transitively.

But as I said, this is a rather minor nit. Thanks for addressing my comments.

PR #418, opened 2015-12-18
Upgrade to WildFly 10

ppalaga:


ppalaga:
@jsanda or somebody else, could you please review and eventually merge?

ppalaga:
e92930b is using hawkular-parent 31 which allows versions like 1.2.3-SNAPSHOT to make the perf job happy.

All tests are also passing locally on my laptop with e92930b .

ppalaga:
I'd say we can interpret the CI as passing. The hawkular integration build cannot work, because hawkular is not ready for WF10 yet.

Is there anything else against merging?

PR #419, opened 2015-12-21
Bus comes now from bus module, not from nest module.

jpkrohling:


rhqci:

rhqci:

jpkrohling:
DON'T MERGE YET

I updated the commit, so that it now consumes the Accounts Feature Pack (includes Nest, Bus and the necessary JMS endpoints, which, for some reason, are only on the Accounts FP). Ideally, the JMS definitions would be on Bus, but I actually think that using the Accounts FP is not a bad idea (might be required for HAWKULAR-549) 

*But*, the tests for the Hawkular Component are still failing. A constructor was removed from the Bus during the last migration, and I'm not sure what's the appropriate way to set username/session to the session via bus. So, I'm leaving the commit as it is right now, and I'd ask @jsanda to adjust it (as it seems he did this last migration).

jsanda:
I would like to hold off on this for now, and I will instead open a different metrics PR based on the bus changes in https://github.com/hawkular/hawkular-commons/pull/36.

stefannegrea:
I think this PR is superseded by PR #423. @jsanda can we close this PR now?

jsanda:
Stefan is correct. PR #423 refactored the tests and changed them to arquillian integration tests that run against a stock WF 10 distro with the necessary JMS configuration. I think we can close this out.

stefannegrea:
Thanks @jsanda for the clarification. @jpkrohling, per comments above this PR is closed without merging.

PR #420, opened 2015-12-22
[HWKMETRICS-331] Add 'fromEarliest' parameter to return buckets from …

stefannegrea:
…earliest date data is available, starting with the first non-empty bucket. In case all buckets are empty, return nothing.

### DIFF
--- a/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
+++ b/api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
@@ -56,6 +56,7 @@ import org.hawkular.metrics.model.DataPoint;
 import org.hawkular.metrics.model.Metric;
 import org.hawkular.metrics.model.MetricId;
 import org.hawkular.metrics.model.MetricType;
+import org.hawkular.metrics.model.NumericBucketPoint;
 import org.hawkular.metrics.model.exception.MetricAlreadyExistsException;
 import org.hawkular.metrics.model.param.BucketConfig;
 import org.hawkular.metrics.model.param.Duration;
@@ -219,6 +220,7 @@ public class GaugeHandler {
             @PathParam("id") String id,
             @QueryParam("start") final Long start,
             @QueryParam("end") final Long end,
+            @QueryParam("fromEarliest") Boolean fromEarliest,
             @QueryParam("buckets") Integer bucketsCount,
             @QueryParam("bucketDuration") Duration bucketDuration,
             @QueryParam("percentiles") Percentiles percentiles
@@ -230,6 +232,10 @@ public class GaugeHandler {
         BucketConfig bucketConfig = new BucketConfig(bucketsCount, bucketDuration, timeRange);
         if (!bucketConfig.isValid()) {
             return badRequest(new ApiError(bucketConfig.getProblem()));
+        }
+
+        if (bucketConfig.getBuckets() == null && Boolean.TRUE.equals(fromEarliest)) {
+            return badRequest(new ApiError("fromEarliest can only be used with bucketed results"));
         }
 
         MetricId<Double> metricId = new MetricId<>(tenantId, GAUGE, id);
@@ -247,9 +253,34 @@ public class GaugeHandler {
                     percentiles = new Percentiles(Collections.<Double>emptyList());
                 }
 
-                return metricsService
-                        .findGaugeStats(metricId, timeRange.getStart(), timeRange.getEnd(), buckets,
-                                percentiles.getPercentiles())
+                Observable<TimeRange> observableTimeRange = Observable.just(timeRange);
+
+                if (Boolean.TRUE.equals(fromEarliest)) {
+                    metricsService.findMetric(metricId).first().map((metric) -> {
+                        long now = System.currentTimeMillis();
+                        long earliest = now - metric.getDataRetention() * 24 * 60 * 60 * 1000L;
+                        return new TimeRange(earliest, now);
+                    });
+                }
+
+                final Percentiles localPercentiles = percentiles;
+                return observableTimeRange
+                        .flatMap((localTimeRange) -> metricsService.findGaugeStats(metricId, localTimeRange.getStart(),
+                                localTimeRange.getEnd(), buckets, localPercentiles.getPercentiles()))
+                        .map((list) -> {
+                            if (Boolean.TRUE.equals(fromEarliest)) {
+                                int index = 0;
+                                for (NumericBucketPoint item : list) {
+                                    if (!item.isEmpty()) {
+                                        break;
+                                    }
+                                    index++;
+                                }
+                                return list.subList(index, list.size());
+                            }
+
+                            return list;
+                        })
                         .map(ApiUtils::collectionToResponse)
                         .toBlocking()
                         .lastOrDefault(null);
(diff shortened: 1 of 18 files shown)


api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
```diff
@@ -247,9 +253,34 @@ public Response findGaugeData(
                     percentiles = new Percentiles(Collections.<Double>emptyList());
                 }
 
-                return metricsService
-                        .findGaugeStats(metricId, timeRange.getStart(), timeRange.getEnd(), buckets,
-                                percentiles.getPercentiles())
+                Observable<TimeRange> observableTimeRange = Observable.just(timeRange);
+
+                if (Boolean.TRUE.equals(fromEarliest)) {
```
jsanda: What is this ```if```` block doing? You are not doing anything with the results. Am I missing something? Same question applies to corresponding method in GaugeHandler.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
```diff
@@ -300,8 +307,34 @@ public void findGaugeData(
...
+            if (Boolean.TRUE.equals(fromEarliest)) {
+                metricsService.findMetric(metricId).first().map((metric) -> {
+                    long now = System.currentTimeMillis();
+                    long earliest = now - metric.getDataRetention() * 24 * 60 * 60 * 1000L;
+                    return new TimeRange(earliest, now);
+                });
+            }
+
+            final Percentiles localPercentiles = percentiles;
+            observableTimeRange
```
jsanda: Note that these comments apply also to the corresponding changes in CounterHandler. First, The only thing I see being done when ```fromEarliest``` is true is that we filter out empty buckets. Don't we want to include the oldest data available which would be determined by the retention settings. It looks like this code is just using whatever the value is for the start time param. We are missing code to calculate the start time.

Secondly, this code can be simplified and written in a more functional way, e.g.,

```java
if (Boolean.TRUE.equals(fromEarliest)) {
    metricsService.findGaugeStats(metricId, timeRange.getStart(), timeRange.getEnd(), buckets,
            percentiles.getPercentiles())
            .filter(bucket -> !bucket.isEmpty())
            .map(ApiUtils::collectionToResponse)
            .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
} else {
    metricsService.findGaugeStats(metricId, timeRange.getStart(), timeRange.getEnd(), buckets,
        percentiles.getPercentiles())
        .map(ApiUtils::collectionToResponse)
        .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
}
```

api/metrics-api-jaxrs-1.1/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
stefannegrea: A few lines below there is this code part of the same if block:

```
return list.subList(index, list.size());
```

This if removes all empty buckets from start of the list.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
stefannegrea: Just above this line, there is a query to get the retention setting for the metric. The code in the if statement creates a TimeRange observable that is then fed into metric query that you highlighted. 

Also, the sample below is not good because it filters out all the empty buckets. We want to filter out only buckets that are empty at the start of the sequence; empty buckets in the middle or the end of the sequence do not need to be purged.

```
  .filter(bucket -> !bucket.isEmpty())
```

jsanda:
and to make ```skipeWhile``` as discussed, I think you need to do something like

```java
observableTimeRange.flatMap(range -> metricsService.findGaugeStats(metricId, range.getStart(),
        range.getEnd(), buckets, localPercentiles.getPercentiles()))
        .flatMap(Observable::from)
        .skipWhile(NumericBucketPoint::isEmpty)
        .toList()
        .map(ApiUtils::collectionToResponse)
        .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
```
I was thinking findGaugeStats returns Observable<NumericBucketPoint>, not Observable<List<NumericBucketPoint>>.

stefannegrea:
The default TTL is not saved along with the metric definition, only TTLs provided by users are saved. I would open another JIRA to resolve the TTL issue.

jsanda:
>The default TTL is not saved along with the metric definition, only TTLs provided by users are saved. I would open another JIRA to resolve the TTL issue.

I have not yet reviewed the last two or three commits, but when I last looked the same TTL calculation is being in different methods. Before merging this PR I would like to see that duplicate code refactored. I think it makes sense to do the retention calculations in the ```findMetric``` method.

stefannegrea:
@jsanda, to fix the retention being returned as null, I would have to track all the places where Metric is instantiated and conditionally update the value with the default from MetricsService if missing. It is doable, but do you want to do it in the context of this JIRA? That why I proposed creating a new JIRA for this issue.

The calculation that is being duplicated in two methods is to find the earliest date possible that can have data points stored for a metric based on the retention. Even if the TTL is always returned back as non-null, that calculation will still be in the handler method since it is not a MetricsService concern. But we could make it part of the Metric model. Do you want to go that route? Add a method on the Metric model class that returns the earliest date possible that can have data based on retention value?

jsanda:
>to fix the retention being returned as null, I would have to track all the places where Metric is instantiated and conditionally update the value with the default from MetricsService if missing.

Why can't you just update ```MetricsService.findMetric```?

>Even if the TTL is always returned back as non-null, that calculation will still be in the handler method since it is not a MetricsService concern. But we could make it part of the Metric model. Do you want to go that route?

No I do not want to go that route. We have remove any business logic out of model classes, and I prefer to keep it this way.

stefannegrea:
There are three methods to be updated: `MetricsService.findMetric`, `MetricsService.findMetrics`, `MetricsService.findMetricsWithFilters`. The last one will be updated automatically since it calls the first one; so all in all there are two methods to update. Will commit code shortly.

burmanm:
That will conflict with the PR #422

jsanda:
The verification logic in ```GaugeMetricStatisticsITest.fromEarliestWithData()``` is confusing. The code in question is,

```groovy
assertEquals(200, response.status)
assertEquals(4, response.data.size)

def expectedArray = [new BigDecimal(3), new BigDecimal(2), 'NaN', 'NaN'].toArray()
assertArrayEquals(expectedArray, response.data.min.toArray())
assertArrayEquals(expectedArray, response.data.max.toArray())
assertArrayEquals(expectedArray, response.data.avg.toArray())
```
The response should include four buckets and the second assert verifies that. The ```response.data.min```, ```response.data.max```, and ```response.data.avg``` is a nice Groovy feature that performs a mapping or transformation on the source list and returns back the specified property of each element in the list. The response includes four buckets; so, ```response.data.min``` returns a list containing the ```min``` property of each of the buckets. I find it confusing how ```expectedArray``` is used to compare values across all four buckets. Doing the verification like this would likely fail with different data sets. Can you clean this up some? 

I think everything else looks good.

stefannegrea:
I setup each bucket to contain exactly one data point, and thus the min, max, and avg are identical. Should I inline the expected array creation in each assert? Would that make it less confusing?

jsanda:
I looked at the actual JSON in the response, and the helps make it more clear for me. I'm fine without any further changes. Thanks for working through everything!

PR #421, opened 2015-12-30
[HWKMETRICS-333] Add support for encoded tag queries

burmanm:
Allows tags with ":" "/" and equal to be queried as long as they're encoded correctly in the URL. @tsegismont You had prevented this in the past in the Tags, any particular reason? I think it would get caught in the TagsConverter already as the splitting would fail / give empty value.


burmanm:
retest this please

jsanda:
This is a pretty minor code change that addresses an issue we need fixed, so I am going to merge. @tsegismont please follow up with discussion or another ticket if there is an issue with allowing reserved chars in tags.

PR #423, opened 2016-01-07
Metrics bus refactoring

jsanda:


stefannegrea:
Thanks @jsanda!

pom.xml
```diff
@@ -134,6 +135,21 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
```
tsegismont: @jsanda we had a long discussion back in July in order to avoid importing the Wildfly BOM in all modules of the project. Could you please remove this and import the BOM only where needed? As a recall, if you import the BOM here, then you force artifact versions for all sub-modules, even those not meant to run on Wildfly. In the past, this has broken the build and made me loose quite some time trying to find the issue.

PR #424, opened 2016-01-08
HWKMETRICS-334 - Limit and order for raw data queries

stefannegrea:


api/metrics-api-jaxrs/src/main/rest-doc/base.adoc
```diff
@@ -139,6 +139,41 @@ Either or both of them can be omitted. `start` defaults to _now_ minus 8 eight h
 /hawkular/metrics/gauges/paris01.web001.example.com.cpu.usage/data?end=20202020
 ----
 
+=== Limit and Order
+
+The API offers the ability to limit the number of data points returned as well specify the oder of the results.
```
mwringe: minor spelling issue: 'oder'

api/metrics-api-jaxrs/src/main/rest-doc/base.adoc
```diff
@@ -139,6 +139,41 @@ Either or both of them can be omitted. `start` defaults to _now_ minus 8 eight h
...
+
+# Explicit limit of 10, implicit time range, implicit descending order
+/hawkular/metrics/gauges/paris01.web001.example.com.cpu.usage/data?limit=10
+
+# Explicit limit of 10, explicit time range, implicit descending order
+/hawkular/metrics/gauges/paris01.web001.example.com.cpu.usage/data?start=10101010&end=20202020&limit=10
+
+# Explicit limit of 10, explicit start, implicit end, implicit ascending order
+/hawkular/metrics/gauges/paris01.web001.example.com.cpu.usage/data?start=20202020&limit=10
+----
```
mwringe: maybe add an example where the order is specified as a query parameter?

api/metrics-api-jaxrs/src/main/rest-doc/base.adoc
```diff
@@ -139,6 +139,41 @@ Either or both of them can be omitted. `start` defaults to _now_ minus 8 eight h
...
+* limit of 0 or negative is equivalent to no limit
+* if the user specifies limit, then the order is important because it returns data points up to limit in the specified order
+* start time and end time influence the default order if actual limit is specified
+ ** if only start time is specified then order is ascending
+ ** if only end time is specified then order is descending
+ ** if both are specified then order is descending
+ ** if none are specified then order is descending
+ ** order defaults are overridden if order specified
+* default order is descending if limit is not specified regardless of specified time range
+
```
mwringe: You might want to make it a bit clear what happens if you don't specify a start or endtime.
Will this return all the results? Or only ever a maximum of 8 hours (or whatever the limit is)

jsanda:
With these changes, we will now have a good bit of parameter parsing logic across all of the query endpoints. I think that we should implement a JAX-RS ParamConverter that encapsulates all of the query parameter logic which in turn eliminates duplication across endpoints and (hopefully) allows us to test the parameter parsing in isolation. We can tackle that in a separate ticket after this lands in master.

pilhuhn:
Are you sure that it is a good idea to depend the order depending on start/endtime being given? I'd prefer to always see ascending order unless overridden by an explicit 'order' parameter.

mwringe:
To me it makes sense if I were to specify ?startTime=X&limit=10 to give me the next 10 elements following time X. And ?endTime=Y&limit=10 to give me the 10 elements preceding time Y.

I think it might be a bit confusing if I specified ?startTime=X&limit=10 and it returns me the last 10 elements from the current point in time.

But I that is only my opinion. As long as we specify in the docs what is happening, then I think its fine to always default to one order type.

PR #425, opened 2016-01-08
[HWKMETRICS-332] Fix the path for multi-counter rate queries

stefannegrea:
/counters/rate path is now aligned with the path for rate queries with specified metric id.


mwringe:
LGTM, merged

PR #426, opened 2016-01-08
HWKMETRICS-338: only fetch the Cassandra seeds on the first invocatio…

mwringe:
…n and not on every invocation.


stefannegrea:
Thanks @mwringe!

PR #429, opened 2016-01-14
hawkular-parent 33

ppalaga:


ppalaga:
haw-perf-stability-test fails because Hawkular is not prepared for Metrics 11 yet.

@jsanda could you please review and eventually merge?

stefannegrea:
Thanks @ppalaga!

PR #430, opened 2016-01-14
[HWKMETRICS-340] remove unused virtual clock

jsanda:
I created a [branch](https://github.com/hawkular/hawkular-metrics/tree/virtual-clock) with the virtual clock stuff intact for future reference should we decide to use it again.


stefannegrea:
Thanks @jsanda!

PR #431, opened 2016-01-15
remove wildfly bom from parent pom

jsanda:
Based on prior discussions (which I had forgotten about) we agreed to keep the
wildfly bom out of the parent pom. Putting it in the parent pom forces artifact
versions for all sub-modules, including those not meant to run on WildFly. In
the past this has broken the build.

### DIFF
--- a/bus/pom.xml
+++ b/bus/pom.xml
@@ -42,6 +42,14 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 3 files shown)


jsanda:
@tsegismont would you review this please?

tsegismont:
@jsanda sure, thanks for the PR.

tsegismont:
LGTM. I will wait for the Travis green flag before merging, just in case.

PR #434, opened 2016-01-25
Purge JAX-RS 1.1 related code.

stefannegrea:


jsanda:
LGTM. I will wait for an ack from @mwringe before merging.

containers/pom.xml
```diff
@@ -112,8 +108,8 @@
       </activation>
       <modules>
         <module>cassandra-seed-provider</module>
-        <module>hawkular-metrics-openshift-jaxrs-1.1-integration</module>
-        <module>metrics-api-jaxrs-1.1-openshift</module>
+        <module>hawkular-metrics-openshift-integration</module>
+        <module>metrics-api-jaxrs-openshift</module>
       </modules>
```
tsegismont: What's the difference between the `openshift` profile and the `eap-containers` profile? Shouldn't we keep just one?

tsegismont:
I believe that, in addition to code removal, we should simplify the project module layout: we no longer need separate {{metrics-api-common}} {{metrics-api-util}} modules. Also, we should stop building the rests-tests tests JAR.

tsegismont:
When this PR will be merged, we must send an email to warn metrics contributors that they should remove the git pre-commit hook from their `.git/hooks` directory.

jsanda:
I think most of what is currently in metrics-api-common should go into metrics-api-jaxrs. The classes in metrics-api-common were those shared by the two jaxrs impls. The classes in metrics-api-util are classes shared by metrics-api-jaxrs and by hawkular-metrics-bus and the hawkular-metrics-component modules. We could probably use a better module/artifact name other than metrics-api-util.

stefannegrea:
All changes should be in now.

tsegismont:
Thanks @stefannegrea !

PR #436, opened 2016-01-28
Reapplied HWKMETRICS-69

tsegismont:
Reapplied HWKMETRICS-69


tsegismont:
I've applied the PR #432 diff to commit d23b70a and rebased on master.

PR #437, opened 2016-01-28
Some filters should be pre-matching filters

tsegismont:
Some filters should be pre-matching filters: they should be applied very early in the request processing pipeline


tsegismont:
Can some please review?

stefannegrea:
Thanks @tsegismont!

PR #439, opened 2016-01-29
Do not reject (yet) requests with Hawkular-Tenant

jpkrohling:


jpkrohling:
This is just a proposal, for clients of Metrics to adapt their calls. Instead of returning a 400 Bad Request, it will just log in ERROR a message. The old code is commented out, and should be used again at a specific point in time (couple of weeks from now?)

rhqci:

rhqci:

jpkrohling:
Talked with @tsegismont  , and this PR will probably get rejected. There was a thread on the -dev list about this change, and clients have to be updated *anyway*, so, better now than RELEASE_DAY-1 :)

jpkrohling:
I think @tsegismont wants the requests to fail, so, this PR is not going to be merged. I'm closing it, but if something like this is needed, I can adjust the message.

PR #440, opened 2016-01-29
Remove last lingering api-common references.

stefannegrea:


tsegismont:
Thanks @stefannegrea !

PR #441, opened 2016-02-03
HWKMETRICS-343 Grafana should be able to connect to a Metrics server …

tsegismont:
…protected by Accounts

The server now supports InfluxDB request parameter based authentication
(with an Undertow extension).

The database name is used to create the Hawkular-Persona header. This
allows to select the authenticated user tenant or an organization tenant.

Also, CORS filters are now removed from the component. CORS is managed
by Keycloack within Hawkular.

### DIFF
--- a/hawkular-component/pom.xml
+++ b/hawkular-component/pom.xml
@@ -68,14 +68,6 @@
       <artifactId>hawkular-accounts-api</artifactId>
       <version>${version.org.hawkular.accounts}</version>
     </dependency>
-    <dependency>
-      <groupId>org.jboss.logging</groupId>
-      <artifactId>jboss-logging-annotations</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.jboss.logging</groupId>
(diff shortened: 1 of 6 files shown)


tsegismont:
@jpkrohling would you mind to review?

jpkrohling:
Looks good to me! I particularly like how you avoid the exception on Accounts for invalid persona ID. +1 to that!

jpkrohling:
Talking about exceptions, *if* you want to prevent the one thrown from `UUID#fromString(String)` in some cases, you could try this:

https://github.com/jpkrohling/secret-store/blob/master/secret-store-api/src/main/java/org/keycloak/secretstore/api/RequestRewriter.java#L106-L110

Not sure which cost would be higher for you here: maintaining this, or catching an occasional exception on `UUID#fromString(String)`

tsegismont:
Cool, thanks for the review.

tsegismont:
Will do this in another PR for you to review, coming in a few minutes :)

PR #444, opened 2016-02-05
HWKMETRICS-346 Fix bogus Hawkular Component security rules

tsegismont:
HWKMETRICS-346 Fix bogus Hawkular Component security rules

Do not use declarative security in web.xml.

Note that the REST endpoints are still all secured by default, only a couple are white-listed for free access.

### DIFF
--- a/hawkular-component/src/main/java/org/hawkular/metrics/security/PersonaFilter.java
+++ b/hawkular-component/src/main/java/org/hawkular/metrics/security/PersonaFilter.java
@@ -44,17 +44,17 @@ import org.hawkular.metrics.model.ApiError;
 @Provider
 @PreMatching
 public class PersonaFilter implements ContainerRequestFilter {
-    public static final String TENANT_HEADER_NAME = "Hawkular-Tenant";
-    public static final String TENANT_HEADER_NOT_ALLOWED = "The " + TENANT_HEADER_NAME + " header is not allowed. " +
+    private static final String TENANT_HEADER_NAME = "Hawkular-Tenant";
+    private static final String TENANT_HEADER_NOT_ALLOWED = "The " + TENANT_HEADER_NAME + " header is not allowed. " +
             "The tenant is determined from the credentials supplied with the request";
 
     @Inject
-    Instance<Persona> persona;
+    Instance<Persona> personaInstance;
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
         String path = requestContext.getUriInfo().getPath();
-        if (path.startsWith("/status") || path.equals("/")) {
+        if (path.equals("/status") || path.equals("/")) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


tsegismont:
The (minor IMO) issue with this change is that errors like this show up in the logs for request without credentials:
```
10:54:04,570 ERROR [org.hawkular.accounts.api.internal.impl] (default task-1) HAWKACC100006: There's an incoming request for an non-secure location, but a component has requested information about the current persona. This probably indicates a mismatch in the integration of the component. Either the component should accept non-secure requests and not ask for a Persona, or should only accept secure requests. Hawkular Accounts is returning NULL to the component. Lookout for NullPointerExceptions!
```

tsegismont:
@jpkrohling here is the pull request I was talking about in #441. Would you mind to review?

hawkular-component/src/main/java/org/hawkular/metrics/security/InfluxAuthHttpHandler.java
```diff
@@ -68,9 +73,7 @@ public void handleRequest(HttpServerExchange httpServerExchange) throws Exceptio
         HeaderMap requestHeaders = httpServerExchange.getRequestHeaders();
 
         String databaseName = matcher.group(1);
-        try {
```
jpkrohling: The regex above is not meant to replace this try/catch. It might be the case that it matches the regex but still is not spec-compliant. In such case, Accounts will throw a RuntimeException, so, it's a matter of taste now :)

hawkular-component/src/main/java/org/hawkular/metrics/security/InfluxAuthHttpHandler.java
```diff
@@ -41,6 +40,12 @@
  * @author Thomas Segismont
  */
 public class InfluxAuthHttpHandler implements HttpHandler {
+    private static final Pattern UUID_PATTERN;
+
+    static {
```
jpkrohling: Just out of curiosity: why is this wrapped into a static block? You have only one statement and there's no exception to catch, so, it has the same effect as if you were assigning directly at the declaration above, no ?

jpkrohling:
Review done. If you are OK with Accounts throwing an exception should the UUID not be a valid UUID, then this looks good to me. The comment about the static block was just out of curiosity.

jpkrohling:
About the ERROR log entry, we could discuss about having an "ACK" flag on the `web.xml`, such as, `org.hawkular.accounts.ignore-missing-persona-warning`.

hawkular-component/src/main/java/org/hawkular/metrics/security/InfluxAuthHttpHandler.java
tsegismont: OK, I'll simply revert the second commit then.

hawkular-component/src/main/java/org/hawkular/metrics/security/InfluxAuthHttpHandler.java
tsegismont: No reason except that it keeps things in one line. No too much important 
to me.

tsegismont:
Thanks for the review, I'll ping you when the revert/rebase is ready.

tsegismont:
That would be just perfect :)

PR #459, opened 2016-02-19
HWKMETRICS-342 ptrans should work with a Metrics server protected by Accounts

tsegismont:
Added new config properties:
- server.type: metrics/hawkular
- auth.id
- auth.secret
- persona.id

tenant config property is now only used when server.type == metrics

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/JacksonConfig.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/JacksonConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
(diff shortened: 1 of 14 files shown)


tsegismont:
retest this please

tsegismont:
Would you mind to review @pilhuhn ?

tsegismont:
Can someone please review?

tsegismont:
Can we get this merged before 0.13 is released? The PR has been available for review for a week now. Thanks.

jsanda:
A couple questions. 

First, do we need to worry about accounts integration any more? 

Second, could you add an integration test?

tsegismont:
Le 26/02/2016 15:42, jsanda a écrit :
> A couple questions.
>
> First, do we need to worry about accounts integration any more?

I'm not aware of any plans to stop securing Hawkular with Accounts. If 
that changes, then I'll update ptrans as needed. In any case, we're just 
authenticating with HTTP Basic auth, so it shouldn't change much if 
Accounts is involved or not.

>
> Second, could you add an integration test?

I tested this code manually but chose not add such an integration test 
because, IMHO the cost (start both a Metrics and a Hawkular server 
during the build) outweighs the benefit.

tsegismont:
Thank you John.

PR #462, opened 2016-02-19
HWKMETRICS-350: update required to function properly with the latest Openshift

mwringe:


stefannegrea:
Thanks @mwringe! 

Will not wait for performance tests since this is an old release/branch.

PR #464, opened 2016-02-26
HWKMETRICS-356 Use RxJavaGuava instead of RxUtil

tsegismont:
HWKMETRICS-356 Use RxJavaGuava instead of RxUtil

hawkular-metrics-rx-java-driver module has an RxUtil class, which was copied from https://github.com/ReactiveX/RxJavaGuava, because in the beginning the RxJavaGuava JAR was not in Maven Central.

Now that it is released in Central, we should use it instead of maintaining the code here.


tsegismont:
Rebased on master. Can someone please review?

jsanda:
The changes look good, but looks like travis failed with unrelated integration test failure for tests that run with the full hawkular server. I might need to tweak things for those tests to get them running more reliably. I restarted travis and will merge when travis is happy.

tsegismont:
Thank you John!

PR #466, opened 2016-03-01
There is no reason to manage io.fabric8:fabric8-arquillian

ppalaga:


ppalaga:
Travis failed. I restarted the job to see if it can be reproduced.

PR #469, opened 2016-03-16
HWKMETRICS-364

jsanda:


ppalaga:
Both distros [1] and [2] are putting the metrics war to `standalone/deployments`. If there is a chance that any of them should ever be productized, `standalone/deployments` is not a good choice. AFAIK, the underlying reason is that the patching mechanism is programmed not to touch any apps in `standalone/deployments` because they are supposed not to come from Red Hat. This is the reason why Nest was created. cc @jmazzitelli 

[1] https://github.com/hawkular/hawkular-metrics/pull/469/commits/8564769f2900b604d7bef39dc793335003fefcbf#diff-b2d53e62482bea9fe1f7284ab713f820R82
[2] https://github.com/hawkular/hawkular-metrics/pull/469/commits/8564769f2900b604d7bef39dc793335003fefcbf#diff-aa3159fcb2c5b22b0ef78d17ac7e1a0bR88

distros/wildfly-standalone/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+             <includeArtifactIds>hawkular-metrics-api-jaxrs</includeArtifactIds>
+             <overWriteIfNewer>true</overWriteIfNewer>
+             <excludeTransitive>true</excludeTransitive>
+           </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-assembly-plugin</artifactId>
+        <version>2.5.4</version>
```
ppalaga: Please remove `<version>2.5.4</version>` from here and add an explicit `groupId`. If you need to override the version 2.6 managed in hawkular-parent, please rather override the `version.maven-assembly-plugin` property with a proper comment why you need it.

distros/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.hawkular.metrics</groupId>
+    <artifactId>hawkular-metrics-parent</artifactId>
+    <version>0.14.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hawkular-metrics-distros</artifactId>
```
ppalaga: Is `distros` name of the module and directory neutral enough? It sounds a little bit too casual to my non-native ears.

How about `hawkular-metrics-dist-parent` or simply `hawkular-metrics-dist` ?

distros/wildfly-standalone/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.hawkular.metrics</groupId>
+    <artifactId>hawkular-metrics-distros</artifactId>
+    <version>0.14.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>hawkular-metrics-wildfly-standalone</artifactId>
```
ppalaga: Is module stores a distribution and therefore I vote for a name that contains `dist` on some place, e.g. `hawkular-metrics-standalone-dist`. `wildfly` does not need to be there in the name unless there exists a plan to have a non-wf distro.

jsanda:
> Both distros [1] and [2] are putting the metrics war to standalone/deployments. If there is a chance that any of them should ever be productized, standalone/deployments is not a good choice. AFAIK, the underlying reason is that the patching mechanism is programmed not to touch any apps in standalone/deployments because they are supposed not to come from Red Hat. This is the reason why Nest was created. cc @jmazzitelli 

The motivation for creating these distros is primarily for improved developer/community member experience. As for productization, we have OpenShift and ManageIQ that we need to worry about. Packaging is already handled for both.

distros/wildfly-standalone/pom.xml
jsanda: > unless there exists a plan to have a non-wf distro.

I would not say so much a plan but more a desire.

ppalaga:
Both distros unzip `wildfly-dist` zip, which triggers quite a big download after every WildFly version upgrade. 

Alternativelly, one could extend the wildfly-feature-pack and use it to build these distros. With feature packs, version upgrades trigger download of only those parts of the distro that really changed compared to the previous version, which usually means much less bits than the whole wildfly-dist.zip. I am ready to prepare the feature-pack based setup if you are interested.

[1] https://github.com/hawkular/hawkular-metrics/pull/469/commits/59b71ec2b42029c3f316b44ff3c87ab74544a26f#diff-b2d53e62482bea9fe1f7284ab713f820R51

ppalaga:
I vote for using the standalone dist in the existing integration tests.

It would be nice if the distro with embedded cassandra would also be itested somehow - perhaps not by default, but there should at least be a profile that would have to be run at least before a release.

distros/pom.xml
jsanda: I know that other hawkular projects follow the naming convention of hawkular-component-module-name. I am not opposed to this convention; however, metrics does not follow it.

jsanda:
> I am ready to prepare the feature-pack based setup if you are interested.

That would be great!

distros/pom.xml
ppalaga: > convention of hawkular-component-module-name. I am not opposed to this convention; however, metrics does not follow it.

My proposal was not about strictly following any convention, but rather about replacing `distros` with `dist`, preferring a standard word to a slang expression.

stefannegrea:
Thanks @jsanda!

PR #473, opened 2016-04-01
[HWKMETRICS-24] 

stefannegrea:
/data endpoint became too complex to maintain since it was serving two purposes (retrieve raw and statistical data). This is also confusing to users because the decision as to what data to return was made at runtime based on the parameters.

Here are the changes in this commit:
- all /data APIs have been deprecated
- POST /data has been moved to POST /raw
- GET $id/data has been split into two
  - $id/raw - to retrieve raw data
  - $id/stats - to retrieve bucketed/statistical data
- Documentation for deprecated APIs has been updated to point to the new APIs.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
@@ -242,7 +242,7 @@ public class CounterHandler {
     @Deprecated
     @POST
     @Path("/data")
-    @ApiOperation(value = "Deprecated. Please use /raw endpoit.")
+    @ApiOperation(value = "Deprecated. Please use /raw endpoint.")
     public void deprecatedAddData(
             @Suspended final AsyncResponse asyncResponse,
             @ApiParam(value = "List of metrics", required = true) List<Metric<Long>> counters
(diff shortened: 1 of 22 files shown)


jsanda:
We have an endpoint that returns rate stats, but not one that returns the rate data points by themselves. For consistency and in case we and an endpoint to return the rate data points, should we change the current endpoint from ```/counters/$id/rate``` to ```/counters/$id/rate/stats```?

stefannegrea:
@jsanda did not notice the `/rate` endpoint. I think we should migrate the single endpoint to two endpoints `/counters/$id/rate/raw` and `/counters/$id/rate/stats` because that single method is having today that same behavior as `/data` (sometimes bucketed and sometimes not).

jsanda:
Looking ahead to HWKMETRICS-373 which add supports for tag-based bucketing, what do you think the  API should be for that? What about something like ```/gauges/$id/stats/tags/x:*,y:1|2```.

stefannegrea:
@jsanda, that is a really good suggestion for the adding support for stats on tagged data. An alternative would be `/gauges/$id/tags/stats/x:*,y:1|2.`. But I like your suggestion more since the tag related stats are intuitively a sub-category to simple stats, so your proposal looks better.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
```diff
@@ -541,22 +541,11 @@ public void findStatsCounterData(
...
-            "rates when resets occur.", response = DataPoint.class, responseContainer = "List")
-    @ApiResponses(value = {
-            @ApiResponse(code = 200, message = "Successfully fetched metric data."),
-            @ApiResponse(code = 204, message = "No metric data was found."),
-            @ApiResponse(code = 400, message = "buckets or bucketDuration parameter is invalid, or both are used.",
-                    response = ApiError.class),
-            @ApiResponse(code = 500, message = "Unexpected error occurred while fetching metric data.",
-                    response = ApiError.class)
-    })
+    @Path("/{id}/ratel")
```
jsanda: Is this a typo?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
```diff
@@ -749,4 +824,24 @@ public void findCounterRateDataStats(
                     .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
         }
     }
+
+    @Deprecated
+    @GET
+    @Path("/ratel")
```
jsanda: Is ```ratel``` a typo?

PR #483, opened 2016-04-19
[HWKMETRICS-197] Add ability to fetch tag values

burmanm:
This feature adds the ability to fetch only the tag values (ignoring metric definitions) using the same tag-query-language that we have currently in the metric definition fetching.

WIP: Missing REST-tests, I'm suggesting @Path("/tags/{tags}") (for all types separately and then in the metrics for fetching all the metrics)

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
@@ -168,6 +168,22 @@ public class AvailabilityHandler {
     }
 
     @GET
+    @Path("/tags/{tags}")
+    @ApiOperation(value = "Retrieve gauge type's tag values", response = Map.class)
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Tags successfully retrieved."),
+            @ApiResponse(code = 204, message = "No matching tags were found"),
+            @ApiResponse(code = 500, message = "Unexpected error occurred while fetching tags.",
+                    response = ApiError.class)
+    })
+    public void getTags(@Suspended final AsyncResponse asyncResponse,
+                        @ApiParam("Tag query") @PathParam("tags") Tags tags) {
+        metricsService.getTagValues(tenantId, AVAILABILITY, tags.getTags())
+                .map(ApiUtils::mapToResponse)
+                .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
+    }
+
+    @GET
     @Path("/{id}/tags")
     @ApiOperation(value = "Retrieve tags associated with the metric definition.", response = String.class,
                   responseContainer = "Map")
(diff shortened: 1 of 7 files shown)


burmanm:
retest this please

jsanda:
I realize that this is a WIP, but I have a question and a couple comments about the commit. In the new ```MetricsService.getTagValues``` method, why are we calling ```dataAccess.findMetricsByTagName```? Maybe I am missing something, but it seems like we are doing a lot of extra, unnecessary work. Couldn't we simply do ```SELECT tvalue FROM metrics_tags_idx WHERE tenant_id = ? AND tname = ?```
``` for each tag key and then call ```tagFilterValue``` and then build the sets.

Building the sets brings me to my next comment. I have started thinking more about our API in terms of scalability. We build up results completely in memory before sending them to the client which makes us a lot more susceptible to heap pressure and OOMs when there are higher read loads. I wrote up some thoughts on this in [HWKMETRICS-392](https://issues.jboss.org/browse/HWKMETRICS-392). Building the sets that are essentially unbounded gave me pause. 

Lastly, I would really prefer to see a new test method rather than testing the new functionality in the ```createAndFindMetricsWithTags``` test.

**Comment by burmanm** [comment]
@jsanda No, this is the case what I tried to explain in the IRC. The metric name is the only way to link two different set of tags together. Our tags query language has "AND", but that can't be applied without knowing which tags are connected to each other.

So I need to fetch the metricName to see that the metricId has all of the requested tags.

**Comment by jsanda** [comment]
I remember the conversation. I am just not sure it makes sense after having thought about it some. Why can't we look only at the tag values?

**Comment by burmanm** [comment]
@jsanda Imagine a case where someone is fetching data from 1000 hostnames, some are in "QA", some are in "test" and some are in "prod". Now lets say you needed to fetch all the hostnames for a list that includes only the "prod" environment.

Now you could do: ``hostname:*,env:prod``. If we only fetch the available values, it's impossible to give restrictions other than the single value.

It also creates tags with yet another query language that's not compatible with the other ones. I actually had an implementation that did what you asked, but I deleted it already to keep the consistency (as @stefannegrea wished in the IRC)

**Comment by burmanm** [comment]
I give up with Travis, it's complaining about ptrans test failures which have nothing to do with this PR and are most likely failing only because of Travis.

**Comment by jsanda** [comment]
Thanks for the example. That helps clarify things.

**Comment by FilipB** [comment]
retest this please

**Comment by burmanm** [comment]
Don't merge yet, I actually found a bug in our current tags handling which I'd like to fix also.. we don't handle empty requests well (reduce function can't be applied to empty observable)

**Comment by burmanm** [comment]
Squashed commits, ready for review/merge.

**Comment by burmanm** [comment]
retest this please

**Comment by jsanda** [comment]
I already merged your other PR for removing use of Optional. Looks like you need to rebase off of master now.

**Comment by jsanda** [comment]
I think the changes look good and we can merge after you rebase. I would also like a walk through of the new MetricsServiceImpl.getTagValues method to better understand the data flow. I do have one question right now. For the if statement at line 541 in the first reduce:

```java
reduce((map1, map2) -> {
                            map2.forEach((key, subMap) -> {
                                if (map1.containsKey(key)) {
                                    Map<String, String> tagMap = map1.get(key);
                                    // There can't be duplicate keys (other than same value)
                                    subMap.forEach(tagMap::put);
                                } else {
                                    map1.put(key, subMap);
                                }
                            });
                            return map1;
                        }))
``` 

is ```map.containsKey(key)``` ever going to be true? If I understand the code (and I'm not entirely sure I do), aren't we at this point just combing values for a given tag key? If so, then the map keys would necessarily be distinct because correspond to the metric ids which are the clustering columns.

burmanm:
Rebased, and no, it's not going to be true ever (refactored that part).

PR #484, opened 2016-04-21
Refactor getMetricTags to remove use of Optional

burmanm:
The combination of Optional + Observable is useless as there's nothing extra the Optional could present in this case. This removes it and cleans up the code.


burmanm:
retest this please

FilipB:
retest this please

PR #490, opened 2016-04-26
HWKMETRICS-385 Deleting tags should only require the tag keys and not the values

tsegismont:
HWKMETRICS-385 Deleting tags should only require the tag keys and not the values

Added a new JAX-RS param class. It allows to map a list of plain names mixed with name:value pairs (for backwared compatibility)
Updated the core service implementation so that tag values are loaded and from metric index. This is needed otherwise we can't delete rows from the tags index table (tvalue is part of the primary key).

### DIFF
--- a/api/metrics-api-jaxrs/pom.xml
+++ b/api/metrics-api-jaxrs/pom.xml
@@ -252,13 +252,6 @@
           <plugin>
             <groupId>org.codehaus.gmavenplus</groupId>
             <artifactId>gmavenplus-plugin</artifactId>
-            <dependencies>
-              <dependency>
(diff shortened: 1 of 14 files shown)


tsegismont:
@jsanda I had to reopen a PR because my last changes were not included. I updated the doc and remove groovy-all version as it is managed in parent POM.

tsegismont:
It's the second time you close my PR GitHub. Stop it.

tsegismont:
Rebased on master

tsegismont:
@jsanda all green. Could you merge? Thanks!

PR #492, opened 2016-04-26
Add untyped generic endpoint for tagValues fetching

burmanm:
Adds the generic "/metrics/tags/{tags}" endpoint to fetch tag values. This extends the HWKMETRICS-197 implementation.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -116,7 +116,7 @@ public class MetricHandler {
 
     @GET
     @Path("/tags/{tags}")
-    @ApiOperation(value = "Retrieve gauge type's tag values", response = Map.class)
+    @ApiOperation(value = "Retrieve metrics' tag values", response = Map.class)
     @ApiResponses(value = {
             @ApiResponse(code = 200, message = "Tags successfully retrieved."),
             @ApiResponse(code = 204, message = "No matching tags were found"),
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
@@ -27,6 +27,7 @@ import static org.hawkular.metrics.model.MetricType.COUNTER;
 import static org.hawkular.metrics.model.MetricType.GAUGE;
 
 import java.net.URI;
+import java.util.Map;
 import java.util.regex.PatternSyntaxException;
 
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


burmanm:
retest this please

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -113,6 +115,24 @@
     }
 
     @GET
+    @Path("/tags/{tags}")
+    @ApiOperation(value = "Retrieve gauge type's tag values", response = Map.class)
```
tsegismont: Copy/paste? gauge -> metric

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/rest/TagsITest.groovy
```diff
@@ -297,17 +297,27 @@ class TagsITest extends RESTTest {
...
+
+      def untypedResponse = hawkularMetrics.get(path: "metrics/tags/a1:*,d1:B%3AA", headers: [(tenantHeaderName):
+                                                                                                  tenantId])
+
+      [response, genericResponse, untypedResponse].each { lresponse ->
+        assertEquals(200, lresponse.status)
+        assertEquals([
+            a1: ['a', 'A/B'],
+            d1: ['B:A']
+        ], lresponse.data)
```
tsegismont: If the result is always the same, it means the user get the metric name but not the type when querying without the type info, correct? Can you elaborate on the use case?

burmanm: No, the user does not get any metric name either, only the possible tag values. For this you could think that you want all the hostnames queried, but all of them don't have say gauge metric definitions. So just ask from all the metric types (as duplicates are removed).

In this test the reply is the same, but it would be different if there were different tag values on another type.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
burmanm: Yeah. I'll fix that.

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/rest/TagsITest.groovy
tsegismont: Got you. Thanks

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
tsegismont: Could you fix that today so that this PR is included in the next release?

stefannegrea:
Thanks @burmanm!

PR #493, opened 2016-04-26
[HWKMETRICS-148]

stefannegrea:
Update metrics creation to allow users to overwrite previously created metric definitions. The flag defaults to false and is not required.


jsanda:
The ticket mentions updating the operations for creating tenants as well as metrics. I would like to keep the behavior consistent across operations to the extent possible.

jsanda:
It can be a little confusing that in the createMetric method we do ```!overwrite && !resultSet.wasApplied()``` but in the createTenant method we are just checking ```!resultSet.wasApplied()```. It turns out that because the overwrite logic is in DataAccessImpl, this works out ok. If the query is not a LWT, then wasApplied() will always return true.

jsanda:
Looks like there are some REST test failures. Once those are fixed, I think this is ready to merge.

stefannegrea:
The Travis error is really confusing because one build passed and the other failed.

stefannegrea:
I am still surprised why one Travis build failed and other did not (nor did it fail locally prior to this). I updated the test code to show the line for errors and also rebased on latest master.

PR #496, opened 2016-04-28
[HWKMETRICS-384] Add string metric type

jsanda:


stefannegrea:
From a safety perspective, there needs to be a limit on the string length. @jsanda, do you mind adding that?

stefannegrea:
@jsanda, the `/strings` endpoint is not comparable with other endpoints in terms of functionality. What is the plan there?

jsanda:
I realize that the available endpoints is not consistent with what we have for other metric types. That is intentional. With the `/stats` endpoint for example, it is too early to know what those should look like for string metrics. We could take a guess and put something out there. I would rather get a little feedback to help determine what exactly we want/need.

I was even looking at the swagger annotations to see if there is something like ```@Experimental``` so that we can indicate that these new endpoints may have breaking changes.

I am not sure I see a need to create tickets unless we know specifically what functionality we want. The docs should be very clear though about where there are inconsistencies in the API and why.

stefannegrea:
I like the idea with `experimental`. We can add a sentence as the `description` on the top `@Api` annotation.

stefannegrea:
Thanks @tsegismont!

jsanda:
I just pushed changes to set and enforce a max size for strings. The limit is stored in the new system_settings table. Right now there is no way to change the limit without directly updating that table. I will open up a separate ticket to discuss and address that. I also updated the ```@Api``` annotation in StringHandler, adding a description that says it is experimental.

stefannegrea:
Thanks @jsanda!

PR #497, opened 2016-04-29
HWKMETRICS-390 Endpoint for fetching rate data points need to support standard query params and sort behavior

tsegismont:
I added a method in Order.java which factors out the logic to determine a default order. I updated the considering endpoints.

I change the get rate endpoint to support the new parameters. While its bucket parameters are deprecated and not removed, we need to check that they can't be set if order and limit are.

While updating the code in MetricsServiceImpl to adapt to the new  findRate signature, I extracted the code of #bucketize methods and created separated Transformer classes. Using these transformers with #compose is a more natural RxJava construct.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/Order.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/Order.java
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.hawkular.metrics.core.service;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -53,6 +54,18 @@ public enum Order {
             throw new IllegalArgumentException(text + " is not a recognized order");
         }
         return order;
+    }
+
+    /**
+     * Determines a default value based on API parameters.
+     *
+     * @param limit maximum of rows returned
+     * @param start time range start
+     * @param end   time range end
+     * @return {@link #ASC} if limit is positive, start is not null and end is null; {@link #DESC} otherwise
+     */
+    public static Order defaultValue(int limit, Long start, Long end) {
+        return limit > 0 && start != null && end == null ? ASC : DESC;
     }
 
     @Override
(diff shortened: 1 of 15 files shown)


tsegismont:
Please do not merge yet. I created the PR to gather feeback early. I need to had a test for rate in descending order

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/Order.java
```diff
@@ -55,6 +56,18 @@ public static Order fromText(String text) {
...
 
+    /**
+     * Determines a default value based on API parameters.
+     *
+     * @param limit maximum of rows returned
+     * @param start time range start
+     * @param end   time range end
+     * @return {@link #ASC} if limit is positive, start is not null and end is null; {@link #DESC} otherwise
+     */
+    public static Order defaultValue(int limit, Long start, Long end) {
```
jsanda: This is confusing. Why not simply default to either ascending or descending?

tsegismont: Because there is no simple default :)

Frankly I believe this is a bit confusing as well. But since the code for
that default was present in all the endpoints, I assumed it came from some
requirement. Maybe Stefan knows better ?

stefannegrea:
Postponing this PR to the next release.

tsegismont:
@stefannegrea @jsanda I just updated the findRate test and rebased on master

tsegismont:
@burmanm I believe you wrote the code which is now in TaggedBucketPointTransformer and added the `TODO refactor this to be more functional and replace java 8 streams with rx operators`. In this case can you please have a look at c3ac260 ? Is this what you were thinking about?

tsegismont:
Rebased on master

tsegismont:
retest this please

burmanm:
@tsegismont I think it was @jsanda who wrote that part.

jsanda:
The refactoring in TaggedBucketPointTransformer looks good. Thanks @tsegismont!

PR #498, opened 2016-05-02
[HWKMETRICS-398] Introduce new configuration service

jsanda:


jsanda:
This PR introduces a simple service configuration service. Even though the service is currently only used by MetricsServiceImpl, I put it in a new, separate module because the job scheduler I am working on for HWKMETRICS-360 will also depend on it (as soon as I rebase in my job scheduler branch.) The job scheduler has no dependency on MetricsService, hence the new module for the configuration service.

core/configuration-service/src/main/java/org/hawkular/metrics/sysconfig/ConfigurationService.java
```diff
@@ -0,0 +1,73 @@
...
+    // I could have just as easily passed the session as a constructor arg. I am doing it in the init method because
+    // eventually I would like service initialization async.
+    public void init(RxSession session) {
+        this.session = session;
+        findConfiguration = session.getSession().prepare("SELECT name, value FROM sys_config WHERE config_id = ?");
+        updateConfiguration = session.getSession().prepare(
+                "INSERT INTO sys_config (config_id, name, value) VALUES (?, ?, ?)");
+    }
+
+    public Observable<Configuration> load(String id) {
```
burmanm: If this is always a single returned value, we should use the Single<Configuration> instead of Observable. This was introduced in some RxJava post-1.0 release. http://reactivex.io/documentation/single.html

core/configuration-service/src/main/java/org/hawkular/metrics/sysconfig/ConfigurationService.java
```diff
@@ -0,0 +1,73 @@
...
+        return session.executeAndFetch(findConfiguration.bind(id))
+                .map(row -> {
+                    Map<String, String> map = new HashMap<>();
+                    map.put(row.getString(0), row.getString(1));
+                    return map;
+                })
+                .reduce((m1, m2) -> {
```
burmanm: Wouldn't it make sense to use toMap() instead of .map().reduce() ? Avoids some object creation.

core/configuration-service/src/main/java/org/hawkular/metrics/sysconfig/ConfigurationService.java
jsanda: The Single class is marked experimental which presumably means it could go away in future releases. There are other places where this would be good to use like on inserts where we return Observable<Void>, but do we want to depend on experimental features that may go away?

jsanda: I am not necessarily opposed to using it, but if we decide to use it, then I prefer creating another ticket to do the refactoring everywhere.

jsanda:
@burmanm asked a good question - should we have a type configuration object rather than a Map<String, String>. I think we might. I did not see the need for it now and decided that can be introduced later as necessary. I will write up some more thoughts on the ticket.

stefannegrea:
Thanks @jsanda!

PR #500, opened 2016-05-04
HWKMETRICS-383 Allow fetch min and max timestamp for a metric

tsegismont:
Added a MinMaxTimestampTransformer which loads min and max timestamp for each metric in a sequence.
Applied the transformer to all endpoints fetching metric definitions.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
@@ -50,6 +50,7 @@ import javax.ws.rs.core.UriInfo;
 
 import org.hawkular.metrics.api.jaxrs.handler.observer.MetricCreatedObserver;
 import org.hawkular.metrics.api.jaxrs.handler.observer.ResultSetObserver;
+import org.hawkular.metrics.api.jaxrs.handler.transformer.MinMaxTimestampTransformer;
 import org.hawkular.metrics.api.jaxrs.util.ApiUtils;
 import org.hawkular.metrics.core.service.Functions;
 import org.hawkular.metrics.core.service.MetricsService;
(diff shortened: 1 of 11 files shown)


tsegismont:
Note for reviewers: in case you wonder why `response.data` needs to become `(response.data as List).sort({ m1, m2 -> m1.id.compareTo(m2.id)`, the server sends requests for min and max timestamp asynchronously, so there is no guarantee with respect to the order of metrics returned.

lucasponce:
@tsegismont I have reviewed and with this change we could improve the MiQ code. 
Today we have a use case that is:
def min_max_timestamps(client, metric_id)
 min = client.get_data(metric_id, :starts => 0, :limit => 1, :order => 'ASC')
 max = client.get_data(metric_id, :starts => 0, :limit => 1, :order => 'DESC')
 [min[0]['timestamp'], max[0]['timestamp']]
end
Which will be reduced with this new change in a single call.
Thanks.

tsegismont:
Can someone please review?

jsanda:
I think the changes look good. I have one question.  I realize that the feature request was only for the timestamps, but would it make sense to include the whole data points rather than just the timestamps?

tsegismont:
Not sure. Especially for counters which values are often not useful per se.

2016-05-11 15:32 GMT+02:00 jsanda <notifications@github.com>:

> I think the changes look good. I have one question. I realize that the
> feature request was only for the timestamps, but would it make sense to
> include the whole data points rather than just the timestamps?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/hawkular/hawkular-metrics/pull/500#issuecomment-218459782>
>



-- 
Thomas Segismont
JBoss ON Engineering Team

jsanda:
It definitely depends. I know with APIMan the raw counts are useful. I think the changes are good and I am fine with merging. If we find that users want the whole data points we can introduce those changes if/when necessary.

PR #501, opened 2016-05-09
HWKMETRICS-365 Support rate data for gauge metrics

tsegismont:
Added rate related endpoints to GaugeHandler
Added GAUGE_RATE internal type (also, added missing STRING type to the "all" set in MetricType)
Changed MetricsService API so that rate and rate stats work on Metric<? extends Number> instead of Metric<Long>
Updated multiple metric stats algorithms to support GAUGE_RATE

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
@@ -23,6 +23,8 @@ import static org.hawkular.metrics.api.jaxrs.filter.TenantFilter.TENANT_HEADER_N
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.badRequest;
 import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.serverError;
 import static org.hawkular.metrics.model.MetricType.GAUGE;
+import static org.hawkular.metrics.model.MetricType.GAUGE_RATE;
+import static org.hawkular.metrics.model.MetricType.UNDEFINED;
 
 import java.net.URI;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
```diff
@@ -724,4 +725,164 @@ public void findPeriods(
...
+            return;
+        }
+
+        MetricId<Double> metricId = new MetricId<>(tenantId, GAUGE, id);
+
+        if (limit == null) {
+            limit = 0;
+        }
+        if (order == null) {
+            order = Order.defaultValue(limit, start, end);
```
burmanm: Why isn't OrderConverter forcing these default values and avoiding null?

tsegismont: Because we need to know the value of the limit parameter. Can you access
other param values in a converter ?

Besides, I'm not sure a converter is the right place to store a default
value logic. It should simply convert a string to another string. Actually,
if the param is absent, I don't think RestEasy will invoke the converter.
When possible, a @Default annotation with the string representation is
better, I think.

2016-05-10 12:02 GMT+02:00 Michael Burman <notifications@github.com>:

> In
> api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/GaugeHandler.java
> <https://github.com/hawkular/hawkular-metrics/pull/501#discussion_r62643773>
> :
>
> > +            @ApiParam(value = "Limit the number of data points returned") @QueryParam("limit") Integer limit,
> > +            @ApiParam(value = "Data point sort order, based on timestamp") @QueryParam("order") Order order
> > +    ) {
> > +        TimeRange timeRange = new TimeRange(start, end);
> > +        if (!timeRange.isValid()) {
> > +            asyncResponse.resume(badRequest(new ApiError(timeRange.getProblem())));
> > +            return;
> > +        }
> > +
> > +        MetricId<Double> metricId = new MetricId<>(tenantId, GAUGE, id);
> > +
> > +        if (limit == null) {
> > +            limit = 0;
> > +        }
> > +        if (order == null) {
> > +            order = Order.defaultValue(limit, start, end);
>
> Why isn't OrderConverter forcing these default values and avoiding null?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/hawkular/hawkular-metrics/pull/501/files/8dc33a2df7c4b5e24ee658da82928ea1620b20ce#r62643773>
>



-- 
Thomas Segismont
JBoss ON Engineering Team

tsegismont:
Rebased on master

PR #502, opened 2016-05-11
Bind metrics service to JNDI

pavolloffay:
Bind metrics service to JNDI so other hawkular components can use inject it.

@jsanda @burmanm could you please review?

### DIFF
--- a/hawkular-component/src/main/java/org/hawkular/metrics/component/MetricsJNDIPusblisher.java
+++ b/hawkular-component/src/main/java/org/hawkular/metrics/component/MetricsJNDIPusblisher.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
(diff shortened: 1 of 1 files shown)


rhqci:

jsanda:
Couldn't we achieve the same functionality with a CDI producer which would not require EJB?

pavolloffay:
When I tried to bind it in ``MetricsServiceLifecycle``  I got ``WFLYNAM0043: Naming context is read-only``


This is the same approach as for Inventory-Pinger.

jsanda:
>This is the same approach as for Inventory-Pinger.

Be that as it may, I would prefer to avoid introducing the additional dependencies unless we really have to. More often than not we have found that CDI alone with EJB provided the functionality we need.

pavolloffay:
@jsanda now without EJB

pavolloffay:
injection works but in data mining Im getting this http://pastebin.com/5NgD0UHP

pavolloffay:
@jsanda can it go to the master?

PR #504, opened 2016-05-13
HWKMETRICS-305 - Split MetricsServiceITest

stefannegrea:


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -216,7 +216,8 @@ protected void initPreparedStatements() {
         readMetricsIndex = session.prepare(
             "SELECT metric, tags, data_retention " +
             "FROM metrics_idx " +
-            "WHERE tenant_id = ? AND type = ?");
+            "WHERE tenant_id = ? AND type = ? " +
+            "ORDER BY metric ASC");
```
jsanda: The ORDER BY clause is redundant. Here is the schema:

```
CREATE TABLE hawkulartest.metrics_idx (
    tenant_id text,
    type tinyint,
    metric text,
    data_retention int,
    tags map<text, text>,
    PRIMARY KEY ((tenant_id, type), metric)
) WITH CLUSTERING ORDER BY (metric ASC)
```

jsanda:
This is a good, overdue change!

PR #505, opened 2016-05-16
HWKMETRICS-399 Make Metrics independent of Hawkular Accounts

stefannegrea:


stefannegrea:
This PR should be ready for merge after the src-dep introduced in the first commit is removed.

ppalaga:
@stefannegrea commons 0.7.1.Final released.

jsanda:
I think the hawkular-server-tests module needs to be updated as well. AuthenticationITest.groovy can probably be deleted and we need to remove the dependency on the accounts feature pack.

stefannegrea:
@jsanda, I removed the tests you mention, however we will probably need to rework that entire module later since it is not running at the moment as part of the build.

stefannegrea:
retest this please

PR #514, opened 2016-06-07
HWKMETRICS-412 Add CORS filters back to hawkular component

tsegismont:
When Metrics was integrated with Accounts, the CORS filters were removed
because CORS had to be managed by Keycloak.

Now that Accounts is removed, we must add them back, otherwise we can't
send CORS requests to Metrics.

### DIFF
--- a/hawkular-component/pom.xml
+++ b/hawkular-component/pom.xml
@@ -135,11 +135,6 @@
             <overlay>
               <groupId>${project.groupId}</groupId>
               <artifactId>hawkular-metrics-api-jaxrs</artifactId>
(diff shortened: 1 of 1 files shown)


tsegismont:
Added filters back while waiting to see if there is interest in sharing them (see http://lists.jboss.org/pipermail/hawkular-dev/2016-June/002743.html). Added them back now allows to configure CORS in master builds already (and test the Grafana datasource plugin with a Hawkular server in direct mode).

stefannegrea:
Thanks @tsegismont!

PR #516, opened 2016-06-15
Gets around the 'unknown error' message in the log

mwringe:
This gets around an issue where users can see an 'unknown error' message in the logs when the Cassandra hostname could not be resolved.

Updates the log message to make it a bit more user friendly


stefannegrea:
Thanks @mwringe!

PR #518, opened 2016-06-17
HWKMETRICS-393

jsanda:
This PR is *not* ready to merge. I still have refactoring to do to reduce duplicate code, add javadocs, and clean up error handling. I am ready though to start getting some feedback. I have taken a different approach with these new endpoints in that I am streaming the response rather than realizing it entirely in memory before sending to the client.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.metrics.api.jaxrs;
+
+import java.util.List;
+
+/**
+ * @author jsanda
+ */
+public class QueryRequest {
+
+    private List<String> ids;
+
+    private Long start;
+
+    private Long end;
+
+    private Integer limit;
+
+    private String order;
+
+    public List<String> getIds() {
+        return ids;
+    }
+
+    public void setIds(List<String> ids) {
+        this.ids = ids;
+    }
+
+    public Long getStart() {
+        return start;
+    }
+
+    public void setStart(Long start) {
+        this.start = start;
+    }
+
+    public Long getEnd() {
+        return end;
+    }
+
+    public void setEnd(Long end) {
+        this.end = end;
+    }
+
+    public Integer getLimit() {
+        return limit;
+    }
+
+    public void setLimit(Integer limit) {
+        this.limit = limit;
+    }
+
+    public String getOrder() {
+        return order;
+    }
+
+    public void setOrder(String order) {
+        this.order = order;
+    }
+
+    @Override public String toString() {
+        return "QueryRequest{" +
+                "ids=" + ids +
+                ", start=" + start +
+                ", end=" + end +
+                ", limit=" + limit +
+                ", order=" + order +
+                '}';
+    }
+}
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
@@ -90,16 +90,7 @@ import rx.schedulers.Schedulers;
 @Produces(APPLICATION_JSON)
 @Api(tags = "String", description = "This resource is experimental and changes may be made to it in subsequent " +
         "releases that are not backwards compatible.")
-public class StringHandler {
-
-    @Inject
-    private MetricsService metricsService;
-
-    @Inject
-    private ObjectMapper mapper;
-
-    @HeaderParam(TENANT_HEADER_NAME)
-    private String tenantId;
+public class StringHandler extends MetricsServiceHandler {
 
     @POST
     @Path("/")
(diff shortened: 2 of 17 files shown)


burmanm:
Could we mark this QueryRequest API as experimental and not final? Given that we'll need at some point a proper QueryLanguage/Request (which this isn't, given that it is very limited) as our feature set grows. Otherwise we'll have yet another API to maintain soon..

jsanda:
>Could we mark this QueryRequest API as experimental and not final?

Sure but keep in mind that the API/parameters are the same as those for our other query endpoints. The only difference here is that we are using POST instead of GET.

jsanda:
`NamedDataPointObserver` is the class that generates the JSON that is streamed in the response. It sends an array of metric objects where each object contains an array of data points. I have some concerns around error handling. If there is an error, I think the StreamingOutput needs to throw a WebApplicationException with an appropriate status code and message. I am not sure what this means though for the response body since we've already started sending it to the client. For example, suppose we stream back data points from 3 out of 5 metrics. Then suppose the query for the fourth metric fails which results in `NamedDataPointObserver.onError` being called. As the code is currently written, I think we could end up with malformed JSON. But does it matter? In this particular scenario, we can and probably should gracefully recover from this. But what if we can't? How should the response be handled?

stefannegrea:
I would like to keep the request method with all the parameters inline because it is serves as documentation for the project. When I am asked what a method does and what params it takes, I often find myself linking directly to the code with the end-point method definition.

jsanda:
> I would like to keep the request method with all the parameters inline because it is serves as documentation for the project

I do not think this is possible since we are POST and supplying the request parameters in the request body.

tsegismont:
Have you tried to see what would happen by writing a fake endpoint?

Errors while sending a HTTP response could always happen. Even if you're 
simply sending a CSS or JS file.

Yet usual error cases (e.g. exception thrown before anything is sent to 
the wire) should lead to standard error codes.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
```diff
@@ -0,0 +1,85 @@
...
+ * limitations under the License.
+ */
+package org.hawkular.metrics.api.jaxrs;
+
+import java.util.List;
+
+/**
+ * @author jsanda
+ */
+public class QueryRequest {
```
tsegismont: I'm not entirely sure but I believe Swagger can analyze doc in body params.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/ObjectMapperProducer.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+/**
+ * @author jsanda
+ */
+@ApplicationScoped
```
tsegismont: Scope definition should be on the producer methods. Otherwise a new object mapper is created whenever it is needed for injection.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
```diff
@@ -263,6 +277,48 @@ public void addStringData(
...
+        Observable<NamedDataPoint<String>> dataPoints = metricsService.findDataPoints(metricIds, timeRange.getStart(),
+                timeRange.getEnd(), limit, order).observeOn(Schedulers.io());
+
+        StreamingOutput stream = output -> {
+            JsonGenerator generator = mapper.getFactory().createGenerator(output, JsonEncoding.UTF8);
+            CountDownLatch latch = new CountDownLatch(1);
+            dataPoints.subscribe(new NamedDataPointObserver<>(generator, latch, STRING));
+
+            try {
+                latch.await();
```
tsegismont: This is not right. We can't make task threads block until the whole response is loaded. This not only goes against our reactive design principle, but defeats the purpose of streaming the reponse.

stefannegrea:
> I do not think this is possible since we are POST and supplying the request parameters in the request body.

Sorry for the confusion. Do you mind then adding swagger documentation to the method parameter?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
stefannegrea: The usage of the latch is confusing. @jsanda do you mind clarifying why it is needed?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/NamedDataPointObserver.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+            currentMetric = dataPoint.getName();
+        } catch (IOException e) {
+                log.warn("Streaming data to client failed", e);
+        }
+    }
+
+    @Override
+    public void onError(Throwable e) {
+        try {
```
jsanda: Yeah, I need to update that.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
jsanda: > This is not right. We can't make task threads block until the whole response is loaded. This not only goes against our reactive design principle, but defeats the purpose of streaming the reponse.

We do not block until the whole response is loaded. The response starts getting streamed as soon as we get data back from Cassandra. We do not realize the whole response in memory. All of our other endpoints which do *not* stream the response do realize the entire response in memory before sending it back to the client.

Without the CountDownLatch the Response object gets returned before we finish streaming which results in an NPE when we call `generator.close()` here:

```java
public void onCompleted() {
        try {
            generator.writeEndArray();
            generator.writeEndObject();
            generator.writeEndArray();
            generator.close();
            latch.countDown();
        } catch (IOException e) {
                log.warn("Error while finishing streaming data", e);
        }
    }
```

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/ObjectMapperProducer.java
jsanda: I wasn't aware of that. What are the semantics then of having the class level annotation like I do?

tsegismont: There will be just one instance of the class.

Le 20/06/2016 à 19:56, jsanda a écrit :
> I wasn't aware of that. What are the semantics then of having the class
> level annotation like I do?
>

-- 
Thomas Segismont
JBoss ON Engineering Team

tsegismont: As discussed on IRC, I had not paid attention well enough, if the class is annotated with application scope and the mapper created in post construct, then getMapper will be called repeatedly but will always return the same instance.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
tsegismont: As discussed on IRC, it is true that the response will be streamed as data points are loaded. But the Undertow task thread will be blocked until the response is fully sent. Doing this allows to use the streaming features, but it goes against the efforts we've put in building a reactive architecture for Metrics.

jsanda: I disagree that it goes against our reactive architecture because we do not have a fully async/non-blocking architecture. Undertow may be async, but JAX-RS certainly is not. In every endpoint in which we use AsyncResponse we block that thread until the response is fully realized in memory. It simply is not fully reactive. With these new endpoints, we can be dealing with arbitrarily large result sets that could easily introduce lots of heap pressure, making the server not so scalable. I think that avoiding that memory pressure is more important than deferring the work to another thread which will still block until the response is fully realized.

I tried to make the streaming work with AsyncResponse but was unsuccessful. If there is a way to make it work, then I am all for it. If we were running on a fully async stack, then what I have done with the CountDownLatch would be a big red flag.

tsegismont: Le 20/06/2016 à 23:35, jsanda a écrit :
> In
> api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/StringHandler.java
> <https://github.com/hawkular/hawkular-metrics/pull/518#discussion_r67771082>:
>
>> +        } else {
>> +            order = Order.fromText(query.getOrder());
>> +        }
>> +
>> +        List<MetricId<String>> metricIds = query.getIds().stream().map(id -> new MetricId<>(tenantId, STRING, id))
>> +                .collect(toList());
>> +        Observable<NamedDataPoint<String>> dataPoints = metricsService.findDataPoints(metricIds, timeRange.getStart(),
>> +                timeRange.getEnd(), limit, order).observeOn(Schedulers.io());
>> +
>> +        StreamingOutput stream = output -> {
>> +            JsonGenerator generator = mapper.getFactory().createGenerator(output, JsonEncoding.UTF8);
>> +            CountDownLatch latch = new CountDownLatch(1);
>> +            dataPoints.subscribe(new NamedDataPointObserver<>(generator, latch, STRING));
>> +
>> +            try {
>> +                latch.await();
>
> I disagree that it goes against our reactive architecture because we do
> not have a fully async/non-blocking architecture. Undertow may be async,
> but JAX-RS certainly is not. In every endpoint in which we use
> AsyncResponse we block that thread until the response is fully realized

This is not true. The purpose of AsyncReponse is precisely to decouple 
response creation from endpoint invocation. More precisely, in our case, 
it is a thread from the computation scheduler which resumes the 
AsyncResponse when data is loaded.

> in memory. It simply is not fully reactive. With these new endpoints, we
> can be dealing with arbitrarily large result sets that could easily
> introduce lots of heap pressure, making the server not so scalable. I
> think that avoiding that memory pressure is more important than
> deferring the work to another thread which will still block until the
> response is fully realized.
>

I agree, we need to find to a solution for large responses. But 
regardless of the motivation, blocking a container task thread from the 
moment the request is received until the response is fully sent is one 
of the things we have tried to avoid.

> I tried to make the streaming work with AsyncResponse but was
> unsuccessful. If there is a way to make it work, then I am all for it.
> If we were running on a fully async stack, then what I have done with
> the CountDownLatch would be a big red flag.
>

I don't have the solution at hand. If going back to blocking handler 
implementation is the price to pay to avoid OOMEs with the current 
software stack, then fine.

But since it's a major departure from the current architecture goals, I 
believe it's worth the notice.

jsanda: It is certainly worth discussion and probably in a broader context outside of this PR review. I will post something on the mailing list. I do not see it s a major departure from our architecture goals. Performance and scalability are two of primary goals. Implementing an async, reactive design is an attempt to achieve those goals.

jsanda:
retest please

jsanda:
retest this please

FilipB:
@jsanda this PR introduces performance drop 5% but only for gauges metrics. Does it make sense?
Several test runs show the same result and other PR are green so the results should be valid and NOT caused by env.

jsanda:
It makes no sense as the PR only adds new code for some new query endpoints. There are no changes involved with inserting data.

jsanda:
retest this please

stefannegrea:
Thanks @jsanda!

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/QueryRequest.java
pilhuhn: At least it could in RHQ :-)

PR #519, opened 2016-06-17
HWKMETRICS-417 Improve performance of write path

burmanm:
Use mergeWith instead of concat and map the result to null as we don't use it at the moment. On my machine this improved the benchmark writes by about 10% (not including JAX-RS layer).

Also replace the test Observable.create with Observable.from(list) as it reduces the impact of test. Added support for JMC (Java Mission Control) profiler. Also the test uses now Netty native epoll on linux-x86_64, this further improves the performance by another 5-10% by reducing the CPU load of Cassandra driver (we should probably consider using this elsewhere also).


jsanda:
Will you open a separate ticket about making use of epoll? Now is the time to do it because productization work is going to start very soon.

PR #522, opened 2016-06-22
HWKMETRICS-420 Add ADMIN AvailabilityType

jshaughn:
- Add ADMIN as avail value=3
- Needed to refactor AvailabilityBucketPoint
  - this was flawed as it assumed only UP and DOWN avail, ignoring UNKNOWN
    and now ADMIN.
  - this is publicly exposed via JSON in the GET avail REST endpoints
    - Kept as deprecated some obsolete fields for JSON back compatibility
    - ??? One question for reviewers: Is it OK to now expose a Map in the
      JSON or do you prefer replacing the map with explicit fields ???

@stefannegrea and/or @jsanda please review.

### DIFF
--- a/bus/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
+++ b/bus/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
@@ -19,6 +19,7 @@ package org.hawkular.bus;
 import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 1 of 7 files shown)


rhqci:

FilipB:
ok to test

jsanda:
AvailabilityBucketPoint only assumed up/down because that is all that was required. I thinking deprecating the fields is good. We can mention it in the release notes. Lastly, I think using a map is fine. If other availability types are introduced, we don't have to keep adding fields which is nice.

PR #523, opened 2016-06-24
[HWKMETRICS-422] Add metrics index caches to avoid updating the metri…

stefannegrea:
…cs index if it was recently update. The max cache size in 20k items with automatica eviction of 1 hour since last read.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -41,6 +41,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
@@ -80,8 +81,11 @@ import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.ResultSetFuture;
 import com.datastax.driver.core.Row;
 import com.datastax.driver.core.Session;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
@@ -178,6 +182,8 @@ public class MetricsServiceImpl implements MetricsService {
      */
     private Map<MetricType<?>, Func1<Row, ? extends DataPoint<?>>> dataPointMappers;
 
+    private Map<MetricType<?>, Cache<String, Boolean>> metricsIdxCaches;
+
     private int defaultTTL = Duration.standardDays(7).toStandardSeconds().getSeconds();
 
     private int maxStringSize;
@@ -257,8 +263,19 @@ public class MetricsServiceImpl implements MetricsService {
                 .put(STRING, Functions::getStringDataPoint)
                 .build();
 
+        Builder<MetricType<?>, Cache<String, Boolean>> metricsIdxCachesBuilder = ImmutableMap
+                .<MetricType<?>, Cache<String, Boolean>> builder();
+        MetricType.userTypes().stream().forEach(
+                t -> metricsIdxCachesBuilder.put(t,
+                        CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).maximumSize(20000).build()));
+        metricsIdxCaches = metricsIdxCachesBuilder.build();
+
         initStringSize(session);
         initMetrics();
+    }
+
+    public void clearMetricsIdxCaches() {
+        metricsIdxCaches.values().stream().forEach(m -> m.invalidateAll());
     }
 
     void loadDataRetentions() {
@@ -628,23 +645,13 @@ public class MetricsServiceImpl implements MetricsService {
     public <T> Observable<Void> addDataPoints(MetricType<T> metricType, Observable<Metric<T>> metrics) {
         checkArgument(metricType != null, "metricType is null");
 
-        // We write to both the data and the metrics_idx tables. Each metric can have one or more data points. We
-        // currently write a separate batch statement for each metric.
+        // We write to the data table on every call and to the metrics_idx table only if the metric id is not
+        // in cache. Each metric can have one or more data points. We currently write a separate batch statement
+        // for each metric.
         //
         // TODO Is there additional overhead of using batch statement when there is only a single insert?
         //      If there is overhead, then we should avoid using batch statements when the metric has only a single
         //      data point which could be quite often.
-        //
-        // The metrics_idx table stores the metric id along with any tags, and data retention. The update we perform
-        // here though only inserts the metric id (i.e., name and interval). We need to revisit this logic. The original
-        // intent for updating metrics_idx here is that even if the client does not explicitly create the metric, we
-        // still have it in metrics_idx. In reality, I think clients will be explicitly creating metrics. This will
-        // certainly be the case with the full, integrated hawkular server.
-        //
-        // TODO Determine how much overhead is caused by updating metrics_idx on every write
-        //      If there much overhead, then we might not want to update the index every time we insert data. Maybe
-        //      we periodically update it in the background, so we will still be aware of metrics that have not been
-        //      explicitly created, just not necessarily right away.
 
         Meter meter = getInsertMeter(metricType);
         Func2<Metric<T>, Integer, Observable<Integer>> inserter = getInserter(metricType);
@@ -655,7 +662,18 @@ public class MetricsServiceImpl implements MetricsService {
                         .doOnNext(i -> insertedDataPointEvents.onNext(metric)))
                 .doOnNext(meter::mark);
 
-        Observable<Integer> indexUpdates = dataAccess.updateMetricsIndex(metrics)
+        Cache<String, Boolean> metricsIdxCache = metricsIdxCaches.get(metricType);
+
+        Observable<Metric<T>> updateIdxMetrics = metrics.filter(metric -> {
+            if (metricsIdxCache.getIfPresent(metric.getId()) == null) {
+                metricsIdxCache.put(metric.getId(), true);
+                return true;
+            }
+
+            return false;
+        });
+
+        Observable<Integer> indexUpdates = dataAccess.updateMetricsIndex(updateIdxMetrics)
                 .doOnNext(batchSize -> log.tracef("Inserted %d %s metrics into metrics_idx", batchSize, metricType));
 
         return updates.mergeWith(indexUpdates)
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -262,7 +262,7 @@ public class MetricsServiceImpl implements MetricsService {
                 .put(STRING, Functions::getStringDataPoint)
                 .build();
 
-        metricsIdxCache = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).maximumSize(40000).build();
+        metricsIdxCache = CacheBuilder.newBuilder().expireAfterAccess(60, TimeUnit.MINUTES).maximumSize(40000).build();
 
         initStringSize(session);
         initMetrics();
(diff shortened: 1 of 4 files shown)


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -257,10 +262,16 @@ public void startUp(Session session, String keyspace, boolean resetDb, boolean c
                 .put(STRING, Functions::getStringDataPoint)
                 .build();
 
+        metricsIdxCache = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.HOURS).maximumSize(40000).build();
```
jsanda: The expiration and cache size should be configurable.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -655,7 +656,16 @@ public void setDefaultTTL(int defaultTTL) {
                         .doOnNext(i -> insertedDataPointEvents.onNext(metric)))
                 .doOnNext(meter::mark);
 
-        Observable<Integer> indexUpdates = dataAccess.updateMetricsIndex(metrics)
+        Observable<Metric<T>> updateIdxMetrics = metrics.filter(metric -> {
```
jsanda: All of the index related work can/should be done out of band. We want to treat as an async, background task. This means a couple key changes. I would change the return statement from 

```return updates.mergeWith(indexUpdates)```

to just

```return updates```

And then we will need to subscribe to ```indexUpdates`` (or whatever index related observables we're dealing with) separately so that the work does get performed.

Comment (stefannegrea) [inline core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java]: Not sure if this is a good idea because this will make at least the outcome of some tests non-deterministic, since the metric will not be retrievable. Also, we reduce the metrics idx updates by a lot with this caching strategy, so I would like to keep the guarantee that a metric is retrievable immediately after insertion of data point.

File: core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java (hunk 1 above)
Comment (stefannegrea) [inline core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java]: Agree, I will make them configurable properties and switch the expiration unit to minutes.

File: core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java [hunk 3]
```diff
@@ -655,7 +656,16 @@ public void setDefaultTTL(int defaultTTL) {
...
+            if (metricsIdxCache.getIfPresent(metric.getMetricId()) == null) {
+                metricsIdxCache.put(metric.getMetricId(), true);
```
burmanm: Filter should not do any changes, not even insert to cache. We should not expect in reactive way to expect side effects from filter methods.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: We should not be using Guava caches here, but instead something like Infinispan which provides us the ability of using distributed caches.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jsanda: We can still make tests deterministic by putting some hooks in place. I have do things like this in my job scheduler branch in order to try and make things deterministic. You could do something like this:

```java
private Optional<PublishSubject<Void>> indexUpdatesFinished = Optional.empty();
...
// package access for testing
void setIndexUpdatesFinished(PublishSubject<Void> subject) {
    indexUpdatesFinished = Optional.of(subject);
}

// In the method that has the indexUpdates observable
indexUpdates.subscribe(
    aVoid ->  // nothing really to do here
    t ->   // Need to log the error and for tests notify that the update failed
    () -> indexUpdatesFinished.ifPresent(subject -> subject.onNext(void))
);
```
Your tests pass in the subject and subscribe to it in order to get notification of when the updates finish. It is a little more work, but that's part in parcel when dealing with async code :) Doing the index updates out of band was the main thing I wanted to see happen with this ticket, so I'd like us to try and make whatever testing changes we can to support that.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jsanda: I disagree about using a distributed cache. The purpose of this ticket is to speed requests for inserting data points. We can speed up requests by simply taking the index updates off the write path and doing them out of band in a background task. The cache is an additional optimization in that it reduces the amount and frequency of writes. I am very skeptical that the overhead of a distributed cache would be worth it.

FilipB:
This PR increases write throughput ~8% Does it make sense? If yes please contact me before merging so I can update baselines.

jsanda:
@FilipB This is the expectation. We may even see a bigger increase before merging.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: Distributed does not mean partitioned, it could be asynchronously replicated cache just as well. In that case the performance should be the same when it comes to local cache requests. The point is to reduce unnecessary writes to allow Cassandra to process more inserts when it comes to the datapoints.

Also, the cache should be populated in the start by requesting the metricIds from the Cassandra.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: There are side-effects on the cache just by reading from it. I am not really worried here about the implications on the cache itself. Also the cache is thread safe so regardless of the threading model here the cache operations will work.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: @burmanm, it would have made sense to use Infinispan if it was used somewhere else in the code. But to introduce the Infinispan dependency just for this is a bit too much. I will open a ticket for the future to reconsider this decision once we add Infinispan dependency to the project for other reasons.

stefannegrea: The idea of this cache was to just avoid writing to the metrics index when there is a burst of activity for a single metric. That is why elements in the cache are evicted if there is no activity for a metric for a period of time.

Loading the cache at startup is not a good idea because of the memory implications. There can be a lot of metrics that were written once and never again to be written. I prefer this cache to be based on the current metrics activity.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: @jsanda, I only partially understand your last post with regards to what needs to be done. I would like to keep this PR simple and then enhance. I will clone the current ticket and emphasize to add the improvement to make it out-of-band. And we can tackle that in the next release.

To have the index updated out-of-band we will have to re-write the entire testing infrastructure, not just for the core services but also for REST. And I am more concerned about the impact on the REST integration tests to be honest.

stefannegrea: Just to clarify the last comment, the impact on the tests shows that fact that the REST API becomes non-deterministic. Meaning, when a user submits a new metric data point without creating the metric prior the metric is not available for query for a non-deterministic amount of time because there are no guarantees on the state of the metric index write.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: We need to use Infinispan with the TTL changes also (and that was in a PR already as well). So it will be added sooner or later in any case, as we need to store the TTL value distributed in the cluster for deletion job. Might as well add it here so we don't have to rewrite this instantly.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: It does not make sense for anyone else reading the code that suddenly removing items actually writes items and changes the behavior elsewhere. This just makes the code difficult to follow. And there's no advantage to doing it here.

It's also not a good idea for backpressure aware processing, as filter should not be the one causing backpressure (and IIRC, filters do not support backpressure).

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: The problem with enhancing this with further commits is that the first thing to do is revert this commit (this does not provide any starting point for the functionality @jsanda is thinking about).

burmanm:
Guava's cache statistics should also be available for users (through the metrics library which we otherwise use in the insert pipeline for example). This allows debugging and monitoring of the internal cache.

That is, if we stick to Guava.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: I think the TTL changes will materialize in another form, but @jsanda has more details. I will hold off doing anything for this PR until he returns.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: Then I completely do not understand what @jsanda said in his comment. This is turning really complicated really fast. My goal was to reduce the number of metrics idx inserts when possible. And the implementation so far in this PR is just that; have a cache that keeps track of recent metric index inserts and avoid inserting to the metrics index if there is a cache hit.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
stefannegrea: Where do you propose to push this functionality? The data access layer is clearly not good because that is just queries; and the REST layer is just the glue code for the container. So only the core services is left. 

Also, how else can this be done other than filtering out metrics that do not need metrics index updates before passing them to the data access layer?

stefannegrea:
Who should have access to such statistics? How should they be exposed?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: The most logical place for it is in the indexUpdates Observable. It's part of the process, if we insert the index to the Cassandra then we update the cache. Should something happen (like Cassandra write fail), we would not write that information to the cache either. Updating cache is part of that index-writing-process.

burmanm:
I guess @tsegismont could answer best the Codahale & MetricRegistry questions? I assume they were meant for both internal monitoring as well as potentially some debugging. Exposing the cache metrics would follow the same line.

tsegismont:
Yes, internal monitoring. It's a pity to be a metric storage engine and 
not expose a few metrics :)

There are metrics added to a Dropwizard registry already. And the 
registry is already configured for JMX reporting. Pretty useful when you 
are (load-)testing the server.

Le 29/06/2016 à 19:56, Michael Burman a écrit :
> I guess @tsegismont <https://github.com/tsegismont> could answer best
> the Codahale & MetricRegistry questions? I assume they were meant for
> both internal monitoring as well as potentially some debugging. Exposing
> the cache metrics would follow the same line.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/hawkular/hawkular-metrics/pull/523#issuecomment-229436454>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABbltsXZgIQbuQcjUYLnzWue5ZMALUDvks5qQrHjgaJpZM4I-Ce1>.
>

-- 
Thomas Segismont
JBoss ON Engineering Team

jsanda:
I just want to reiterate what I expressed on the mailing list discussion. Provided that we do not have use cases in which we need to support querying metrics_idx in order to load all meta data for a metric, then I am in favor of dropping the table. And even if we do have that use case, we would still change when we write to metrics_idx to only be when we set/update data retention and set/update tags.

stefannegrea:
I will close this PR since the direction to be taken is completely different than the code in the branch. I will open a new PR with a new set of changes.

PR #524, opened 2016-06-25
[HWKMETRICS-425] fix schema upgrade bugs

jsanda:
There were a couple bugs in cassalog that broke upgrades. I have fix and
released a new version of cassalog. This commit bumps the cassalog version and
also fixes a couple issues in bootstrap.groovy. Lastly,I have made resetting
schema during integration tests in metrics-core-service optional with the
resetdb system property which expects a value of true or false.


stefannegrea:
Thanks @jsanda!

PR #525, opened 2016-06-27
HWKMETRICS-427 JsonGenerationException: Current context not an object but ROOT

tsegismont:
Fixed no data for query case.

A couple of side changes:
- the observer might run on different threads so make fields final when possible and volatile otherwise
- simplified writeValue lambda form by using a dedicated functional interface


stefannegrea:
Thanks @tsegismont!

PR #526, opened 2016-06-27
[HWKMETRICS-426] Change Percentiles to return original percentile

burmanm:
Change percentile queries to return the original requested value instead of calculated value from PSPercentile (from apache.commons.math). This prevents client from having to deal with floating point imprecision in quantile keys.


stefannegrea:
Thanks @burmanm!

PR #527, opened 2016-06-28
[HWKMETRICS-417] Revert change to check performance impact for curren…

stefannegrea:
…t perf CI job.


stefannegrea:
Please do not merge this!

FilipB:
This PR can be closed. It does not improve performance as expected. It seems the original problem introduced in PR#519 was only visible before another change introduced in PR#520

PR #531, opened 2016-07-07
HWKMETRICS-432 Status page cannot be displayed when status is not "STARTED"

tsegismont:
Do not abort status page request if status is not "STARTED"


stefannegrea:
Thanks @tsegismont!

PR #535, opened 2016-07-10
[HWKMETRICS-424] Fetch stats from multiple metrics in a single request

jsanda:
This PR adds an endpoint for fetching stats (i.e., bucket data points) from multiple metrics in a single request. Right now it only supports gauges and counter. I will update the endpoint to also support gauge and counter rate stats as well. I wanted to go ahead submit the PR for review though because adding support for rates is mostly boiler plate along with additional integration tests.

@lucasponce would it make sense to include availability stats as well?

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
@@ -16,6 +16,7 @@
  */
 package org.hawkular.metrics.api.jaxrs;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -25,7 +26,7 @@ import java.util.Objects;
  */
 public class StatsQueryRequest {
 
-    private Map<String, List<String>> metrics;
+    private Map<String, List<String>> metrics = new HashMap<>();
 
     private Long start;
 
@@ -36,6 +37,8 @@ public class StatsQueryRequest {
     private String bucketDuration;
 
     private String percentiles;
+
+    private String tags;
 
     public Map<String, List<String>> getMetrics() {
         return metrics;
@@ -85,12 +88,21 @@ public class StatsQueryRequest {
         this.percentiles = percentiles;
     }
 
+    public String getTags() {
+        return tags;
+    }
+
+    public void setTags(String tags) {
+        this.tags = tags;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         StatsQueryRequest that = (StatsQueryRequest) o;
         return Objects.equals(metrics, that.metrics) &&
+                Objects.equals(tags, that.tags) &&
                 Objects.equals(start, that.start) &&
                 Objects.equals(end, that.end) &&
                 Objects.equals(buckets, that.buckets) &&
@@ -106,6 +118,7 @@ public class StatsQueryRequest {
     @Override public String toString() {
         return "StatsQueryRequest{" +
                 "metrics=" + metrics +
+                "tags=" +
                 ", start=" + start +
                 ", end=" + end +
                 ", buckets=" + buckets +
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
@@ -16,10 +16,13 @@
  */
 package org.hawkular.metrics.api.jaxrs;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
 
 /**
  * @author jsanda
@@ -39,6 +42,8 @@ public class StatsQueryRequest {
     private String percentiles;
 
     private String tags;
+
+    private List<String> types = new ArrayList<>();
 
     public Map<String, List<String>> getMetrics() {
         return metrics;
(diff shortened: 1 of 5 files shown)


lucasponce:
Hi @jsanda, for consistency it would make sense to include it, but today in MiQ availability is fetched individually. In the graphs we are not showing it yet, it is only showed when the inventory is refreshed.
So, it could be a nice to have but it can be added in future effort IMO.

pilhuhn:
On 11 Jul 2016, at 7:52, Lucas Ponce wrote:

> Hi @jsanda, for consistency it would make sense to include it, but 
> today in MiQ availability is fetched individually. In the graphs we 
> are not

There is actually an effort going on to fetch multiple at once
via some other HWK-metric jira, that got added in the latest release.

-- 

jsanda:
@pilhuhn are you referring to [HWKMETRICS-393](https://issues.jboss.org/browse/HWKMETRICS-393)? That added an endpoint per metric type that allow you to fetch raw data points from multiple metrics in the same request. The metrics have to be of the same type though. This PR is different in two ways. First, it returns stats, i.e., bucket data points. Secondly, it will return data points across different metric types.

pilhuhn:
On 11 Jul 2016, at 15:39, jsanda wrote:

> @pilhuhn are you referring to 
> [HWKMETRICS-393](https://issues.jboss.org/browse/HWKMETRICS-393)? That 
> added an endpoint per metric type that allow you to fetch raw data 
> points from multiple metrics in the same request. The metrics have to 
> be of the same type though. This PR is different in two ways. First, 
> it returns stats, i.e., bucket data points. Secondly, it will return 
> data points across different metric types.

Yes. I was referring to the "no one fetches multiple availabilites in on 
go" comment, which is no
longer true with -393

stefannegrea:
@jsanda, I recently merged PR #526  from @burmanm that changed the percentile implementation. Do you mind updating your code to match the change?

jsanda:
> Do you mind updating your code to match the change?

Done. I just rebased.

jsanda:
I've added support for availability in addition to gauges and counters. For completeness I think we ought to also support returning rate bucket data points in the response as well. I am not sure though how that look in the request. For example, we could do

```javascript
{
  "metrics": {
    "counter_rate": ["C1", "C2", "C3"]
  }
}
```

where we explicitly require the client to specify the ids of the metrics for which rate data is requested. If we are searching by tags though, the behavior thus far in the PR is to return data for metrics of all types that match the tag filters. I don't think we should automatically include rates as it involves extra work. The client should make it explicit that the response should include rates. After giving it some thought I think the best way to indicate that rates should be included is with a ```includeRates``` property in the request body. This property should be used whether searching directly by ids or by tags.

And if we want more fine grained control, such as including counter rates but not gauge rates, we could add support for additional properties to indicate which rates should be included, e.g., ```includeCounterRates``` and ```includeGaugeRates```.

stefannegrea:
The individual stats endpoints have `fromEarliest` parameter. I wonder if it relevant in this context too.

stefannegrea:
@jsanda, externally the rates are not metrics types. So we should not have them exposed alone in the response. How about using a parameter on the request to request rates along with regular stats? Or a flag to request either stats or rates for the gauge and counter metrics?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/StatsQueryRequest.java
```diff
@@ -0,0 +1,116 @@
...
+                Objects.equals(bucketDuration, that.bucketDuration) &&
+                Objects.equals(percentiles, that.percentiles);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(metrics, start, end, buckets, bucketDuration, percentiles);
+    }
+
+    @Override public String toString() {
```
stefannegrea: Do you mind using `MoreObjects.toStringHelper` for the toString method?

jsanda:
> The individual stats endpoints have fromEarliest parameter. I wonder if it relevant in this context too.

Good question. I actually only see the ```fromEarliest``` parameter on the endpoints for gauge and counter stats. I don't see it for availability or for the rate stats endpoints. Do you know if this is by design or an oversight?

jsanda:
For rates I am leaning towards a flag to indicate that rates are requests.

> externally the rates are not metrics types. So we should not have them exposed alone in the response. 

Why do you say we shouldn't expose them alone? More often than not users are probably more interested in seeing counter rate data vs raw counter data.

stefannegrea:
> For rates I am leaning towards a flag to indicate that rates are requests.

+1

> > externally the rates are not metrics types. So we should not have them exposed alone in the response. 

> Why do you say we shouldn't expose them alone? More often than not users are probably more interested in seeing counter rate data vs raw counter data. 

Sorry the confusing comment, I did not provide enough context. You gave this example in a previous comment:
```javascript
{
  "metrics": {
    "counter_rate": ["C1", "C2", "C3"]
  }
}
```

My comment was related to that, in the sense that I would prefer to have a flag for including rates rather than request rates separately via something like `counter_rate` since that is not a top metric (just a sub-metric type).

stefannegrea:
> Good question. I actually only see the fromEarliest parameter on the endpoints for gauge and counter stats. I don't see it for availability or for the rate stats endpoints. Do you know if this is by design or an oversight?

It's a bit of both. `fromEarliest` was requested specifically for stats on gauges and counters. The concept could be extended to raw data but since the code was originally under `/data`, I stopped at what was requested. Now that we have separate end-points for raw and stats we can go back and enhance raw with this option too.

So, I think it makes sense to add that option in this case too.

jsanda:
Here is the challenge I see with respect to rates. I see three use cases:

* Fetch data without any rates
* Fetch only rate data
* Fetch a mixture of non-rate and rate data

All three are valid use cases. There are times, and maybe more often than not, where the user wants counter rates but not the raw data. We should be able to return only rate data if that is what the user wants. We could have a flag like `onlyRates` to satisfy this scenario. Now consider the scenario in which I want a mix of gauge and counter rate data, and I don't need the raw counter data. I am trying to figure out a good way to express this in the request.

stefannegrea:
@jsanda, not sure if having rate types as an option is a good idea since rates are not first class metric types.

jsanda:
Sure rates are derived, but how/why is that problematic? How would you give the user the option to return only rates?

stefannegrea:
@jsanda, one more thing that I overlooked in my comment. The code allows storing two metrics with the same id but different types. The endpoint needs to return only the metrics the user asks for without ambiguity.

jsanda:
> The code allows storing two metrics with the same id but different types. The endpoint needs to return only the metrics the user asks for without ambiguity.

If we allow searching by tag filters then there is necessarily some ambiguity in terms of searching, particularly if the request does not also restrict types. It isn't a problem though because results are grouped by type, eliminating any ambiguity.

jsanda:
I have pushed commits that add support for a `types` property in the request property.  By default rates are not returned. If you want rates, you need to specify it using `types`.  If want the raw data stats as well, that needs to be made explicit, e.g.,

```javascript
"types": ['gauge', 'gauge_rate', 'counter', 'counter_rate'] 
```

When querying by tags, this endpoint will return raw data stats for all metric types by default, excluding tags though. The `types` property can be used in conjunction with tag filters to restrict the types that are queried.

Since this PR has already gotten pretty big and has taken longer than anticipated, I am going to defer adding support for `fromEarlierest` to a separate PR. See [HWKMETRICS-445](https://issues.jboss.org/browse/HWKMETRICS-445).

stefannegrea:
@jsanda, really like that we avoided having one extra flag for rates. It is a bit more intuitive and easy to remember how to just ask for them in the types.

Also, I agree splitting the work since this PR is getting big. As soon as the build passes, I will merge this PR.

stefannegrea:
Thanks @jsanda!

PR #536, opened 2016-07-11
[HWKMETRICS-440] Create Interface for JAXRS Handlers

stefannegrea:
Create a basic interface for JAXRS handlers for methods like creating, listing, and adding data to metrics.

I created [HWKMETRICS-435](https://issues.jboss.org/browse/HWKMETRICS-435) to track implementing 'fromEarliest' for `/raw` endpoints for gauge and counter. Somehow this was missed during the migration away from `/data`.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/template/IMetricsHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/template/IMetricsHandler.java
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.hawkular.metrics.api.jaxrs.handler;
+package org.hawkular.metrics.api.jaxrs.handler.template;
 
 import java.util.List;
 import java.util.Map;
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/template/IMetricsHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/template/IMetricsHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/template/IMetricsHandler.java
```diff
@@ -0,0 +1,65 @@
...
+import org.hawkular.metrics.model.DataPoint;
+import org.hawkular.metrics.model.Metric;
+import org.hawkular.metrics.model.param.TagNames;
+import org.hawkular.metrics.model.param.Tags;
+
+/**
+ * @author Stefan Negrea
+ *
+ */
+public interface IMetricsHandler<V> {
```
burmanm: We've used <T> to denote type elsewhere, any reason this is <V> (I guess short for Value) ? While it has no impact on functionality, it is confusing. Unless there's some hidden meaning I didn't catch.

stefannegrea: Good point. I will update the code to use T.

jsanda:
I am ok with the changes, but I am really starting to wonder whether or not we really need all of the different handlers and endpoints for each metric type. We have take our internal database schema design and imposed that on the REST API. Rather than have a separate implementation of IMetricsHandler per metric type, why not just have those methods/endpoints in MetricHandler where each one has a type parameter. I do want to reiterate that I have already been thinking about, but this PR makes for a good opportunity to discuss it. I haven't yet gone through all of the endpoints, but I think the majority of them could easily be combined by adding a type parameter.

jsanda:
I am neither suggesting nor advocating anything with respect to implementation. In fact that is entirely a secondary concern here. I am not sure that I agree about things being less clear. We can still have type-specific endpoints where it makes sense. Think of a new user writing client code for Hawkular Metrics. There are a lot of endpoints for inserting data. The behavior is (or should be) the same for all of those, but won't be clear to the user. The same goes for creating metrics as well as several other redundant endpoints. We already have endpoints in MetricHandler for creating metrics, fetching metric definitions, and for inserting data points. When [HWKMETRICS-423](https://issues.jboss.org/browse/HWKMETRICS-423) and [HWKMETRICS-424](https://issues.jboss.org/browse/HWKMETRICS-424) are done, MetricHandler will also have endpoints for fetching raw and stats data. Do we really want to have all of these redundant endpoints?

stefannegrea:
I just created [JIRA HWKMETRICS-440](https://issues.jboss.org/browse/HWKMETRICS-440) to track this change.

PR #542, opened 2016-07-13
Disable users from being able to write metrics by default.

mwringe:
Simple switch to allow us to enable or disable users from writing metrics into the system.

This is to allow the system to be locked down so that something only like Heapster will have access to be able to write metrics.

### DIFF
--- a/containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/TokenAuthenticator.java
+++ b/containers/hawkular-metrics-openshift-integration/src/main/java/org/hawkular/openshift/auth/TokenAuthenticator.java
@@ -36,6 +36,7 @@ import static io.undertow.util.StatusCodes.BAD_REQUEST;
 import static io.undertow.util.StatusCodes.CREATED;
 import static io.undertow.util.StatusCodes.FORBIDDEN;
 import static io.undertow.util.StatusCodes.INTERNAL_SERVER_ERROR;
+import static io.undertow.util.StatusCodes.UNAUTHORIZED;
 
 import java.io.Closeable;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


burmanm:
This would also disable Grafana interface as it uses POST-method to request data.

stefannegrea:
@mwringe, please create a corresponding Metrics JIRA too.

mwringe:
@stefannegrea I always create a new jira before submitting a PR. Look at the branch name :) its https://issues.jboss.org/browse/HWKMETRICS-444

@burmanm is there any easy way based on the URL to determine if a POST is for read or write purposes? Using POST for GET operations is usually problematic with REST interfaces :(

burmanm:
@mwringe, other than looking at the URL, no, not really. POST is awful fallback to avoid the Golang's buggy URL handling. Of course I guess you could avoid this scenario by forcing Grafana to use service account..

mwringe:
@burmanm right, but if we were to look at the URL, is just looking at the URL enough to determine if a POST to that URL is for a read or for a write?

And if so, what URLS would correspond to only reads and which ones would correspond to only writes?

mwringe:
hmm, none of the POST methods listed in the API docs mention that you can also use it to return values instead of creating something.

burmanm:
It's not in the documentation yet as it's in the master. @POST @Path("/raw/query") is the path, which the Grafana datasource also uses.

mwringe:
ok, if its just one path with post, the PR can be updated for that

stefannegrea:
@mwringe, sorry, I did not spot the JIRA in the branch name. 

Also, there is a second end-point that will be merged soon that follows the POST pattern. Please look at PR #535.

mwringe:
Ok, this has been updated to take into consideration that certain POST commands can really just be for  certain types of queries.

PR #544, opened 2016-07-14
[HWKMETRICS-360] job scheduler

jsanda:
The work is not yet complete, but tests are in a good state so I wanted to open this for review. This [design doc](https://goo.gl/3S5Dza) is probably a good reference to have while going through the code.


core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.16.0.groovy
```diff
@@ -0,0 +1,80 @@
...
+    job_id uuid,
+    PRIMARY KEY (time_slice, job_id)
+) WITH compaction = { 'class': 'LeveledCompactionStrategy' }
+"""
+}
+
+schemaChange {
+  version '2.3'
+  author 'jsanda'
+  tags '0.16.x'
```
stefannegrea: I think these tags need to be moved at least to 0.18.0 since that is the version we are going to release with this feature.

core/datetime-service/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.hawkular.metrics</groupId>
+    <artifactId>hawkular-metrics-parent</artifactId>
+    <version>0.15.0-SNAPSHOT</version>
```
stefannegrea: This version needs to be updated too.

core/schema/src/main/resources/org/hawkular/schema/cassalog.groovy
```diff
@@ -20,4 +20,5 @@ include '/org/hawkular/schema/bootstrap.groovy'
 
 setKeyspace keyspace
 
-include '/org/hawkular/schema/updates/schema-0.15.0.groovy'
\ No newline at end of file
+include '/org/hawkular/schema/updates/schema-0.15.0.groovy'
+include '/org/hawkular/schema/updates/schema-0.16.0.groovy'
```
stefannegrea: Is cassalog able to skip versions? In this case you would have a file for 0.15.0 and the next one is 0.18.0.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/api/RepeatingTrigger.java
```diff
@@ -0,0 +1,156 @@
...
+
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * @author jsanda
+ */
+public class RepeatingTrigger implements Trigger {
+
+    private Long triggerTime;
```
stefannegrea: Should this fields be final?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/api/RepeatingTrigger.java
```diff
@@ -0,0 +1,156 @@
...
+        public RepeatingTrigger build() {
+            return new RepeatingTrigger(interval, delay, repeatCount);
+        }
+
+    }
+
+    private RepeatingTrigger() {
+    }
+
+    private RepeatingTrigger(Long interval, Long delay, Integer repeatCount) {
```
stefannegrea: How would you schedule a job that needs to be executed until cancelled? Basically with infinite repeat?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -0,0 +1,639 @@
...
+        queueScheduler = Schedulers.from(queueExecutor);
+
+        queryExecutor = Executors.newFixedThreadPool(4,
+                new ThreadFactoryBuilder().setNameFormat("query-thread-pool-%d").build());
+        queryScheduler = Schedulers.from(queryExecutor);
+
+        lockManager = new LockManager(session);
+
+        insertJob = session.getSession().prepare(
+                "INSERT INTO jobs (id, type, name, params, trigger) VALUES (?, ?, ?, ?, ?)");
```
stefannegrea: Should these CQL statements into the core library where the rest of the data access code is? And keep this library high level with no direct C* interactions?

core/schema/src/main/resources/org/hawkular/schema/cassalog.groovy
jsanda: The versions and the file names are completely arbitrary as far as Cassalog is concerned. The `include` directive is telling Cassalog which scripts to execute. It just so happens that we have adopted a convention to have a separate script per release (when there are schema changes).

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/api/RepeatingTrigger.java
jsanda: You would use a RepeatingTrigger. Note the logic around `repeatCount` in the `nextTrigger` method:

```java
public Trigger nextTrigger() {
    // TODO what should we do if interval is null?

    if (repeatCount != null && executionCount + 1 > repeatCount) {
        return null;
    }
    RepeatingTrigger next = new RepeatingTrigger();
    next.interval = interval;
    next.delay = delay;
    next.triggerTime = triggerTime + interval;
    next.repeatCount = repeatCount;
    next.executionCount = executionCount + 1;

    return next;
}
```

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: No, these CQL statements should stay in the job-scheduler module. The job-scheduler module does not and will not depend on metrics-core-service. It may be the other way around where metrics-core-service depends on job-scheduler. Furthermore, there is nothing inherently specific to metrics data access in the job scheduler stuff, so pushing the CQL related code into metrics-core-service would make it somewhat convoluted IMO.

jsanda:
@burmanm tests have been passing consistently now thanks to your help. Would you review so we can get this merged? There are still changes that are needed, but considering how this work has dragged on, I would prefer to tackle additional changes in follow up tickets. One thing I did want to do is have an actual job implemented to see what the API will be and how to integrate things in the web app. I am implementing a tenant deletion job but doing it for [HWKMETRICS-446](https://issues.jboss.org/browse/HWKMETRICS-446). I'll post a link to that branch shortly.

jsanda:
Tenant deletion work is being done [here](https://github.com/jsanda/hawkular-metrics/tree/hwkmetrics-446).

PR #549, opened 2016-07-26
[HWKMETRICS-360] initial job scheduler work

jsanda:
There are still changes that are needed, but considering how this work has dragged on, I would prefer to tackle additional changes in follow up tickets. One thing I did want to do is have an actual job implemented to see what the API will be and how to integrate things in the web app. I am implementing a tenant deletion job but doing it for HWKMETRICS-446. Tenant deletion work is being done [here](https://github.com/jsanda/hawkular-metrics/tree/hwkmetrics-446).


FilipB:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        scheduler.start();
+    }
+
+    @Override
+    public void shutdown() {
+        scheduler.shutdown();
+    }
+
+    @Override
+    public Single<JobDetails> submitDeleteTenantJob(String tenantId) {
```
burmanm: Do we really have to add new jobs to the JobsServiceImpl?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -0,0 +1,606 @@
...
+
+        doOnTick(() -> {
+            logger.debug("Activating scheduler for [" + currentMinute().toDate() + "]");
+
+            Date timeSlice = currentMinute().toDate();
+            // TODO Figure out why updateActiveTimeSlices() with Completable doesn't work
+            // When using updateActiveTimeSlices() that returns a Completable, we get intermittent failures. Need
+            // understand why. Maybe it has something to do with converting from ListenableFuture to Observable to
+            // Completable.
+            updateActiveTimeSlices(currentMinute().toDate()).andThen(findTimeSlices())
```
burmanm: Won't this feature break if used on multiple nodes? As we already had previously issues with concurrency in a single computer, these changes will not help if ran on multiple nodes.

jsanda: We need to keep track of the time slices that all job scheduler instances collectively are working on. What specifically are you referring to?

burmanm: It will return the same information to each node, and they all start the same process (just like the issue with 2 ticks at the same time), as the filter later on has no effect on another nodes. Not that it necessarily matters as acquireTimeSliceLock should prevent two from moving on..

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -263,10 +259,10 @@ public void start() {
                             return Observable.just(time);
                         });
                     })
-//                    .doOnNext(activeTimeSlices::remove)
                     .subscribe(
                             d -> {
                                 logger.debug("Finished post job execution clean up for [" + d + "]");
+                                // TODO should this be in a synchronized block?
```
burmanm: Not unless you need some return values. Both threads can do remove and ifPresent they can do something else. If you needed to check that if something succeeds, then do something then this would be a different case. Why can't two threads execute both of these?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: That's correct. The same information will be returned to each node, and that is expected. The concurrency issue you debugged actually is specific to single node as it involves the ```activeTimeSlices``` cache which is a local variable that keeps track of the active time slices for the running node. We want to allow multiple job scheduler nodes to run concurrently, executing jobs over the same set of time slices. Locks are the coordination mechanism here.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: I think multiple thread execute the remove concurrently which is why I do't have it in a synchronized block. I wanted you to look at it as well though.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
jsanda: There are a few reasons I introduced JobsService. First, having a level of indirection can make writing tests easier. Secondly, if you look at PR 547, you will see that it takes care of initializing the job instances. The tenant deletion job gets created and scheduled upon client request. There will be other jobs, like computing rollups and data purge, which should be created upon server start up. JobsService can encapsulate this work.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: I just created [HWMETRICS-451](https://issues.jboss.org/browse/HWKMETRICS-451).

PR #550, opened 2016-07-27
[HWKMETRICS-446] endpoint for tenant deletion

jsanda:


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/DeleteTenant.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public DeleteTenant(RxSession session, MetricsService metricsService) {
+        this.session = session;
+        this.metricsService = metricsService;
+        deleteTenant = session.getSession().prepare("DELETE FROM tenants WHERE id = ?");
+        deleteData = session.getSession().prepare(
+                "DELETE FROM data WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = 0");
+        deleteFromMetricsIndex = session.getSession().prepare(
+                "DELETE FROM metrics_idx WHERE tenant_id = ? AND type = ? AND metric = ?");
+        deleteTag = session.getSession().prepare("DELETE FROM metrics_tags_idx WHERE tenant_id = ? AND tname = ? " +
```
burmanm: Wouldn't it make more sense to just delete a partition? (tenant_id, tname). What's the benefit of giving tvalue, type and metric properties? We could get one row tombstone instead of multiple cell tombstones.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/DeleteTenant.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private MetricsService metricsService;
+
+    public DeleteTenant(RxSession session, MetricsService metricsService) {
+        this.session = session;
+        this.metricsService = metricsService;
+        deleteTenant = session.getSession().prepare("DELETE FROM tenants WHERE id = ?");
+        deleteData = session.getSession().prepare(
+                "DELETE FROM data WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = 0");
+        deleteFromMetricsIndex = session.getSession().prepare(
```
burmanm: This would be better as DELETE FROM metrics_idx WHERE tenant_id = ? and type = ?

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/DeleteTenantITest.java
```diff
@@ -0,0 +1,302 @@
...
+
+    private PreparedStatement getRetentions;
+
+    private String jobName;
+
+    @BeforeClass
+    public void initClass() {
+        tenantCounter = new AtomicInteger();
+
+        getTags = session.prepare(
```
burmanm: Why do we need this info? Referring to previous commits, we don't need these, as we can do delete by (tenant_id, tname) ; no need to fetch tvalue, type or metric information.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/DeleteTenantITest.java
```diff
@@ -0,0 +1,302 @@
...
+                "SELECT tvalue, type, metric FROM metrics_tags_idx WHERE tenant_id = ? AND tname = ?");
+        getRetentions = session.prepare("SELECT metric FROM retentions_idx WHERE tenant_id = ? AND type = ?");
```
burmanm: Same here, (tenant_id, type) is enough for deletions and we have that information already. No need to fetch.

burmanm: And if we need to check that everything was deleted, maybe checking something like "SELECT DISTINCT tenant_id, type FROM retentions_idx WHERE tenant_id = ? AND type = ? should be enough? It fetches filtered partition keys and should return nothing if everything is deleted.

jsanda:
I am recreating this PR. Because of the way this branch was created, I have a bunch of merge conflicts when I rebase. It is easier for me to simply recreate the branch and new PR.

PR #555, opened 2016-07-29
[HWKMETRICS-446] tenant deletion

jsanda:


jsanda:
This is not yet ready to be merged. I need [PR 558](https://github.com/hawkular/hawkular-metrics/pull/558) merged first.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/DeleteTenant.java
```diff
@@ -0,0 +1,107 @@
...
+    private PreparedStatement deleteRetentions;
+
+    private MetricsService metricsService;
+
+    public DeleteTenant(RxSession session, MetricsService metricsService) {
+        this.session = session;
+        this.metricsService = metricsService;
+        deleteTenant = session.getSession().prepare("DELETE FROM tenants WHERE id = ?");
+        deleteData = session.getSession().prepare(
+                "DELETE FROM data WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = 0");
```
stefannegrea: I am really not sure I like the fact that we have CQL statements in different places now. Could we collocate all the job scheduler CQL statements in a single place? This could be addressed in a different PR as long as we open a JIRA.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/DeleteTenantITest.java
```diff
@@ -0,0 +1,322 @@
...
+import rx.Subscription;
+import rx.functions.Action1;
+import rx.schedulers.Schedulers;
+import rx.schedulers.TestScheduler;
+import rx.subjects.PublishSubject;
+
+/**
+ * @author jsanda
+ */
+public class DeleteTenantITest extends BaseITest {
```
stefannegrea: Can you please add tests for deleting non-existing tenants and also requesting twice the deletion of the same tenant?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -379,10 +384,16 @@ private int getDefaultTTL() {
     }
 
     private void initJobsService() {
+        RxSession rxSession = new RxSessionImpl(session);
         jobsService = new JobsServiceImpl();
         jobsService.setMetricsService(metricsService);
-        jobsService.setSession(new RxSessionImpl(session));
-        jobsService.setScheduler(new SchedulerImpl(new RxSessionImpl(session)));
+        jobsService.setSession(rxSession);
+        if (schedulerMode.equals("test")) {
```
stefannegrea: Would it be possible to overlay this class in the context of tests only? And not have it in the main jaxrs code base?

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/rest/TenantITest.groovy
```diff
@@ -172,4 +173,160 @@ class TenantITest extends RESTTest {
 
     assertNotNull("tenantId = $tenantId, Response = $response.data", response.data.find { it.id == tenantId })
   }
+
+  @Test
```
stefannegrea: Can you please add tests for deleting non-existing tenants and deleting the same tenant twice?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
```diff
@@ -71,7 +73,7 @@ public void shutdown() {
     @Override
     public Single<JobDetails> submitDeleteTenantJob(String tenantId) {
         return scheduler.scheduleJob(DeleteTenant.JOB_NAME, DeleteTenant.JOB_NAME, ImmutableMap.of("tenantId",
-                tenantId), new SingleExecutionTrigger.Builder().build());
+                tenantId), new SingleExecutionTrigger.Builder().withDelay(5, TimeUnit.MINUTES).build());
```
stefannegrea: Why a 5 minutes delay? What happens if the user resends the request to delete a tenant?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/DeleteTenant.java
jsanda: While I do not entirely disagree with what you are saying, I am also not in favor of just lumping all data access code into one class for the sake of it because it results in a monolith like we have now. Let's discuss this more and figure out what we want to do going forward and then create some tickets.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/DeleteTenantITest.java
jsanda: Sure, those would be good tests to include. I also want to point that the job is idempotent (as most data access operations should be in Cassandra to the extent possible). Deleting a non-existent tenant should basically be a no-op. Deleting the same tenant twice would result in duplicate work, but because the job is idempotent, nothing would break. With that said, I am thinking it might be worthwhile to add some checks that will avoid submitting a duplicate job to avoid the extra work.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
jsanda: I think we can do that.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
jsanda: The duration of the delay is somewhat arbitrary, but I used a delay to avoid lock contention. Let's say we try to schedule the job right away, which would be the next minute. There is a much greater chance of the job scheduler obtaining the time slice lock to start executing job for the next minute than for some point later on. And if that does happen, the job has to be resubmitted, resulting in more light weight transactions.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
jsanda: To overlay the class means I would have to maintain a separate copy of it in the metrics-api-jaxrs-test module. Having to keep the two source files in sync seems too error prone. I have added a layer of indirection for creating the scheduler instance with a new class JobSchedulerFactory, which does get replaced in the metrics-api-jaxrs-test module.

The config/system property that I initially used to activate using TestScheduler has been removed, which I think is the most important thing.

stefannegrea:
Thanks @jsanda!

PR #558, opened 2016-07-31
[HWKMETRICS-453] add retry policy for non-repeating jobs

jsanda:
A retry policy specifies if and when a job should be retried in the event of a
failure. Retry policies are effective only for non-repeating jobs. If a
repating job fails, its trigger determines when it executes again. By default,
failed jobs will not be retried.


stefannegrea:
Thanks @jsanda!

PR #560, opened 2016-08-01
revert RxJava for perf regression testing (DO NOT MERGE)

jsanda:


FilipB:
Results for this PR shows that for bigger msgs 500 and 5000 requests per sec it's performing much better with older io.reactivex.rxjava version.
Difference is 60-70%. There is almost no difference for small msg.

jsanda:
That is a big difference. RxJava 1.1.8 was released recently, though it looks to be a very small bug fix release. Do we want to test other versions?

jsanda:
RxJava 1.1.9 was released a coupled days ago. I upgraded to it and the perf tests pass now. @FilipB, do you want to do any additional testing to see if this version upgrade resolves the performance drop?

stefannegrea:
Closing this for now because RXJava 2.0 is in beta now and we will upgrade current 1.x once released.

PR #564, opened 2016-08-08
Add in support for relative timestamps.

mwringe:
See https://issues.jboss.org/browse/HWKMETRICS-358

### DIFF
--- a/core/metrics-model/src/main/java/org/hawkular/metrics/model/param/TimeRange.java
+++ b/core/metrics-model/src/main/java/org/hawkular/metrics/model/param/TimeRange.java
@@ -31,22 +31,50 @@ import com.google.common.base.MoreObjects;
 public class TimeRange {
     static final long EIGHT_HOURS = MILLISECONDS.convert(8, HOURS);
 
-    private final long start;
-    private final long end;
-    private final boolean valid;
-    private final String problem;
+    private long start;
+    private long end;
+    private boolean valid;
+    private String problem;
+
+    public TimeRange (String start, String end) {
+        long now = System.currentTimeMillis();
+
+        try {
+            this.start = (start == null || start.isEmpty()) ? now - EIGHT_HOURS : handleOffsets(start);
+            this.end = (end == null || end.isEmpty()) ? now : handleOffsets(end);
+            if (!isValidTimeRange(this.start, this.end)) {
+                valid = false;
+                problem = "Range end must be strictly greater than start";
+            } else {
+                valid = true;
+                problem = null;
+            }
+        } catch (NumberFormatException e) {
+            valid = false;
+            problem = "The start (" + start + ") or end (" + end + ") value does not correspond to a valid number.";
+        } catch (IllegalArgumentException e) {
+            valid = false;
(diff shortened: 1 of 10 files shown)


burmanm:
I think this feature should use Durations for these timestamps (see org.hawkular.metrics.model.param.Duration)

mwringe:
Updated to use durations

mwringe:
retest this please

core/metrics-model/src/main/java/org/hawkular/metrics/model/param/TimeRange.java
```diff
@@ -31,24 +31,52 @@
...
+            problem = "The start (" + start + ") or end (" + end + ") value does not correspond to a valid duration.";
+        }
+    }
+
+    private Long handleOffsets(String number) throws NumberFormatException, IllegalArgumentException{
+        long now = System.currentTimeMillis();
+        if (number.startsWith("+")) {
+            String offset = number.substring(1);
+            Duration duration = new Duration(offset);
+            return now + duration.getValue();
```
stefannegrea: I do not think this is correct. I think you need to get the value in millis of the duration.

core/metrics-model/src/main/java/org/hawkular/metrics/model/param/TimeRange.java
```diff
@@ -31,24 +31,52 @@
...
             valid = false;
-            problem = "Range end must be strictly greater than start";
+            problem = "The start (" + start + ") or end (" + end + ") value does not correspond to a valid duration.";
+        }
+    }
+
+    private Long handleOffsets(String number) throws NumberFormatException, IllegalArgumentException{
+        long now = System.currentTimeMillis();
+        if (number.startsWith("+")) {
+            String offset = number.substring(1);
```
stefannegrea: I would like to see some trimming of white spaces after removing the + or - because the users could send something like `+ 10s`.

core/metrics-model/src/test/java/org/hawkular/metrics/model/param/TimeRangeTest.java
```diff
@@ -50,20 +50,84 @@ public void testLowerBoundDefault() {
...
-        TimeRange timeRange = new TimeRange(123456789L, null);
+        TimeRange timeRange = new TimeRange("123456789", null);
         assertTrue(timeRange.getProblem(), timeRange.isValid());
         assertEquals(123456789L, timeRange.getStart());
         // Unless run on a damn slow system, this should pass
         assertTrue("End should default to now", Math.abs(System.currentTimeMillis() - timeRange.getEnd()) < 2);
     }
 
     @Test
+    public void testRelativeTimeStamps() {
```
stefannegrea: Do you mind adding some tests with deltas with time units other than ms?

mwringe:
Updated to make sure we are receiving the right millisecond values from the duration.
Updated to trim the space after "+" or "-"
Updated tests to check different duration values other than just ms

stefannegrea:
Thanks for the updates @mwringe. I will merge as soon as the perf tests pass.

PR #568, opened 2016-08-11
Filter published metric on bus

lucasponce:
This is an experimental PoC to study the demand publishing pattern inside hawkular-services/metrics.
I have opened the PR with the intention to record the comments but this is just a WIP draft to continue the discussion of HAWKULAR-1102 issue, this could lead in something formal to contribute or definitely discard it, but making technical proposal may help in any case.
First, @FilipB it could be possible for you to build a hawkular-services distribution with these changes and run the performance tests ?
Thanks.


rhqci:

lucasponce:
I don't have rights to put a "don't merge" or equivalent label.
Please, can anyone set it up ?
Thanks.

pilhuhn:
Is there a counterpart for alerts, that publishes the desired metric ids? Otherwise I'd expect that performance is slightly worse as before because of the additional checking done.

FilipB:
ok to test

lucasponce:
It is probably that this jms queue should be added manually in the test environment
https://github.com/hawkular/hawkular-metrics/pull/568/files#diff-bb054d140308871bcca6b1725cb26310L342
If you find additional issues, please, let me know.
Thanks.

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author Lucas Ponce
+ */
+@ApplicationScoped
+@Eager
+public class PublishCommandTable {
+
+    private HashMap<String, Set<String>> published = new HashMap<>();
+
+    public boolean isPublished(String tenantId, String id) {
```
burmanm: Why the structure of HashMap<String, HashSet<>> ? (which is actually HashMap<String, HashMap<..>>). Wouldn't something like HashMap<Key, ..> be better? Just change the definition of the key, even if you need deeper and deeper lookup syntax.

burmanm: Not to mention this tenantId/id is not enough as it's missing "type". In metrics a unique key consists of ``tenantId + metricId + type``

burmanm: .. and after that you could just think this as ``HashSet<Key>`` and remove the whole class.. (and use ``@Provides`` for creating the HashSet and its applicationScope)

FilipB:
@lucasponce 
From performance point of view this PR improves write throughput significantly. It's almost at the same level as standalone h metrics - 28443 req/sec vs. 29553 req/sec. For bigger msgs there is no difference at all.
Memory problem is resolved as well.
Note that no functional testing has been done. I can only tell that https://issues.jboss.org/browse/HAWKULAR-1102 is resolved by this PR

pilhuhn:
Are those numbers with no data to be forwarded or how many metrics are forwarded?

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
lucasponce: Yes, this class can be improved.
Point is that outside of metrics type is not well propagated.
https://github.com/hawkular/hawkular-metrics/blob/master/bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java#L99
This is not a blocker, it simply publishes all metrics with same tenant/name whatever type it is.
Introducing type in other components perhaps might have other side effects.
I put it on the todo list but without less priority.

lucasponce:
These numbers should be without any forwarded, which is in fact the expected behaviour.
Worst case, forwarding all (which means that for all metrics we have trigger defined on it) would lead a similar result as HAWKULAR-1102 (or perhaps a little bit worse as it was noted we have added an additional validation).
To address that, I think that we can optionally buffer the publication of metrics in configurable time period.
Alerting don't need real time metrics, as it computes in time period design.

pilhuhn:
So we should try to come up with a test that forwards 1% and 10% of metrics to see how it works out.

burmanm:
Ärr, don't use rebase yet (after it's ready, squash).. it hides all the older comments that are still valid. The fact that you ignore the id type is dangerous, imagine having metric.server avail and someone creates metric.server gauge. Now they're both transferred to the bus and most likely will cause some lovely corruption.

lucasponce:
Yes, it seems that could be a security issue, not only for this PR but for current hawkular-services.
In HServices the agent controls the metric creation and it doesn't define type as part of the primary key.
This needs to be propagated also into alerting and reviewed in other areas as well.
By the way, where in the doc an end user is aware of that ?
http://www.hawkular.org/docs/rest/rest-metrics.html#Metric
http://www.hawkular.org/docs/rest/rest-metrics.html#POST__gauges
http://www.hawkular.org/docs/rest/rest-metrics.html#_counter
I guess that in the tenant discussion it were said that id should be unique within the tenant (at least for other models) but not unique within tenant + type tuple.
Anyway, this is just a comment, I will apply the changes.
P.S.: It is one-click to show hidden comments after a PR sync, that style of syncing is the default one in MiQ.

burmanm:
It's mentioned in http://www.hawkular.org/hawkular-metrics/docs/user-guide/#_metric_ids

The MiQ style is idiotic and is probably something they copied over from CVS processes, we don't have to copy their style here. It makes reviewing further updates difficult as I can't view just the changes you made after the last commits.

stefannegrea:
@lucasponce, what happens with the cache when the Metrics/Services server is restarted?

lucasponce:
In this draft, it is not persisted yet.
Options are to use C* for this task (I think it could be a good option) or thinking further we can even use ISPN local/distribute instance for this (thinking in an scalable solution).

lucasponce:
It has been added:
- Buffer forwarding into bus in a configurable way (in a similar way as alerting processes messages).
- PublishCommandTable implemented as ISPN cache (loader/store can be defined via configuration).
Please, could you review /cc @stefannegrea @burmanm @jsanda .
Also /cc @jshaughn as it is interesting you participate too.
I haven't implemented string type messages on this PR, that is a side task that can be out of this one.
This PR needs sync with Alerting (HWKALERTS-156) before it can be integrated on hawkular-services.
I have noticed that the integration-tests/hawkular-server-tests are not active, so I wonder if there is some lack of integration testing somewhere, hawkular-services only have smoke tests we could increase the coverage re-activating this or perhaps putting in a different level.
That could be a discussion for a separate thread as well, just leaving a note here for the record.

bus/src/main/java/org/hawkular/metrics/component/publish/AvailDataMessage.java
```diff
@@ -104,6 +104,8 @@ public String toString() {
         @JsonInclude
         private String tenantId;
         @JsonInclude
+        private String type;
```
stefannegrea: Not sure what this is? What is availability type?

lucasponce: It is similar as MetricDataMessage.SingleMetric.type field used to build the dataId used internally in alerting.
Probably it is not necessary as it can be used as a constant, but then if the datatype changes in metric (i.e. new availability types in the future) constant needs to be in sync and updated on alerting too.
This is a redundant field that helps to unify that flow.

jshaughn: I don't think we should add this, it's an unlikely future and can cause (actually has caused) confusion.

lucasponce: Ok, thanks for the feedback, it seems it creates more complexity.
I am reverting this one.

lucasponce:
The preliminar PRs for metrics and alerts (https://github.com/hawkular/hawkular-alerts/pull/197) are ready for review.
As commented, there is a lack of integration tests that should go in hawkular-services in the case these PRs are accepted and merged in both components.

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
```diff
@@ -40,7 +40,9 @@
             "CASSANDRA_MAX_REQUEST_CONN", false),
     WAIT_FOR_SERVICE("hawkular.metrics.waitForService", null, null, true),
     DEFAULT_TTL("hawkular.metrics.default-ttl", "7", "DEFAULT_TTL", false),
-    DISABLE_METRICS_JMX("hawkular.metrics.disable-metrics-jmx-reporting", null, "DISABLE_METRICS_JMX", true);
+    DISABLE_METRICS_JMX("hawkular.metrics.disable-metrics-jmx-reporting", null, "DISABLE_METRICS_JMX", true),
+    METRICS_PUBLISH_DELAY("hawkular-metrics.publish-delay", "1000", "METRICS_PUBLISH_DELAY", false),
+    METRICS_PUBLISH_PERIOD("hawkular-metrics.publish-period", "2000", "METRICS_PUBLISH_PERIOD", false);
```
jshaughn: Maybe 2s is too long of a delay?  The longer the delay the slower the alert generation and the greater the chance that data does not get processed by alerting on a shutdown.   Also, is there an option to not delay at all and push each datum immediately?

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandMessage.java
```diff
@@ -0,0 +1,164 @@
...
+package org.hawkular.metrics.component.publish;
+
+import java.util.List;
+
+import org.hawkular.bus.common.AbstractMessage;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * A bus message that will indicate which metrics will be published/unpublished to the bus.
```
jshaughn: Let's perhaps not call this published to the bus but rather just published for Alerting consumption.  I think what we're doing here is more tightly coupling these two components, which is not a bad thing.

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.enterprise.context.ApplicationScoped;
+
+import org.hawkular.metrics.api.jaxrs.util.Eager;
+import org.hawkular.metrics.model.MetricId;
+import org.infinispan.Cache;
+
+/**
+ * A table to store a cache of published metrics on the bus under hawkular-services context.
+ * This is a PoC and this cache is only valid for single node sceanarios.
+ * Clustered scenarios requires to transform the local HashMap into a ISPN cache.
```
jshaughn: Could we not just use ISPN in any configuration?  Can Metrics and Alerting share an ISPN cache such that Alerting can maintain the metricIds it needs and Metrics can read from there to determine which datums to publish to JMS?

jsanda:
I want to briefly reiterate what I said on the hawkular-dev list. I understand that the motivation for these changes is performance-related. Bypassing JMS altogether in favor of direct method invocations (while still being loosely coupled) is absolutely going to yield better performance. Neither the core metrics nor core alerts code need to have dependencies on one another. It would be handled in the integration layers just as it is today. Rather than receiving notifications from a JMS topics, alerts would receive emissions from an Observable which could support back pressure. 

This approach will improve performance, simplify architecture, and keep loose coupling. Unless there is a really compelling reason to use JMS, I think I am -1 on the approach in this PR.

lucasponce:
I see there is no consensus.
This PR resolves the performance issues whithin the current architecture as preliminar QE tests solves.
The approach to coupling two different application together in the same .war and using an Observable is a complete different paradigm that I don't share.
I escalate this to @theute and @pilhuhn they will decide.
I tried to do my best whithin the same architecture but this is beyond that.
It is again changing an architecture that was decided and it was supposed to follow.
If business requeriments have been changed we should be notified about that and have these things clear.

stefannegrea:
@lucasponce, I think there is consensus that we need to do something however @jsanda just voiced some concerns with regards to the solution proposed in this PR.

This PR just masks system crashes when there is a huge load on the bus. With these proposed changes publishing of stuff on the bus is reduced and consequently the system does not crash. But that still does not address the main problem, the system will still crash under heavy JMS load. In addition to the current problem there are right now two open issues regarding the use of JMS in Hawkular Services around startup.

If this PR is a provisional solution then probably there would be more consensus around these changes. But I did not see an indication regarding this. What @jsanda proposed is aimed at solving the main problem, system crashes under heavy load.

@lucasponce , I think we just need a little bit more discussion around the issue before making a decision.

jshaughn:
I'm not totally sure we should close the PRs just yet.  As Lucas mentions, this does provide a solution to the problem at hand. And it's a;ready done.

John's alternative also has merit and I think the biggest issues there are to figure out the packaging and also how Alerting informs metrics of the relevant dataIds.  So, I'd agree that we need @pilhuhn and/or @theute to decide whether to refine this PR and go with it, or to commission an effort to try John's approach.

jsanda:
I have a fully plate of tasks at the moment, but when I am bit more freed up, I'd be happy to help with packaging. With my proposal, there is no need for alerts to inform metrics about ids. Alerts can simply filter the emissions it receives.

lucasponce:
As discussed there are short term and mid/long term approaches.

With a code freeze in less than one month this PR aimed to offer a solution
to a business issue within the current architecture.

This PR just masks system crashes when there is a huge load on the bus.
> With these proposed changes publishing of stuff on the bus is reduced and
> consequently the system does not crash. But that still does not address the
> main problem, the system will still crash under heavy JMS load. In addition
> to the current problem there are right now two open issues regarding the
> use of JMS in Hawkular Services around startup.
>
I am still having the open question if "under heavy JMS load" is a business
requeriment or a consequence of forwarding everything, so there is still
subjective (but valids, I don't say that all angles could be ok) views on
that.
 If this PR is a provisional solution then probably there would be more
consensus around these changes. But I did not see an indication regarding
this. What @jsanda <https://github.com/jsanda> proposed is aimed at solving
the main problem, system crashes under heavy load.

> @lucasponce <https://github.com/lucasponce> , I think we just need a
> little bit more discussion around the issue before making a decision.
>
Again, as discussed we are in several timeframes, short and mid/long term.

I don't see that coupling components in a quick way is going to be a good
solution for a short term status as we are.

That is potentially more risky than still using the current architecture
which hawkular-services has defined.

For mid/long term, as discussed, open the discussion to the main
architecture is valid for me.

So I would appreciate if we have some clarification from @pilhuhn
<https://github.com/pilhuhn> and/or @theute <https://github.com/theute> in
which status we are (short/mid/long).

There are still too many variables to introduce new ones with a big change
on the architecture.

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
lucasponce: I missed your review on the PR @jshaughn, sorry, responding now.
This 2s delay is in line with the period used in alerting for computing, so it is configurable to tweak it as result of performance tests.
I don't see it needs to be increased in our business domain, we are not doing real time alerting.
For example, MiQ has a default evaluation time of 10 minutes.
Having said that, the implementation lets you to set the value you need.

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandMessage.java
lucasponce: +1

bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
lucasponce: Yes, indeed the table where ids are stored is backed by an ISPN.
Updating this shared ISPN from alerting will reduce the logic of this approach and it will simplify it in general.
+1

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
jshaughn: That's fine, although I don't think we should be guided by MIQ's impl for this :)  In general, people want alert as soon as possible.

theute:
@pilhuhn is on PTO now and I don't take architectural decision.

But as I said I never had issues about having Metrics+Alerts+... packaged together. I just never wanted Metrics+Alerts+UI all mixed in a large piece of mud and have clear responsibilities for each part, it doesn't have much consequences on how they work together (I initially though direct method invocation, but we went with a bus design, and again, I don't give the architectural decisions).

As I said, I haven't seen many (any ?) situation when one wants metrics without alerts, so it's definitely critical that they work well together, so thanks for providing a solution !

From usecases:
   - I see situations where you want to send data to the alert engine without necessarily need to go through Metrics and store the data.
   - I see situations where you want to store data but not need to evaluate the data for alert.

So I would not send all the data to both, either the client knows if it needs to be stored and evaluated (and then query the appropriate API, a glue API, not multiple calls), and/or something configurable on the server side need to do the filtering job.

Now we need to be pragmatic and changing API now would be a problem. Also if there is no alternative and as it seems to improve the situation, it sounds to me like we need to go forward anyway for this release unless I missed something...

PR #569, opened 2016-08-12
[HWKMETRICS-458] explicitly set keyspace due to cassalog bug

jsanda:
@mwringe can you test this since the error I got was different than what you reported?


mwringe:
@jsanda LGTM

PR #571, opened 2016-08-17
[HWKMETRICS-1] initial support for roll ups

jsanda:
This PR adds initial support for generating the following rollups:

* 1 minute
* 5 minute
* 1 hour

I suspect the roll ups used can/will change, and in fact different sets of rollups may make more sense from one deployment to another, so I don't want the review process to get hung up on 5 minutes vs 15 minutes for example.

I have only updated one /stats endpoint so far because I want to get the review process started. This is also a big change in that it introduces Infinispan distributed caches into our architecture.

For right now, at least temporarily, I have rollups enabled automatically so that we can see what the impact is with the perf test job. On inserts each data point is now also stored in a per-minute cache. The rollup job runs every minutes and writes the 1 minute data points to Cassandra. This introduces a lot more load on Cassandra. If the perf test job is writing N metrics with 1 data point each, then every minute we are doing 2N writes, and every 5 minutes we are doing 3N writes. Given this, I necessarily expect to see some performance degradation


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -383,6 +387,17 @@ protected void initPreparedStatements() {
...
                 "WHERE tenant_id = ? AND tname = ? AND tvalue = ?");
+
+        find5MinStats = session.prepare(
+                "SELECT time, max, min, avg, median, samples, sum, percentiles " +
+                "FROM rollup300 " +
+                "WHERE tenant_id = ? AND metric = ? AND shard = 0");
+
+        find5MinStatsDesc = session.prepare(
+                "SELECT time, max, min, avg, median, samples, sum, percentiles " +
+                "FROM rollup300 " +
```
burmanm: Do we need new table for each rollup? Wouldn't it make sense to have something like "tenant_id, metric, shard, rollup_period" in the partition key?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/cache/CacheServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.hawkular.metrics.core.service.transformers.NumericDataPointCollector;
+import org.hawkular.metrics.datetime.DateTimeService;
+import org.hawkular.metrics.model.DataPoint;
+import org.hawkular.metrics.model.MetricId;
+import org.infinispan.Cache;
+import org.infinispan.commons.util.concurrent.NotifyingFuture;
+import org.infinispan.manager.DefaultCacheManager;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.joda.time.Duration;
```
burmanm: Lets not add this to the new code? joda.Time is deprecated for JDK8.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/rollup/RollupServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        percentiles.forEach(p -> map.put((float) p.getQuantile(), p.getValue()));
+        return map;
+    }
+
+    @Override
+    public Observable<List<Rollup>> getRollups(MetricId<Double> id, int rawTTL) {
+        return Observable.just(asList(
+                new Rollup(RollupBucket.ROLLUP60.getDuration(), rawTTL * 2),
+                new Rollup(RollupBucket.ROLLUP300.getDuration(), rawTTL * 4),
+                new Rollup(RollupBucket.ROLLUP3600.getDuration(), rawTTL * 6)
```
burmanm: Not a huge fan of this. Could we have different TTL settings just for each duration rather, that can be set? I could see user wanting to keep these far longer time than raw data.

jsanda: Consider the `getRollups` method throw away code. I just wanted to have something in place quickly for some prototyping. There needs to be reasonable data retention defaults, and they also need to be configurable.

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.19.0.groovy
```diff
@@ -24,4 +24,70 @@ schemaChange {
...
+  version '3.1'
+  author 'jsanda'
+  tags '0.19.x'
+  cql """
+CREATE TABLE rollup60 (
+    tenant_id text,
+    metric text,
+    shard  bigint,
+    time timestamp,
+    min double,
```
jsanda: >What about map of aggregates? 
What exactly do you have in mind?

>Also, just one table for rollup60, rollup300, rollup3600 is necessary in my opinion.

This is something I've gone back and forth with for data modeling in general with Cassandra. Of the other Cassandra based systems I have looked at, the majority break things out into multiple tables. What is the basis for your opinion?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/cache/CacheServiceImpl.java
jsanda: I agree that we should switch to the new Time API in Java 8, but it is a low priority IMO. There are some critical functions we use for rounding dates/timestamps that we need to make sure are properly ported.

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.19.0.groovy
burmanm: Instead of min, max, avg, median (which is 50% percentile, so it should be with percentiles..), sum, percentileA, percentileB etc.. soon someone will want "std deviation", "mode", "variance" and so on. And we always need new columns and we might only want to calculate part of the information as well. I would like to avoid that, instead getting a map of different aggregates (like the data table placeholder) which would then be returned to the requester.

Same applies to the more tables, user configurable rollup periods is something we could relatively easily provide and they would always use the same table -> no change in schema required if everything is in same table.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/cache/CacheServiceImpl.java
burmanm: There's nothing in this class that would require the use of Duration for older methods. It's used as a convenience for Duration.standardMinutes(1);. No harm would be done to migrate that to Java 8 standard API. This is new code.

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.19.0.groovy
jsanda: I forgot to include some relevant details when I opened the PR. I was going back and forth about whether or not to use typed columns versus storing everything as a blob for the very reasons you mention. It mean more encoding/decoding on the client side, but it gives us more flexibility and maybe even more efficient storage. I am not sure that a CQL map gives us that same flexibility.

I have also considered the same things about 1 table versus many. Using one table allows us to introduce different rollups without requiring schema changes. If we go that route, we also need to create Infinispan caches on demand. Creating the different rollups and caches is easy enough, but keeping track of everything could get cumbersome.  I do question how much of that flexibility we will need in practice.

jsanda:
retest this please

PR #572, opened 2016-08-18
Allow rate queries for read-only users

jotak:
The POST endpoint /counters/rate/query is used to read data, it should be authorized for read-only users (similar to /counters/raw/query)


rhqci:

burmanm:
ok to test

burmanm:
@mwringe Do you need this in the 0.18.x branch?

mwringe:
Yes, we will need this in the 0.18 branch as well

PR #574, opened 2016-08-19
HAWKULAR-1102 Filter published metric on bus

lucasponce:
Re-opening this topic with @jshaughn recommendations.
Basically it performs two changes on mainstream:
- Filtering ids instead to forward all messages to JMS.
- Buffering forwarding.
Ids are stored in a ISPN cache shared between components.
(Grouping the previous commits into one for clarity).
Discussion are still alive and we are waiting some feedback from @theute and @pilhuhn.

Please, comment or close if you thing this direction has been canceled, reading on the dev list or in the comments is not clear about that.


bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
```diff
@@ -131,5 +149,43 @@ void shutdown() {
...
+                        publishNumeric(numeric);
+                    }
+                });
+            }
+        }
+    }
+
+    private synchronized List<Metric> getAndClearPendingMetrics() {
+        ArrayList<Metric> result = new ArrayList<>(pendingMetrics);
+        pendingMetrics.clear();
```
burmanm: This can cause a data loss. What's the point of synchronized method when pendingMetrics is not protected against concurrent modifications?

You should use a lock here that's the same as with the "onInsertedData" to prevent data being added between fetching the results and clearing.

Just ditch the synchronized keyword. It doesn't protect the pendingMetrics list here at all.

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
```diff
@@ -64,24 +71,35 @@
...
         log.tracef("Inserted metric: %s", metric);
-        if (metric.getMetricId().getType() == AVAILABILITY) {
-            @SuppressWarnings("unchecked")
-            Metric<AvailabilityType> avail = (Metric<AvailabilityType>) metric;
-            publishAvailablility(avail);
-        } else {
-            @SuppressWarnings("unchecked")
-            Metric<? extends Number> numeric = (Metric<? extends Number>) metric;
-            publishNumeric(numeric);
+        if (publish.isPublished(metric.getMetricId())) {
```
burmanm: Also, since RxJava is used here otherwise.. this code should follow RxJava patterns.

Basically, instead of this codeblock, use something like

```java
metrics.filter(m -> publish.isPublished(m.getMetricId())).map(pendingMetrics:add)
```

Or make it even prettier with some more reactive handling.

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
```diff
@@ -131,5 +149,43 @@ void shutdown() {
...
+    private int getPublishPeriod() {
+        try {
+            return Integer.parseInt(publishPeriodProperty);
+        } catch (NumberFormatException e) {
+            log.warnf("Invalid publish period. Setting default value %s", METRICS_PUBLISH_PERIOD.defaultValue());
+            return Integer.parseInt(METRICS_PUBLISH_PERIOD.defaultValue());
+        }
+    }
+
+    private class PublishTask extends TimerTask {
```
burmanm: Same here, we already have RxJava, why not use proper reactive methods here? No TimerTask is needed, just subscribe to the observable, do the filtering and publishing.

Reactive, immediate (with optional backpressure / windows if you're afraid of JMS) and fits nicely with the published format of the metrics (Observable pattern)

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
burmanm: And even better, if you wanted batched outputs.. you can omit this whole method and use a window of time X (and avoid all the concurrency issues).

lucasponce: Yes, this is a bug, it should protect also the add() operation.
I am not sure if I understand your proposal of using window of time.

burmanm: http://reactivex.io/documentation/operators/window.html

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
lucasponce: If there is an alternative to do the same here, I am fine with that.
Perhaps I'm re-implementing some pattern that is well known in the RxJava.
As this PR is relatively small, could you send a commit with that proposal on it ?

burmanm: https://gist.github.com/burmanm/040c95f71c71dbc36f3b8a9a5367aa53

That should accomplish everything you've done to this class. If you want the buffer size to be configurable, then you'll need to return those settings, although I'm not sure why you want the buffer window at all (does it help with JMS performance issues?)

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
```diff
@@ -40,7 +40,9 @@
             "CASSANDRA_MAX_REQUEST_CONN", false),
     WAIT_FOR_SERVICE("hawkular.metrics.waitForService", null, null, true),
     DEFAULT_TTL("hawkular.metrics.default-ttl", "7", "DEFAULT_TTL", false),
-    DISABLE_METRICS_JMX("hawkular.metrics.disable-metrics-jmx-reporting", null, "DISABLE_METRICS_JMX", true);
+    DISABLE_METRICS_JMX("hawkular.metrics.disable-metrics-jmx-reporting", null, "DISABLE_METRICS_JMX", true),
+    METRICS_PUBLISH_DELAY("hawkular-metrics.publish-delay", "1000", "METRICS_PUBLISH_DELAY", false),
```
lucasponce: This delay property is not needed.
Fixing it on current commit.

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
```diff
@@ -64,8 +68,21 @@
...
+    @ConfigurationProperty(METRICS_PUBLISH_PERIOD)
+    private String publishPeriodProperty;
+    private int publishPeriod;
+
     public void onMetricsServiceReady(@Observes @ServiceReady ServiceReadyEvent event) {
-        Observable<List<Metric<?>>> events = event.getInsertedData().buffer(50, TimeUnit.MILLISECONDS, 100)
+        publishPeriod = getPublishPeriod();
+        Observable<List<Metric<?>>> events = event.getInsertedData()
+                .filter(m -> m.getType() != MetricType.STRING)
+                .filter(m -> publish.isPublished(m.getMetricId()))
```
jshaughn: @lucasponce Is this comparing apples to apples?  is m.getMetricId() going to match the id placed in the cache by alerting?  It may be OK, just checking to make sure alerting is not using a type-prefixed value and this value perhaps not being type-prefixed.  Also asking whether tenant needs to be involved.

jshaughn:
I think this is looking pretty good now, I only have the one comment about the filtering test.  @burmanm thanks for the excellent review feedback and Rx help.

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
lucasponce: It is filtering by the MetricId type 
https://github.com/hawkular/hawkular-metrics/pull/574/files#diff-c8641cf1ba58303ba15fba81de0f0c1aR42
Which is supplied from alerting in the proposal
https://github.com/hawkular/hawkular-alerts/pull/197/files#diff-d64b0cb70fe108f47cd45e05615cec19R160

If I haven't introduced any bug I think the design is valid.

lucasponce:
By the way, locally the tests passed correctly but I can see some errors in the CI 
https://travis-ci.org/hawkular/hawkular-metrics/builds/153549900#L6525
They don't seem related to the changes of this PR but in anyone could take a look we can confirm it.

bus/src/main/java/org/hawkular/metrics/component/publish/InsertedDataSubscriber.java
jshaughn: OK, I see what you are doing, you are stripping the HMetrics prefix on the alerting side, when putting the metricId in the cache.  Ignoring tenantId and type is likely fine for filtering, at worst we process some unnecessary metrics in Alerting.

It's unfortunate that Hmetrics decided to make metric uniqueness involve type.  I think it may have been better if they assigned a type to a metric but still disallowed duplicate metricIds within a tenant.  Now there really is no choice but to ensure that Halerting dataIds for Hmetrics data are type-prefixed.  I think it could be useful to change the prefixes to be something more specific to Hmetrics, so that there can be no confusion between dataIds defined for Hmetrics metrics and dataIds for data being fed from other sources.  Instead of: AVAILABILITY, GAUGE, etc, perhaps HM_A_, HM_G, etc.  Something more unique.  All clients creating triggers for Hmetrics metrics will need to follow the dataId naming convention.

lucasponce: Ok, I'll apply this suggestion into the alerts PR.

jsanda: > It's unfortunate that Hmetrics decided to make metric uniqueness involve type. I think it may have been better if they assigned a type to a metric but still disallowed duplicate metricIds within a tenant. 

Metrics does not check or enforce uniqueness of ids. Performing those kinds of checks would have to be done on every write which would be bad performance-wise. There is another important reason why the type is part of the id. We use a statically typed schema
There are a couple reasons for having the type as part of the id. Values are stored either as a double, long, text, etc. We infer the data type of values based on the URI of the HTTP request. We need to store that type info somewhere. Storing it as non-primary key column doesn't make sense as it would have to be repeated for every data point. Alternatively, we could have used a separate table for each metric type, but we would still be in the same situation.

jshaughn: @jsanda thanks for the explanation, it all makes sense, I was just whining :)

lucasponce:
As commented in the last meeting, on this PR now there is additional properties to enable/disable whole forwarding and filtering.
It should be ready from my side.
Please @jshaughn @stefannegrea, let me know if there is any pending action for this task.
Thanks.

jshaughn:
@lucasponce From my side I have nothing blocking these PRs.  When @stefannegrea is OK we can merge, I think.  Stefan, please let Lucas know of anything he has yet to address so that we can move this forward. Thanks, Jay

stefannegrea:
@lucasponce, what happens to the cache if the filtering is disabled for a while and then re-enabled. Will the cache be updated to reflect any alert definition changes after the feature is re-enabled?

lucasponce:
@stefannegrea as commented we are going to add a flag into the alerting to enable/disable filtering and also to clear/recreate the cache at startup.

stefannegrea:
Based on the clarification from @lucasponce and @jshaughn the additional features for the cache management will be added to Alerts component. This PR is ready to merge as is.

lucasponce:
One last comment, the shared cache needs to be defined too in hawkular-services.
On the upgrade of metrics version in HS this definitions needs to be added.

PR #575, opened 2016-08-19
[HWKMETRICS-422] Avoid writing to the metrics index on data point inserts

stefannegrea:
Avoid writing to the metrics index on data point inserts. However, keep all the other writes to the metrics index in place.

Here are the changes to query metrics:
1) When requesting info about a single metric
  a) query the metrics index first, if found then return the info (just like before this change)
  b) if no result there, query the data table for a single data point, if found then the metric exists just not in the metrics index
  c) if no result in either metrics index or data table then there is no such metric

2) When requesting a list of all metrics for a tenant
  a) query the metrics index first
  b) query the data table for all metrics and filter by tenant and type if needed
  c) concat the two lists
  d) return only distinct values; note that the results from metrics index take precendence since they contain info about data retention and tags.

Note: After this change the metrics index will still contain metrics that have tags, non-default retention values, or have been explicitly created by the user. Metrics that just have data points will not have an entry in the metrics index table.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -963,11 +963,7 @@ public class MetricsServiceImpl implements MetricsService {
 
     @Override
     public Observable<Boolean> idExists(final MetricId<?> metricId) {
-        return this.findMetrics(metricId.getTenantId(), metricId.getType())
-                .filter(m -> {
-                    return metricId.getName().equals(m.getMetricId().getName());
-                })
-                .take(1)
+        return this.findMetric(metricId)
                 .map(m -> Boolean.TRUE)
                 .defaultIfEmpty(Boolean.FALSE);
     }
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -47,6 +47,8 @@ import java.util.regex.Pattern;
 import org.hawkular.metrics.core.service.log.CoreLogger;
 import org.hawkular.metrics.core.service.log.CoreLogging;
 import org.hawkular.metrics.core.service.transformers.ItemsToSetTransformer;
+import org.hawkular.metrics.core.service.transformers.MetricFromDataRowTransformer;
+import org.hawkular.metrics.core.service.transformers.MetricFromFullDataRowTransformer;
 import org.hawkular.metrics.core.service.transformers.MetricsIndexRowTransformer;
 import org.hawkular.metrics.core.service.transformers.NumericBucketPointTransformer;
 import org.hawkular.metrics.core.service.transformers.TaggedBucketPointTransformer;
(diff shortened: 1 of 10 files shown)


FilipB:
@stefannegrea This PR increased performance ~+10%. Please let me know before merging so I can update baselines.

jsanda:
Overall I think the changes look good. Do you have any idea how the new versions of findMetric(s) compare performance-wise with the current version?

jsanda:
@FilipB, we are ready to merge this. Please let me know when you have updated the baselines so we can merge. Thanks.

burmanm:
I would say don't merge yet. The MiQ integration will use the findMetric(id) query a lot in the future. Why do we need to fetch the ids if there's no information attached to them?

stefannegrea:
@burmanm, until now a metric's existence could be checked by looking into the metrics index. But going forward, the metrics index does not get updated on data point insertions. The method still needs to reliably return answer the question if the metric exists or not. Otherwise we get into all sorts of weird situations. A simple such example is: findMetric(id) says the metric does not exist but you can still query data points for it.

findMetric(id) still hits the metricsindex just like it did today, so nothing changed in terms of performance if the metric is in the index. If that is a miss it does a query against the primary key of the data table. From what I understand that query against the primary index is very fast since the primary key is stored in memory and also note the the number of results is limited to 1 so only a single result is returned. There is not much difference between the performance of findMetric before and after this change.

burmanm:
@stefannegrea There's no difference to current behavior, we don't check if metric exists when you query for datapoints, so there's no point in querying the partition table. So checking against partition key gives no benefit to anything we have today.

Also, findMetricInData should just request DISTINCT tenant_id, type, metric, dpart. No need for LIMIT 1 or that sort of things (and why only request tenant_id, metric? In this behavior, Cassandra is pure row storage).

stefannegrea:
>@stefannegrea There's no difference to current behavior, we don't check if metric exists when you query for datapoints, so there's no point in querying the partition table. So checking against partition key gives no benefit to anything we have today.

@burmanm, this is really confusing. The data queries (row or stats) are unchanged in this PR. The only thing that changes is explicitly querying for a metric via `/{metric_type}/{metric_id}` end-points. If you use those end-points then findMetric(id) gets called and it has the extra query if the metric is not in the metrics index.

stefannegrea:
@burmanm, I found a `findMetric(id)` call on one of the stats methods that can be removed. I just committed the code. So now my previous comment is accurate.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -940,11 +940,7 @@ public void setDefaultTTL(int defaultTTL) {
 
     @Override
     public Observable<Boolean> idExists(final MetricId<?> metricId) {
```
burmanm: My IDE can't actually even find this being used anywhere. Any idea why we have it?

stefannegrea: I suspect it was used in one of the earlier versions of the JAX-RS implementations. Should we get rid of it? I see it as a useful method on this core service class.

stefannegrea:
I had to rebase on master since there were lots of merge conflicts.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jsanda: +1 to getting rid of it. It was part of the API from the early part of the project and I believe used with the old metrics explorer. I am not aware of any other uses.

stefannegrea: Removed in the last commit.

jsanda:
I took a closer look at the new `findMetricInData` query. Run into with tracing turned on in cqlsh to see what happens. Here is my output https://goo.gl/y80TAY. Notice lines 15 and 16. It mentions executing a single partition query and acquiring sstable references. Now constrast that with this query https://goo.gl/Jir3Xu. It is clear that the first query is a regular query that may touch sstables. I believe the latter query will only read the partition index. The latter of course is optimized because with a regular query, Cassandra reads the partition index and then from possibly multiple sstables.

stefannegrea:
@jsanda, the trace is really interesting however `findMetric(id)` method is only used when explicitly querying for a metric definition or enabling `fromEarliest` on a data query. Prior to this PR, that method was used when computing stats, but that is unnecessary and the call was removed in this PR.

stefannegrea:
@jsanda, the first query that you traced has an additional `LIMIT 1`. Would that help the performance or response time?

jsanda:
As for the usage of `findMetric` @burmanm mentioned earlier that he anticipates increased usage of the method for MiQ so it might be worthwhile to have some idea of the performance tradeoff.

As for performance I don't think that  limiting the results set is all that relevant here. The first query does more disk I/O than the second query whether there is a limit or not.

stefannegrea:
Here are some tests done on my local box. I used the same hardware and ran to code in master and in this PR.

Premise:
1) Create a fresh single node C* cluster
2) Insert 256k gauge data points each associated with a random metric id via `/gauges/{id}/raw/` 
3) Get metric definitions for 256K random metric ids via `/gauges/{id}` endpoint  (note: the metric ids queried for a not the ones inserted)
4) Repeat the experiment 3 times

Master (before this change)
``` 
Insertion Times: 52.728 seconds, 51.173 seconds, 48.074 seconds

Query Times: 32.782 seconds, 33.855 second, 32.784 seconds
```

jsanda:
Since Stefan's perf testing didn't reveal any issues with the change, I am going to merge. The improved write throughput coupled with significantly reducing the number of requests to Cassandra is a big win. @FilipB can you update the baselines?

PR #578, opened 2016-08-22
[HWKMETRICS-461] Allow fetching multiple metric definitions with id list

burmanm:
If no tags are defined, but id filter is, go for exact matching with a list separated by `|`

@lponce


stefannegrea:
Thank you @burmanm!

PR #579, opened 2016-08-22
HWKMETRICS-450: Removing @Eager to make Hawkular Services start

theute:
Note: this is pure trial and fail/success attempt...

On my machine, Hawkular Services starts when those 2 @Eager are removed, but I don't know what else needs to be tested...


theute:
@tsegismont Would be great if you could explain why you added @Eager and if you think we can remove those....

rhqci:

jsanda:
We don't have the end-to-end integration tests to really be able to verify this, so I am going to merge so the change can get picked up for testing in hawkular-services. Thanks @theute!

PR #580, opened 2016-08-22
[HWKMETRICS-456] Just de-duplicate code added for binding queries wit…

stefannegrea:
…h and without TTL.


burmanm:
retest this please

PR #584, opened 2016-08-24
[HWKMETRICS-465] New POST endpoints gauges/stats/query and counters/stats/query

jotak:
It's a kind of duplication of the GET endpoints "gauges/stats" and "counters/stats".
The reason is the same as the one explained here: https://issues.jboss.org/browse/HWKMETRICS-410 and which ended up in duplicating GET "gauges/raw" endpoint to POST "gauges/raw/query" in a similar way.

Also added integration tests

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AggregatedStatsQueryRequest.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AggregatedStatsQueryRequest.java
@@ -16,20 +16,18 @@
  */
 package org.hawkular.metrics.api.jaxrs;
 
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
 import com.google.common.base.MoreObjects;
 
 /**
- * @author jsanda
+ * Request object used to provide stats on multiple metrics that are aggregated
+ * @author Joel Takvorian
  */
-public class StatsQueryRequest {
+public class AggregatedStatsQueryRequest {
 
-    private Map<String, List<String>> metrics = new HashMap<>();
+    private List<String> metrics;
 
     private String start;
 
@@ -43,13 +41,13 @@ public class StatsQueryRequest {
 
     private String tags;
 
-    private List<String> types = new ArrayList<>();
+    private boolean stacked;
 
-    public Map<String, List<String>> getMetrics() {
+    public List<String> getMetrics() {
         return metrics;
     }
 
-    public void setMetrics(Map<String, List<String>> metrics) {
+    public void setMetrics(List<String> metrics) {
         this.metrics = metrics;
     }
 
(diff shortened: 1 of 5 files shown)


rhqci:

jsanda:
Do we really need this? Couldn't we just use the `POST /hawkular/metrics/metrics/stats/query` endpoint which allows you to query for stats from gauges, counters, and rates in one request?

jotak:
metrics/stats/query works differently and does not perform series aggregation. You provide a list of metric ids and it returns, for each metric, the statistics.

Here this is really a kind of alias of  _gauges/stats_ : all time series are aggregated together. There's also this "stacked" parameter which is useful for such case, and doesn't exist in metrics/stats/query (not only it's doesn't exist, but it wouldn't make sense: why would we stack a gauge over a counter?). I think they are really for different purpose.

jotak:
For instance try to query 2 metrics on /metrics/stats/query , then the same metrics on /gauges/stats: you get different results

jotak:
Maybe they should be renamed for disambiguation? Like "/metrics/stats/query" vs "/gauges/aggregated-stats" ?

jsanda:
Right, /metrics/stats/query basically batches multiple results to save on HTTP requests. Is the main motivation for the endpoints to work around the Grafana issue? Just want to make sure I understand since as you said there is some duplication here.

jotak:
Yes, this is for grafana. I've also heard that @mwringe had similar issues with other endpoints, so maybe that can be useful for other clients of the api

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AggregatedStatsQueryRequest.java
```diff
@@ -0,0 +1,141 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * @author jtakvorian
+ */
+public class AggregatedStatsQueryRequest {
```
jotak: The same class may be used for avail to request mixed availability.
So maybe it should be renamed "MixedStatsQueryRequest"

jotak:
Can we get a status on that? Is this endpoints acceptable or not, given there was a precedent with _gauges/raw/query_ and _counters/raw/query_ .

If not acceptable what do you suggest to make it work with Grafana? We could also provide a specific location for grafana related workaround, like _/grafana/gauges/raw/query_ , _/grafana/gauges/stats/query_ etc. But I must say I'm not very keen on that since those endpoints may be useful for others than grafana.

jotak:
Can someone label this PR "DO NOT MERGE"?
I would like to start a more global discussion on the REST API before merging. I'll wait for @jsanda and @pilhuhn being back to work, but basically, I want to point out some inconsistencies on the REST API and discuss about how to fix it.

pilhuhn:
@jotak I think the best way to discuss it is an email to hawkular-dev ML.
Also please keep in mind that there are clients out there that expect some endpoints to exist. We can't just "move them around".
Rest allows though to alias via a 302 redirect or similar to keep an old endpoint url around.

jotak:
So following the previous discussions on ML, I think it's now ok to review code & merge

jotak:
Build failure => "Trigger time has already passed" :(

stefannegrea:
ok to test

stefannegrea:
@jotak, there is one parameter missing from the request object: `fromEarliest`. Do you want to implement it in this PR? Or merge as-is and open another JIRA?

jotak:
@stefannegrea , I'll update the PR. It also means adding `fromEarliest` to the "get" stats endpoints for multiple queries, it wasn't there, too.

stefannegrea:
@jotak, let's finalize this PR as-is and then do the additional work for `fromEarliest` in the context of [HWKMETRICS-445](https://issues.jboss.org/browse/HWKMETRICS-445).

stefannegrea:
Thank you @jotak!

PR #586, opened 2016-08-24
[HWKMETRICS-436] Create AvailabilityType key serializer (to be used w…

stefannegrea:
…ith maps when AvailabilityType is the key) and add it to the list of custom serializers.


jshaughn:
LGTM

PR #590, opened 2016-08-30
[HWKMETRICS-455] Removed unused schema tables and columns from tables.

stefannegrea:


FilipB:
retest this please

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.19.0.groovy
```diff
@@ -31,4 +31,40 @@ schemaChange {
...
   cql "ALTER TABLE data WITH compaction = {'class': 'DateTieredCompactionStrategy'} AND default_time_to_live = 604800"
-}
\ No newline at end of file
+}
+
+schemaChange {
+  version '3.2'
+  author 'snegre'
+  tags '0.19.x'
+  cql "ALTER TABLE data DROP aggregates"
```
burmanm: Why are we dropping these? With compression at least we could use these (to get some aggregates of blocks of data)

stefannegrea: We can add later anything we need to the schema. Right now these are unused for the original purpose and the PR from @jsanda for aggregates will not use them either.

jsanda: There is another point worth noting. The `aggregates` column has a UDT data type. UDTs are somewhat similar to other features in Cassandra like secondary indexes. They are talked up and marketed, but in practice their use is discouraged. It would be more efficient in terms of storage overhead to store all of that same info as a blob or even as individual columns (at least in Cassandra 3.x).

PR #591, opened 2016-08-31
HWKMETRICS-469 Add missing ISPN config at hawkular-component level

lucasponce:
Moving InsertedDataSubscriber into hawkular-component to inject Configuration properties

This is a following PR to fix an ISPN missing configuration.
In the previous PR, ISPN was config at tested level, and it was missing in hawkular-component artifact.
Also, there is a side effect of the current Configuration CDI mechanism, when trying to inject a property in a inner .jar the CDI gives some problems with the dynamic qualifiers.
A workaround is just to move the InsertedDataSubscriber class at hawkular-component level, as similar as NewDataListener.
I guess is an acceptable workaround instead to refactor the configuration properties.

I have tested manually with hawkular-services as we don't have yet tests at that level.
In a following PR I will update hawkular-services with ISPN changes and additional itests.
Please @stefannegrea @jsanda could you review ?
Thanks


lucasponce:
Fixed the problem with travis.
There was some missing configuration in one of the pom.xml related to CI.

stefannegrea:
Thank you @lucasponce!

PR #595, opened 2016-09-05
ETL compression process based on Gorilla algorithms

burmanm:
Adds ETL based compression methods in the core for Gauge data (for now, I will later in the PR add support for counters & availabilities also).

After this PR is merged, I recommend replacing Deflate with LZ4 again (since we need speed due to the ETL processing)

This PR for now is missing the job scheduler stuff, that is coming later to this same PR.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
@@ -26,10 +26,13 @@ import static org.hawkular.metrics.model.MetricType.GAUGE;
 import static org.hawkular.metrics.model.MetricType.STRING;
 
 import java.nio.ByteBuffer;
+import java.util.Date;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 
+import org.hawkular.metrics.core.service.compress.CompressedPointContainer;
 import org.hawkular.metrics.core.service.transformers.BatchStatementTransformer;
 import org.hawkular.metrics.model.AvailabilityType;
 import org.hawkular.metrics.model.DataPoint;
@@ -78,11 +81,21 @@ public class DataAccessImpl implements DataAccess {
 
     private PreparedStatement findMetricInData;
 
+    private PreparedStatement findMetricInDataCompressed;
+
+    private PreparedStatement findAllMetricsInData;
+
+    private PreparedStatement findAllMetricsInDataCompressed;
+
     private PreparedStatement findMetricInMetricsIndex;
 
     private PreparedStatement getMetricTags;
 
     private PreparedStatement insertGaugeData;
+
+    private PreparedStatement insertCompressedData;
+
+    private PreparedStatement insertCompressedDataWithTags;
 
     private PreparedStatement insertGaugeDataUsingTTL;
 
@@ -105,6 +118,14 @@ public class DataAccessImpl implements DataAccess {
     private PreparedStatement insertStringDataWithTags;
 
     private PreparedStatement insertStringDataWithTagsUsingTTL;
+
+    private PreparedStatement findCompressedDataByDateRangeExclusive;
+
+    private PreparedStatement findCompressedDataByDateRangeExclusiveWithLimit;
+
+    private PreparedStatement findCompressedDataByDateRangeExclusiveASC;
+
+    private PreparedStatement findCompressedDataByDateRangeExclusiveWithLimitASC;
 
     private PreparedStatement findCounterDataExclusive;
 
@@ -134,6 +155,8 @@ public class DataAccessImpl implements DataAccess {
 
     private PreparedStatement deleteGaugeMetric;
 
+    private PreparedStatement deleteDatapoints;
+
     private PreparedStatement insertAvailability;
 
     private PreparedStatement insertAvailabilityUsingTTL;
@@ -157,8 +180,6 @@ public class DataAccessImpl implements DataAccess {
     private PreparedStatement deleteTagsFromMetricsIndex;
 
     private PreparedStatement readMetricsIndex;
-
-    private PreparedStatement findAllMetricsInData;
 
     private PreparedStatement updateRetentionsIndex;
 
@@ -192,10 +213,14 @@ public class DataAccessImpl implements DataAccess {
         findTenant = session.prepare("SELECT id, retentions FROM tenants WHERE id = ?");
 
         findMetricInData = session.prepare(
-            "SELECT DISTINCT tenant_id, metric " +
+            "SELECT DISTINCT tenant_id, type, metric, dpart " +
             "FROM data " +
-            "WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart=? " +
-            "LIMIT 1");
+            "WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = ? ");
+
+        findMetricInDataCompressed = session.prepare(
+                "SELECT DISTINCT tenant_id, type, metric, dpart " +
+                        "FROM data_compressed " +
+                        "WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = ? ");
 
         findMetricInMetricsIndex = session.prepare(
             "SELECT metric, tags, data_retention " +
@@ -236,8 +261,24 @@ public class DataAccessImpl implements DataAccess {
             "ORDER BY metric ASC");
 
         findAllMetricsInData = session.prepare(
-            "SELECT DISTINCT tenant_id, metric, type, dpart " +
+            "SELECT DISTINCT tenant_id, type, metric, dpart " +
             "FROM data");
+
+        findAllMetricsInDataCompressed = session.prepare(
+                "SELECT DISTINCT tenant_id, type, metric, dpart " +
+                        "FROM data_compressed");
+
+        insertCompressedData = session.prepare(
+                "UPDATE data_compressed " +
+                        "USING TTL ? " +
+                        "SET c_value = ? " +
+                        "WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = ? AND time = ? ");
+
+        insertCompressedDataWithTags = session.prepare(
+                "UPDATE data_compressed " +
+                        "USING TTL ? " +
+                        "SET c_value = ?, tags = ? " +
+                        "WHERE tenant_id = ? AND type = ? AND metric = ? AND dpart = ? AND time = ? ");
 
         insertGaugeData = session.prepare(
             "UPDATE data " +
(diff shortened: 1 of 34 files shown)


burmanm:
retest this please

burmanm:
retest this please

burmanm:
Ready for review

burmanm:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -78,12 +78,16 @@
 
     private PreparedStatement findMetricInData;
 
+    private PreparedStatement findAllMetricsInData;
```
stefannegrea: Really not sure what this is doing here ...

stefannegrea: Actually I see now, it's just moved up.

stefannegrea:
Why did you pick the compression window to be 2 hours?

Can we make the compression job optional? With it being opt-in for 0.20.0 and opt-out for 0.21.0?

jsanda:
What is the purpose of the data locality stuff? I saw JAVA-1281, and understand the concept, but there is no guarantees about hawkular-metrics nodes being co-located with Cassandra nodes. If anything, I would assume that they won't be co-located. Is this something we need/want? I fear it introduces complexity (i.e., maintenance, bugs, etc.) without much benefit.

jsanda:
> Why did you pick the compression window to be 2 hours?

I want to say that the algorithm work better with 2 hours blocks, but @burmanm will have to confirm that. 

> Can we make the compression job optional? With it being opt-in for 0.20.0 and opt-out for 0.21.0?

If the pros far outweigh the cons as it seems thus far, I am not sure why we would want it disabled. With that said if we do decide to make it optional, I prefer that we push that off to a later PR because it introduces a lot more complexity. I would rather focus on this PR and then lots of testing to flush out any bugs and make sure it is stable.

jsanda:
I have some comments on `MetricsServiceImpl.compressBlock`. For the deletes I notice that we are doing a separate delete per data point (i.e., per cql row). At some point in 2.x support was added for range deletions. That would be much more efficient than a separate delete per data point which seems really heavyweight. I have another thought on deletes that I will come back to momentarily.

In `DataAccessImpl.deleteAndInsertCompressedGauge` should we use concatWith instead of mergeWith? I realize that concatWith is probably going to be a little slower, but then we can guarantee that the compressed data is successfully inserted before deleting anything.

Should we go ahead implement date bucketing/partitioning in support of this. Partition sizes would match the compression window/block size.  Before this work came about, I know we were thinking about a default bucket size of one day. But if we do two hours, we only have to perform one partition delete after the compressed data is stored, and we get an even better distribution of data.  We also would need to think about how big buckets should be for compressed data.

jsanda:
Another thought related to deletes. If we going to rely on the compression job to purge raw data, should we bother with setting the TTL on the raw? 

And more importantly I noticed that there is no TTL set on compressed data when inserted. How are we planning on purging it?

jsanda:
Instinctively I have reservations about storing the compressed data in the `data` table. I think it would be worthwhile to do some performance analysis on this schema vs storing compressed data in separate table.

burmanm:
@jsanda The data locality investigation was because this is going to cause a lot of I/O when we're doing it. If the data is always compressed in a remote node, it will have to transfer all the data through the network. But, at this point there's no way to trigger that behavior. I can move it for future PR / discussion (I'll comment on the TTL etc on next comment)

As for compressBlock, it actually does deletes using a batch, so the action gets executed at the same time (and based on my previous experience, it's much faster this way). With ranged delete, we might be deleting a row that was inserted while we're compressing (but was not read), but with this there's no chance of losing a datapoint. These are the decisions again how much data loss we accept.

mergeWith/concatWith has no effect in this case, I can change it either way.

@stefannegrea 2 hour block was decided so that we gain somewhat larger blocks because we don't create new rows all the time, but even larger blocks would cause us to store the raw data for too long. So this is just a compromise and can be adjusted later.

burmanm:
@jsanda I have thought about the TTL replacement just like you wrote with a two hour window, dropping a partition and then writing a single row to it (shame Cassandra doesn't support merge-operators in compaction like RocksDB). We would also save at least some percentage of space in the raw-data storage. 

The TTL is just missing from the insert, I can fix that in the short term.

As for storing raw-data in different table vs. compressed. This is trickier because of our current way of preparing statements and Cassandra. What it would in theory mean is replicating all the queries in DataAccessImpl that read the data-table. That's going to be a nightmare to maintain unless we come up with something else to replace the current scheme (like autogenerating the prepared statements instead of manually coding them like now). This is why I removed the original plan of using different table. For performance in the long run, that depends on the compactors (and how the TimeWindow behaves on our usage)

jsanda:
> As for compressBlock, it actually does deletes using a batch...

I see that now; however, we are doing read-before-write which is an anti-pattern, which we should avoid. If we store raw data in buckets with the same size as the compression block, then we can eliminate the reads. This could and probably should be done in a follow up ticket.

jsanda:
We need some performance analysis to determine wether or not a separate table makes sense. *If* performance is better with a separate table, having to write the additional prepared statements is a small price to pay. I do agree though that auto-generating statements would be nice. The driver has an object mapper library, but I fear we would trade performance for convenience.

jsanda:
In `MetricsServiceImpl.findDataPoints(MetricId<T> metricId, long start, long end, int limit, Order order)` I see the logic that determines whether to query raw or compressed data. Maybe I am missing something, but it does not look like it covers the scenario in which you have to query both compressed and raw data. If the start time is more than two hours ago and if the end time is now, wouldn't we need to query both compressed and raw?

jsanda:
@burmanm can you explain the use of BufferUntil. Admittedly, I don't understand very well some of the versions of the buffer and window operators, but would one of them provide the necessary functionality?

burmanm:
@jsanda The `findDataPoints` uses the same fetchers as fetching just the uncompressed data, so both datapoints would be fetched from the Cassandra. In the `DataPointDecompressor` we detect if the datapoint is compressed or not and then proceed accordingly.

There's a test `addAndCompressData` that reads data from both compressed as well as uncompressed rows.

burmanm:
As for BufferUntil, there isn't a suitable method in the RxJava that would do the same. One option is to use .buffer with a closingSelector, but then that would require first duplicating the original Observable to create a separate instance that emits a new item with filter(c_value != null). This creates one unreadable mess with Observable caching preventing any sort of streaming behavior.

jsanda:
Why do we have to sort in DataPointDecompressTransformer? Aren't the data points in the compressed block already sorted?

jsanda:
The last big thing I want to address is the compression job, CompressData.java. Let's say the job starts running at 13:00:00. At 13:20:12 the server is shutdown. Assume that the job did not finish. When the server starts back up, the job scheduler will run the job again for the 13:00 execution. The job iterates over all metrics and calls `MetricsService.compressBlock` to perform the necessary work. In the restart scenario some work will get repeated, and on the one hand that is fine since the operations involved are idempotent. On the other hand as the number of metrics in the systems gets larger, this could be a substantial amount of work to repeat. We may want to introduce something to keep track of completed work and/or remaining work. I am planning to do something to handle this with for roll ups. Maybe this is something we can push down to the job scheduler to some degree.

burmanm:
As for recompressing on crash.. I'm not really worried about that, as the compression really is quite quick (especially if there's already compressed blocks, reading them is quick). Sure, it places some extra load on the Cassandra backend.

But actually, I'm not sure if it works right now (the crashing scenario) as we would try to delete the datapoints that wre inside the compressed block. Maybe in that case the range delete would be better..

jsanda:
> As for recompressing on crash.. I'm not really worried about that, as the compression really is quite quick

I don't doubt that it is quick. When there are tens of thousands (or more) metrics in the system, the job is going to take some time to complete. This is something we can always revisit if necessary.

jsanda:
Here is something else to consider about using a new table vs using the existing data table. With a new table we use something smaller than timeuuid for the time column. That could be a significant savings.

burmanm:
timeuuid vs. something else is just few bytes per 2 hour row. That's not going to make a huge difference ;)

jsanda:
> timeuuid vs. something else is just few bytes per 2 hour row. That's not going to make a huge difference ;)

timeuuid is 16 bytes. We could use timestamp which is 8 bytes. I think that is a pretty substantial difference. But we can do even better. If we go ahead utilize the `dpart` column by implement date partitioning the time column can be an offset which we could probably store with a a smallint which is 2 bytes. That is a savings of 14 bytes per row. That is pretty significant.

burmanm:
Well, lets say we have a datapoint every 10 seconds, that means a single row is going to have 720 datapoints. The savings are 14 bytes per 720 datapoints -> 1,5 bits per datapoint ;)

Only if we design the other table with different dpart would it make sense. So for example 1 day dparts with 2 hour cells for example, and in the data table we would use 2 hour partitions to drop them more easily after compression.

That would get us rid of TTL in both cases, saving disk space and making drop operations faster. That might make sense.

jsanda:
Let's consider a week of data since 7 days is the default retention in openshift. 7 days is 84 compressed, 2 hr rows. The timeuuid overhead for that is 84 * 16 = 1,344 bytes. The smallint overhead would be 84 * 2 = 168 bytes.  The difference is 1176 bytes.  If we have a 100k metrics then the additional overhead of timeuuid vs smallint is 117.6 MB. I agree that that isn't huge, but the  overhead only increases with more metrics.

It also increases quite a bit with a longer data retention. Suppose we use a retention of a month or 31 days. The overhead of timeuuid vs smallint is 5208 bytes per metric in this scenario. At 100k metrics that yields about 5.2 GB addition overhead.

jsanda:
> Only if we design the other table with different dpart would it make sense.

I was thinking about this a bit earlier today. I was thinking that we might use a default bucket size of a week, my ideas apply as well with a bucket size of a day. We do not need to store a full timestamp in dpart. If we are partitioning by day or week, then we don't care about hours, minutes, seconds, and milliseconds. More importantly, we can store in the sys_config table the day on which we start using the compressed data table. This would be done at installation/upgrade time. Then we only need to store an offset in dpart. We can do that with a smallint. Two bytes gives us plenty of room.

jsanda:
@burmanm see [the commit](https://github.com/jsanda/hawkular-metrics/commit/66de2ad774e62d54b0ba4300a6caed114590a5c4) I pushed for fixing the integration test. Don't do verification in the onJobFinished callback. Just use it for synchronization and do verification in the main thread. You will likely run into similar problems if you perform blocking operations/verification  in the onTimeSliceFinished callback. Blocking calls in the jobs itself though should be fine.

burmanm:
Had to squash the commits because of conflicts when rebasing from the master (which had relevant commits)

stefannegrea:
What happens when data is pushed backdated to a time that was already compressed?

burmanm:
At this point, it's not compressed. The out-of-order writes (if done >1 hour after the timestamp) stay uncompressed (and get deleted with TTL)

stefannegrea:
@burmanm, compression is one aspect, what about reading it out?

burmanm:
@stefannegrea The data is read from both sources, compressed and decompressed and then merged, so out-of-orders (more than one hour out-of-order) are just stored uncompressed but there's no difference in usage.

burmanm:
@jsanda So I compared two different approaches to the schema:

```
CREATE TABLE data_old (
    tenant_id text,
    type tinyint,
    metric text,
	dpart bigint,
    time timestamp,
    chunks blob,
    PRIMARY KEY ((tenant_id, type, metric, dpart), time)
) WITH CLUSTERING ORDER BY (time DESC)
```

vs. 

```
CREATE TABLE data_compressed (
    tenant_id text,
    type tinyint,
    metric text,
	dpart bigint,
    time smallint,
    chunks blob,
    PRIMARY KEY ((tenant_id, type, metric, dpart), time)
) WITH CLUSTERING ORDER BY (time DESC)
```

Both used one day dpart (stored as seconds since epoch) and in ``data_compressed`` I used minutes since the dpart. I inserted 84 rows per metric (7 days worth of data, 2 hour rows) for 10 000 metrics, with each row having 720 datapoints (one every 10 seconds). That is, 840 000 rows in total with 604800000 datapoints (605M) in total. Each datapoint was a value between 1 and 720 (+1 from each value) with second precision.

End results:

data_old -> 26MB of diskspace
data_compressed -> 24MB of diskspace

So there isn't a large difference between those two (LZ4 used for both tables).

stefannegrea:
@burmanm, one question regarding the merge you mention in the previous comment. Does the merge apply for the case where the compression is in progress for a block and the data is stored in the compressed table and but not yet deleted from raw data table? If a user queries that particular metric at that exact time, will the merge eliminate duplicates?

jsanda:
In data_old, were you using timeuuid or timestamp? I assume timeuuid but want to double check. I did a similar test but using fewer metrics and the results were about the same - 2 MB difference. I wonder what if any difference it would make if we change dpart to a smallint. There is something else I noticed during my last test run. There is a lot less flushing of the compress_data table compared to the data table. I wonder if this could help with performance.

jsanda:
Slight correction about my previous test. It was using timestamp not smallint. I need to update my branch to get the smallint changes. I'll do that, retest and report results.

burmanm:
@jsanda timestamp (as that's what I was going to use originally in the data_compressed). There's probably even less reasons to use smallint on dpart, in this case the difference would be something like 2MB / 12 (each partition has 12 rows).

burmanm:
@stefannegrea As we talked in the meeting, I would move that merge sort to another PR. While the situation is quite theoretical, we can filter out duplicates.

stefannegrea:
@burmanm, do you mind opening a JIRA against 0.21.0 for that filtering and merge?

burmanm:
@stefannegrea HWKMETRICS-488

burmanm:
Rebased once again (and enabled tags processing), this time without squashing the commits.

jsanda:
@burmanm we're all green! Are we ready to merge?

stefannegrea:
@burmanm, could you please rebase on master.

burmanm:
Squashed the commits and rebased once more..

PR #596, opened 2016-09-06
[HWKMETRICS-461] Fix the regexp split

burmanm:
The regexp was missing double quotation for the regexp character "|", that's why it returned incorrect information. Also the tests used the wrong endpoint and just happened to work.


lucasponce:
Just minor comment.
Can we add a test with several UUIDs ? To check that the "|" pattern works correctly ?
Also, I have one fear, as the real metric ids used has the format:
MI~R~[71daaa4b-da76-4373-8753-68279f33a884/Local~~]~MT~WildFly Memory Metrics~Accumulated GC Duration
I wonder if we can have additional problems with the regexp
I guess not, but if we could set up a test with a couple of these IDs too, I'm sure it will save us time in the future.
Thanks for the fix.

stefannegrea:
Thank you @burmanm!

PR #597, opened 2016-09-06
[HWKMETRICS-472] Implement aggregated availability

jotak:
JIRA: https://issues.jboss.org/browse/HWKMETRICS-472

This PR follows discussion on hawkular-dev (mail title: "Availability metrics: aggregate stats series")

Remark about assertj:
I added assertj, an assertion framework, to the class path in test scope. I've started a topic on hawkular-dev to discuss about assertj. If necessary I will update this PR to remove assertj.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
@@ -25,8 +25,10 @@ import static org.hawkular.metrics.api.jaxrs.util.ApiUtils.serverError;
 import static org.hawkular.metrics.model.MetricType.AVAILABILITY;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.PatternSyntaxException;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -46,11 +48,13 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
+import org.hawkular.metrics.api.jaxrs.AggregationQueryRequest;
 import org.hawkular.metrics.api.jaxrs.QueryRequest;
 import org.hawkular.metrics.api.jaxrs.handler.observer.MetricCreatedObserver;
 import org.hawkular.metrics.api.jaxrs.handler.observer.ResultSetObserver;
 import org.hawkular.metrics.api.jaxrs.handler.template.IMetricsHandler;
 import org.hawkular.metrics.api.jaxrs.handler.transformer.MinMaxTimestampTransformer;
+import org.hawkular.metrics.api.jaxrs.param.TagsConverter;
 import org.hawkular.metrics.api.jaxrs.util.ApiUtils;
 import org.hawkular.metrics.core.service.Functions;
 import org.hawkular.metrics.core.service.Order;
@@ -62,6 +66,7 @@ import org.hawkular.metrics.model.DataPoint;
 import org.hawkular.metrics.model.Metric;
 import org.hawkular.metrics.model.MetricId;
 import org.hawkular.metrics.model.MetricType;
+import org.hawkular.metrics.model.RatioMap;
 import org.hawkular.metrics.model.param.BucketConfig;
 import org.hawkular.metrics.model.param.Duration;
 import org.hawkular.metrics.model.param.TagNames;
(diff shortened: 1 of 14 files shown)


rhqci:

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -910,6 +915,38 @@ public void setDefaultTTL(int defaultTTL) {
...
+        checkArgument(isValidTimeRange(start, end), "Invalid time range");
+        Observable<DataPoint<RatioMap>> result = Observable.from(metrics)
+                .map(id -> new MetricId<>(tenantId, MetricType.AVAILABILITY, id))
+                .flatMap(this::findMetric)
+                .map(metric -> findDataPoints(metric.getMetricId(), start, end, limit, order))
+                .compose(o -> buildRatioMapSeries(o, order, AvailabilityType::getText));
+        if (limit > 0) {
+            // Limit already applied on C* query, but might be exceeded after aggregation
+            // So it's applied again
+            result = result.limit(limit);
```
burmanm: .take(limit) is the default in RxJava, limit is an alias (but in ReactiveX world, the normal term is take, so lets use that)

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -910,6 +915,38 @@ public void setDefaultTTL(int defaultTTL) {
...
+        }
+        return result;
+    }
+
+    @Override
+    public Observable<DataPoint<RatioMap>> findAvailabilityRatioMap(
```
burmanm: This does not accept limiter functions like findMetrics?

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/MetricsServiceImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+                new DataPoint<>(5L, AvailabilityType.DOWN),
+                new DataPoint<>(10L, AvailabilityType.UP));
+        List<DataPoint<AvailabilityType>> serie2 = ImmutableList.of(
+                new DataPoint<>(0L, AvailabilityType.UP),
+                new DataPoint<>(7L, AvailabilityType.DOWN),
+                new DataPoint<>(10L, AvailabilityType.ADMIN));
+        List<DataPoint<AvailabilityType>> serie3 = ImmutableList.of(
+                new DataPoint<>(1L, AvailabilityType.UP),
+                new DataPoint<>(7L, AvailabilityType.DOWN));
+        Observable<Observable<DataPoint<AvailabilityType>>> observable = Observable.just(
```
burmanm: observable is terrible name..

burmanm: And just shouldn't be used here, instead use ``concat`` (for sequential) or ``merge`` (possible parallelism) to merge these observables

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/MetricsServiceImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+                Observable.from(serie1),
+                Observable.from(serie2),
+                Observable.from(serie3));
+        List<DataPoint<RatioMap>> result = service.buildRatioMapSeries(observable, Order.ASC, AvailabilityType::getText)
```
burmanm: You could replace this with:

```java
Observable.merge(.. series ..) // or concat, depends what you want
 .flatMap(..)
```

Thus keeping the flow visible. And actually, instead of result, you could use TestSubscriber and then watch the emitted items.

FilipB:
ok to test

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jotak: You mean like the regex match?

I've built it on the same model than the existing "raw" endpoint on avail, so for now there's no such thing. But it can be interesting, to search for several metrics based on a pattern. I'll have a look at it.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/MetricsServiceImplTest.java
jotak: It must not be merged/concat or flatmaped because the "buildRatioMapSeries" needs to work on each series individually then it proceed flat-mapping and reduction. The "Observable<Observable<DataPoint<AvailabilityType>>> " is the result of loading data points on several series as I expect. The unit test has as little as possible transformations to perform.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/MetricsServiceImplTest.java
jotak: variable renamed

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jotak: done

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/MetricsServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+                        new DataPoint<>(0L, AvailabilityType.UP),
+                        new DataPoint<>(7L, AvailabilityType.DOWN),
+                        new DataPoint<>(10L, AvailabilityType.ADMIN)),
+                Observable.just(
+                        new DataPoint<>(1L, AvailabilityType.UP),
+                        new DataPoint<>(7L, AvailabilityType.DOWN)));
+        List<DataPoint<RatioMap>> result = service.buildRatioMapSeries(allSeries, Order.ASC, AvailabilityType::getText)
+                .toList()
+                .toBlocking()
+                .single();
```
jotak: (Reminder: must try with TestSubscriber)

jotak: done

stefannegrea:
retest this please

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
```diff
@@ -466,4 +471,126 @@ public void getMetricStats(
                 .map(ApiUtils::collectionToResponse)
                 .subscribe(asyncResponse::resume, t -> asyncResponse.resume(serverError(t)));
     }
+
+    @GET
+    @Path("/aggregated")
```
stefannegrea: This path is not good. Aggregated will be used in a different context soon. Can you put this functionality under `/stats`?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
```diff
@@ -466,4 +471,126 @@ public void getMetricStats(
...
+            result = metricsService.findAvailabilityRatioMap(getTenant(), metrics, timeRange.getStart(),
+                    timeRange.getEnd(), limit, order);
+        }
+        result.toList()
+                .map(ApiUtils::collectionToResponse)
+                .subscribe(asyncResponse::resume, t -> asyncResponse.resume(ApiUtils.serverError(t)));
+    }
+
+    @POST
+    @Path("/aggregated/query")
```
stefannegrea: Same comment as above about the usage of `/aggregated`; `stats` is better.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
jotak: I'm not sure if "stats" fits well. For now "stats" is about time-aggregation, with buckets and so on, whereas here it's about multiple series aggregation, without any time bucket. I fear it would be confusing if we use the same word.

On hawkular-dev ML Jay proposed: "/aggregate, /set, /group or whatever" ... is there anyone of these you find better? Or maybe something like "/merged" ?

stefannegrea:
Just to clarify the comments regarding `/stats`, we need roll this feature inside stats end-point and provide the ratio map as an optional function the users can select.

PR #599, opened 2016-09-07
[HWKMETRICS-474] Update the CQL query to use the correct String metri…

stefannegrea:
…c type when querying with ASC order and limit.


mwringe:
LGTM, will merge once the tests have passed

PR #600, opened 2016-09-08
HAWKULAR-1121 Store string as a key on PublishCache

lucasponce:
A bug in an integration test on hawkular-services (https://issues.jboss.org/browse/HAWKULAR-1121) has masked a problem not detected until now.
Storing an object as a key for published cache requires to share same classloader between libraries used by metrics and alert.
So a workaround is to use a primitive type as proposed on this fix.
Counter part on alerting and fixed itest on hawkular-services are on the way.
/cc @stefannegrea @jshaughn

### DIFF
--- a/bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
+++ b/bus/src/main/java/org/hawkular/metrics/component/publish/PublishCommandTable.java
@@ -25,6 +25,7 @@ import javax.enterprise.context.ApplicationScoped;
 import org.hawkular.metrics.api.jaxrs.util.Eager;
 import org.hawkular.metrics.model.MetricId;
 import org.infinispan.Cache;
+import org.jboss.logging.Logger;
 
 /**
  * A table to store a cache of published metrics on the bus under hawkular-services context.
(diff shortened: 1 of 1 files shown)


jshaughn:
I can't quite understand what a change from MetricId key to String solves the problem.  String is a class just like MetricId, it's not a Java primitive.   So, I'm not sure the classloader analysis is correct.  The equals and hascode impls of metricId look OK.  But, you have shown that it works with String key and not MetricId key, so I guess I'm OK with the change if necessary.

lucasponce:
I guess point is that MetricId is not serializable
http://infinispan.org/docs/stable/user_guide/user_guide.html#_plugging_infinispan_with_user_defined_externalizers
Perhaps, a workaround is simply to make it serializable.
It should work.
Please don't merge this one, I'm going to test that.

lucasponce:
Implemented Serializable on MetricId.
I hope travis CI can confirm that this change has not any regression on any perf task.

lucasponce:
Finally storing a string as a key for PublishCache.
Confirmed that using MetricId on local ISPN cache from different isolated apps it loads the MetricId class from different classloaders, so local cache fails to validates keys in this scenario.
Other alternatives like moving local cache into a distributed one to force serialization of objects are costly than a string conversion.

stefannegrea:
@lucasponce, do you mind opening a Hawkular Metrics JIRA for this issue too? Also, once this gets merged in master, do you mind opening a second PR agains the 0.19.0 release branch?

jshaughn:
Fixes https://issues.jboss.org/browse/HWKMETRICS-475

jshaughn:
LGTM.

stefannegrea:
Thank you @lucasponce and @jshaughn!

PR #602, opened 2016-09-12
[HWKMETRICS-476] update repeating trigger api and add tests

jsanda:


jsanda:
@burmanm I've added a builder method for specifying the trigger time. I am not sure though that this is any different than just specifying an initial delay. Let me know what you think.

burmanm:
While it's not a full cron timer, at least it allows something like "run every 00:00" (start time 00:00, repeat every 24 hours).

job-scheduler/src/test/java/org/hawkular/metrics/scheduler/api/RepeatingTriggerTest.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+package org.hawkular.metrics.scheduler.api;
+
+import static org.hawkular.metrics.datetime.DateTimeService.currentMinute;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+import java.util.concurrent.TimeUnit;
+
+import org.testng.annotations.Test;
```
burmanm: Any chance to avoid using TestNG? JUnit @Test would be better ;)

PR #603, opened 2016-09-13
[HWKMETRICS-478] Protect /tenants endpoint with a secret admin token.

stefannegrea:
The token can be set or reset at startup via configuration property.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -21,6 +21,7 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
+import static org.hawkular.metrics.api.jaxrs.config.ConfigurationKey.ADMIN_TOKEN;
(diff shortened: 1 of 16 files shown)


burmanm:
@mwringe I'll need your input on this, as we'll need that /tenants endpoint in Heapster integration. So the token must be available to Heapster. And anything of course that requires fetching tenant information, such as Grafana integration..

mwringe:
@stefannegrea does this affect standalone Hawkular Metrics? How does securing one endpoint really help things if everything else is wide open with respect to security?

stefannegrea:
@burmanm, I would not use this token feature for Heapster, the TenantFilter should not be deployed on anywhere but Hawkular Metrics standalone.

stefannegrea:
@mwringe, this is just the first PR in a series of PRs to add management features for Hawkular Metrics.

mwringe:
@stefannegrea if this is going to be applied on top of a standalone Hawkular Metrics then it needs to be done as an optional filter so that we can continue to use the OpenShift filters without issues.

Or if we want to do this in a more proper fashion, we need to have a plugable security framework in place where we can use this implementation, the OpenShift one, or some other one.

stefannegrea:
Based on the discussion with @mwringe, I will need to rework the tenant filter changes because it is used as-is by OpenShift at this time.

mwringe:
With moving this to a separate filter, and having it disabled in the OpenShift case, I think this is better approach to take. LGTM

stefannegrea:
retest this please

jsanda:
> this is just the first PR in a series of PRs to add management features for Hawkular Metrics.

Do we have documented anywhere what we are planning to do next/long term in this area?

stefannegrea:
@jsanda, we have the design document for the Management Framework. This is the first step in that area.

PR #605, opened 2016-09-16
Metrics w/ Alerting

jshaughn:
This is for Review Only at this point and Goes hand in hand with PRs for Hservices and Halerting.


jshaughn:
Not that anyone would, but...

DO-NOT-MERGE

alerting/src/main/java/org/hawkular/metrics/alerting/PublishCommandTable.java
```diff
@@ -0,0 +1,116 @@
...
+
+    // This just prevents us from having to create a new object on every cache lookup...
+    // Note, not thread safe, lookups must be atomic
+    Map<MetricType<?>, CacheKey> map = new HashMap<>();
+
+    @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/publish")
+    private Cache<CacheKey, String> publishCache;
+
+    public boolean isPublished(MetricId<?> id) {
+        return publishCache.containsKey(getCacheKey(id));
```
lucasponce: Commented on alerts PR but leaving a note here.
Using domain objects in a shared ISPN requires to share classloader.
I think this is adressed in the packaging format.

dist/component-ear/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.hawkular.metrics</groupId>
+      <artifactId>hawkular-metrics-component</artifactId>
+      <version>${project.version}</version>
+      <type>war</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.alerts</groupId>
+      <artifactId>hawkular-alerts-rest</artifactId>
```
lucasponce: Just a comment: we need to offer in alerting two artifacts.
A .war distribution for Hawkular Services scenarios and another .war for just metrics-alerting on standard wildfly distribution.
Today this is done via a conditional maven profile and in a next step we should provide boths artifacts at same time to use them as a dependency like this.

dist/component-ear/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <webModule>
+              <groupId>org.hawkular.alerts</groupId>
+              <artifactId>hawkular-alerts-rest</artifactId>
+              <bundleFileName>hawkular-alerts.war</bundleFileName>
+              <contextRoot>/hawkular/alerts</contextRoot>
+            </webModule>
+            <webModule>
+              <groupId>org.hawkular.alerts</groupId>
+              <artifactId>hawkular-alerter-metrics</artifactId>
+              <contextRoot>/hawkular/__alerter-metrics</contextRoot>
```
lucasponce: I am not sure if I like this url format.
I understand that alerter is headless and having a context root is mandatory but not used.
And I guess that pattern is to warn that this is not used, but I guess we can avoid the "__" chars.

dist/component-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -37,7 +37,7 @@
 
   <resource-env-ref>
     <resource-env-ref-name>cache/publish</resource-env-ref-name>
-    <lookup-name>java:jboss/infinispan/cache/hawkular-metrics/publish</lookup-name>
+    <lookup-name>java:jboss/infinispan/cache/hawkular-alerts/publish</lookup-name>
```
lucasponce: This change needs to be propagated in alerting itest as well.

alerting/src/main/java/org/hawkular/metrics/alerting/PublishCommandTable.java
```diff
@@ -0,0 +1,116 @@
...
+@Eager
+public class PublishCommandTable {
+
+    private static final String AVAILABILITY = "hm_a_";
+    private static final String COUNTER = "hm_c_";
+    private static final String COUNTER_RATE = "hm_cr_";
+    private static final String GAUGE = "hm_g_";
+    private static final String GAUGE_RATE = "hm_gr_";
+    private static final String STRING = "hm_s_";
+    private static final String UNDEFINED = "hm_u_";
```
lucasponce: In which use cases it is needed an UNDEFINED prefix ?
I am not sure if I see it.

dist/component-war/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
```diff
@@ -74,6 +75,8 @@
 import rx.Observable;
 
 /**
+ * TODO: We need to mock out AlertsService, I think, so that InsertedDataSubscriber can instantiate correctly
```
lucasponce: You can add a Custom implementation of AlertsService on the test folder and add it manually as a class.

dist/component-war/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
```diff
@@ -130,6 +137,7 @@ public static void setupPublishing() {
         System.setProperty("hawkular-metrics.publish-period", "100");
     }
 
+    @Ignore
```
lucasponce: I guess this test will be fixed.
The main scenario is to validate the filtering use case, having data points for published and non published metrics and checking that only the published are received in the old JMS.
In the new scenario, a mock AlertsServices as noted in the commented code just has to validate that points are send.

dist/component-war/src/test/resources/web.xml
```diff
@@ -23,7 +23,7 @@
 
   <resource-env-ref>
     <resource-env-ref-name>cache/publish</resource-env-ref-name>
-    <lookup-name>java:jboss/infinispan/cache/hawkular-metrics/publish</lookup-name>
+    <lookup-name>java:jboss/infinispan/cache/hawkular-alerts/publish</lookup-name>
```
lucasponce: Commented in a similar entry.

dist/standalone-alerting-war/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+
+  <name>Hawkular Metrics Alerting WAR for Wildfly</name>
+  <description>Hawkular Metrics Alerting WAR for WildFly</description>
+
+  <dependencies>
+    <!-- Hawkular Alerting dependencies -->
+    <!-- Overlay -->
+    <dependency>
+      <groupId>org.hawkular.alerts</groupId>
+      <artifactId>hawkular-alerts-rest</artifactId>
```
lucasponce: Commented in a previous entry.

dist/standalone-ear/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <webModule>
+              <groupId>org.hawkular.metrics</groupId>
+              <artifactId>hawkular-metrics-standalone-alerting</artifactId>
+              <bundleFileName>hawkular-alerts.war</bundleFileName>
+              <contextRoot>/hawkular/alerts</contextRoot>
+            </webModule>
+            <webModule>
+              <groupId>org.hawkular.alerts</groupId>
+              <artifactId>hawkular-alerter-metrics</artifactId>
+              <contextRoot>/hawkular/__alerter-metrics</contextRoot>
```
lucasponce: Commented in a previous entry.

alerting/src/main/java/org/hawkular/metrics/alerting/PublishCommandTable.java
jshaughn: Right, as mentioned in other PR, this works because of the EAR deployment.

dist/component-ear/pom.xml
jshaughn: Right.  The way I got around this was to use a maven overlay of alerting's published "bus" war artifact, and remove the offensive bus-related jars.  This is actually not too bad, but certainly if alerting published both kinds of artifacts it would be unnecessary.

dist/component-ear/pom.xml
jshaughn: I am totally open to suggestions here and will replace this with anything you like.

dist/component-war/src/main/webapp/WEB-INF/web.xml
jshaughn: I thought I made the change everywhere the cache was used but I will look again.

alerting/src/main/java/org/hawkular/metrics/alerting/PublishCommandTable.java
jshaughn: I added UNDEFINED simply for completeness, because it is a defined metrics type.

dist/component-war/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
jshaughn: Good idea, I will try that.

dist/component-war/src/test/java/org/hawkular/bus/PublishDataPointsTest.java
jshaughn: Yes, this test needs to be fixed up, I will try and do that soon.

jshaughn:
@stefannegrea , @jsanda , @lucasponce The alerting PR has been merged and with the change to an alerting src-dep version this PR has gone green.   I think it is time for any final PR review comments you may have before we prepare to merge this into metrics master.

Note that Lucas is doing some work right now on alerting that we want to have ready before this is merged.  So, the src-dep will likely be updated before this is merged (or possibly we will release a proper alerting version).  But it should not have much affect on this PR logic.

So, please have any review comments in as soon as you can, hopefully by Tuesday.

jshaughn:
@stefannegrea, @jsanda  Assuming travis passes I think we can remove the DoNotMerge label.  I have updated the PR to depend on the new 1.2.0.Final release of Alerting.  I have tested the new ear in Hservices and I have done a quick check of a standalone deploy (just an error-free deploy and a few simple rest endpoints).

Although noted in a few other places, a reminder that it is expected that the ear be deployed as hawkular-metrics.ear.  Hservices does this automatically.  But in standalone it may be a manual deployment.  This naming is required for proper dependency and jndi resolutions.

stefannegrea:
Thank you @jshaughn!

PR #606, opened 2016-09-19
[HWKMETRICS-479] initialze rx test scheduler in constructor

jsanda:


burmanm:
This will not shutdown the tests correctly (they wait forever)

PR #609, opened 2016-09-23
Fix getTTL to calculate the seconds for insert processes

burmanm:
Currently setting TTL to anything else than default will cause the TTL to be seconds instead of days we advertise in the API docs..


burmanm:
HWKMETRICS-484

stefannegrea:
Thank you @burmanm!

PR #610, opened 2016-09-23
[HWKMETRICS-477] reschedule job for immediate execution if it falls behind

jsanda:
If a job runs long and misses its next execution time, the scheduler will now
execute the job again immediately for the missed execution. It will do so until
the job is caught up.

### DIFF
--- a/job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
+++ b/job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
@@ -183,7 +183,7 @@ public class SchedulerImpl implements Scheduler {
 
     // TODO make configurable
     private int getQueryThreadPoolSize() {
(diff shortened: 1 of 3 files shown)


job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -165,6 +181,11 @@ public SchedulerImpl(RxSession session) {
         jobFinished = Optional.empty();
     }
 
+    // TODO make configurable
+    private int getQueryThreadPoolSize() {
+        return Runtime.getRuntime().availableProcessors() / 2;
```
jotak: It could return 0, but should be 1 at least... no ?

jsanda:
retest this please

jsanda:
retest this please

stefannegrea:
Thank you @jsanda!

PR #613, opened 2016-09-28
[HWKMETRICS-486] use exclusive lock for job execution

jsanda:
This commit also increases the timeouts in JobExecutionTest.executeLotsOfJobs
because of intermittent failures on travis. On my box the test only executes
8 jobs, but on travis it runs 64 jobs. I don't think the 30 second timeout
used is always enough time.

### DIFF
--- a/job-scheduler/pom.xml
+++ b/job-scheduler/pom.xml
@@ -31,6 +31,10 @@
 
   <name>Hawkular Metrics Job Scheduler</name>
 
+  <properties>
+    <test.keyspace>hawkulartest</test.keyspace>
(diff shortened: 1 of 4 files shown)


burmanm:
Twist of irony that it failed on the same issue it was supposed to fix? ;)

"java.lang.AssertionError: expected [64] but found [70]"

jsanda:
Yes a twist of iron that had me up late last night and frustrated :) But at least this time I was actually able to get it to fail locally, so some progress.

stefannegrea:
Thank you @jsanda!

PR #617, opened 2016-09-29
HWKALERTS-169 Global front-line filtering

jshaughn:
HWKALERTS-169 Global front-line filtering
- alerting now responsible for filtering so just generate proper dataIds and send directly to alerting.
- remove publish cache decls and PublishCommandTable

These changes result from a change in approach regarding alerting data filtering.  The work is targeted for Metrics 0.20.0.

@stefannegrea DoNotMerge, must be reviewed and also updated with non-SNAPSHOT alerting version.
@lucasponce Please review.  Note that I will work on an i-test in a separate PR.


alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -129,10 +132,33 @@ private void publishNumeric(Metric<? extends Number> numeric) {
         }
     }
 
+    private String getDataId(MetricId<? extends Number> numericId) {
```
burmanm: Wouldn't an enum make more sense here? Or a map created in the beginning of type <MetricId, String> ?

pom.xml
```diff
@@ -97,9 +97,9 @@
     <test.keyspace>hawkulartest</test.keyspace>
     <nodes>127.0.0.1</nodes>
     <!-- Dependencies versions -->
-    <version.org.cassalog>0.3.0</version.org.cassalog>
+    <version.org.cassalog>0.3.1</version.org.cassalog>
     <version.org.hawkular.accounts>2.0.26.Final</version.org.hawkular.accounts>
-    <version.org.hawkular.alerts>1.2.0.Final</version.org.hawkular.alerts>
+    <version.org.hawkular.alerts>1.2.1.Final-SNAPSHOT</version.org.hawkular.alerts>
```
burmanm: As I started a review, I can keep it locked until this is unsnapshotted ;)

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -56,15 +55,20 @@
 public class InsertedDataSubscriber {
     private static final Logger log = Logger.getLogger(InsertedDataSubscriber.class);
 
+    static final String PREFIX_AVAILABILITY = "hm_a_";
```
lucasponce: These prefixes should be javadoc-ed or documented somewhere as they should be known by user.

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -91,7 +95,6 @@ public void onMetricsServiceReady(@Observes @ServiceReady ServiceReadyEvent even
             } else {
                 events = event.getInsertedData()
                         .filter(m -> m.getType() != MetricType.STRING)
-                        .filter(m -> publish.isPublished(m.getMetricId()))
                         .buffer(publishPeriod, TimeUnit.MILLISECONDS, 100)
```
lucasponce: It's not part of the PR. It's a side comment.
But I think that this buffer rate should be parametrized.
It will be very useful on performance tests to evaluate the good threshold for this.

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -129,10 +132,33 @@ private void publishNumeric(Metric<? extends Number> numeric) {
...
+        if (MetricType.COUNTER == metricType) {
+            return PREFIX_COUNTER + numericId.getName();
+        }
+        if (MetricType.GAUGE_RATE == metricType) {
+            return PREFIX_GAUGE_RATE + numericId.getName();
+        }
+        if (MetricType.COUNTER_RATE == metricType) {
+            return PREFIX_COUNTER_RATE + numericId.getName();
+        }
+        if (MetricType.STRING == metricType) {
```
lucasponce: Minor one:
MetricId<? extends Number> numericId works with MetricType.STRING ?

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -129,10 +132,33 @@ private void publishNumeric(Metric<? extends Number> numeric) {
...
+            return PREFIX_STRING + numericId.getName();
+        }
+        if (MetricType.UNDEFINED == metricType) {
```
lucasponce: Same comment, not sure if the parameter filters non numeric types.

alerting/src/test/java/org/hawkular/metrics/alerting/PublishDataPointsTest.java
```diff
@@ -62,7 +61,7 @@
 import rx.Observable;
 
 /**
- * TODO: We need to mock out AlertsService, I think, so that InsertedDataSubscriber can instantiate correctly
+ * Note that this test no longer tests filtering, as data is now filtered in alerting, not metrics.
```
lucasponce: Then, perhaps this test is purely outdated.
Main goal was to test the bus reception of datapoints and next we extended to filtering.
So, I guess this could be removed and added new ones (in a following PR) to test a whole lifecycle.
In Hawkular Services there are a couple of similar tests that can be re-used (I think).

integration-tests/rest-tests-jaxrs/pom.xml
```diff
@@ -203,6 +205,22 @@
                       <commmand>
                         /subsystem=logging/periodic-rotating-file-handler=FILE:write-attribute(name="file",value={path="${project.build.directory}/wildfly-test.log"})
                       </commmand>
+                      <!-- caches -->
+                      <commmand>
+                        /subsystem=infinispan/cache-container=hawkular-alerts/:add(module=org.jboss.as.clustering.infinispan,start=LAZY)
```
lucasponce: Why do you need to add the cache via CLI command ?
That sounds some descriptor is missing at app level.

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
jshaughn: could be done, if I need to change anything else in there I can do it to make you happy :)

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
jshaughn: I'm not sure where this is best documented.  I'll add something here just to explain why the prefixes exist and why we have a convention.  But it may need to be in user docs somewhere in metrics.

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
jshaughn: In a follow up I'd also like to allow us to do some batching here.

alerting/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
jshaughn: Not sure, I'll check this.

alerting/src/test/java/org/hawkular/metrics/alerting/PublishDataPointsTest.java
jshaughn: well, it minimally can just ensure we're sending with a valid prefix.  I'll leave it.  there will also be an itest for ear deployment and alert generation.  the alert generation may be a follow-up PR.

integration-tests/rest-tests-jaxrs/pom.xml
jshaughn: It works fine doing it this way.  It was not east to use the xsl transformer approach given the existing test setup in the pom.

jshaughn:
@stefannegrea , @burmanm , @lucasponce.  PR Feedback addressed, alerting updated to 1.2.1.Final, DNM label removed.  I think this is ready for merge.  This is targeted for 0.20.0 so please merge unless you see a true blocker.  I can address any non-blocker feedback in subsequent PRs.

stefannegrea:
@jshaughn, there is a failing test.

jshaughn:
It passes for me locally, more than once.  I'm not sure why it is failing on Travis.  I'll continue to look but it must have something to do with environment.  It's not a failure, but rather an error.

stefannegrea:
Thank you @jshaughn!

PR #623, opened 2016-10-04
[HWKMETRICS-490] make C* driver request and connection timeout settin…

jsanda:
…gs configurable


stefannegrea:
Thank you @jsanda!

PR #628, opened 2016-10-05
[HWKMETRICS-480] Ability to filter with non-existing tagName

burmanm:
Add the ability to filter metrics from tags without a certain tag being set, query format is !tagName:* (value is actually ignored, so it does not matter what is set)

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
@@ -89,6 +89,8 @@ public class DataAccessImpl implements DataAccess {
 
     private PreparedStatement findMetricInMetricsIndex;
 
+    private PreparedStatement findAllMetricsFromTagsIndex;
+
     private PreparedStatement getMetricTags;
 
     private PreparedStatement insertGaugeData;
@@ -267,6 +269,10 @@ public class DataAccessImpl implements DataAccess {
         findAllMetricsInDataCompressed = session.prepare(
                 "SELECT DISTINCT tenant_id, type, metric, dpart " +
                         "FROM data_compressed");
+
+        findAllMetricsFromTagsIndex = session.prepare(
+                "SELECT tenant_id, type, metric " +
+                        "FROM metrics_tags_idx");
 
         insertCompressedData = session.prepare(
                 "UPDATE data_compressed " +
@@ -500,7 +506,7 @@ public class DataAccessImpl implements DataAccess {
             "WHERE tenant_id = ? AND tname = ? AND tvalue = ? AND type = ? AND metric = ?");
 
         findMetricsByTagName = session.prepare(
-            "SELECT type, metric, tvalue " +
+            "SELECT tenant_id, type, metric, tvalue " +
             "FROM metrics_tags_idx " +
             "WHERE tenant_id = ? AND tname = ?");
 
@@ -994,5 +1000,10 @@ public class DataAccessImpl implements DataAccess {
                 .setUUID(5, getTimeUUID(sliceEnd)))
                 .concatWith(Observable.just(b))
                 .concatMap(st -> rxSession.execute(st));
+    }
+
+    @Override
+    public Observable<Row> findAllMetricsFromTagsIndex() {
+        return rxSession.executeAndFetch(findAllMetricsFromTagsIndex.bind());
     }
 }
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
@@ -1,0 +1,243 @@
+/*
+ * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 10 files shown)


burmanm:
It appears Travis will never succeed with running the test suite, ".NoHostAvailableException: All host(s) tried for query failed (tried: /127.0.0.1:9042 (com.datastax.driver.core.exceptions.ConnectionException: [/127.0.0.1] Write attempt on defunct connection" for .. many many times in a row now.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
```diff
@@ -0,0 +1,243 @@
...
+            Map<Long, List<Map.Entry<String, String>>> costSortedMap = new TreeMap<>();
+            costSortedMap.put(GROUP_B_COST, new ArrayList<>());
+            costSortedMap.put(GROUP_C_COST, new ArrayList<>());
+
+            for (Map.Entry<String, String> tagQuery : tagsQuery.entrySet()) {
+                if(tagQuery.getKey().startsWith("!")) {
+                    // In-memory sorted query, requires fetching all the definitions
+                    List<Map.Entry<String, String>> entries = costSortedMap.get(GROUP_C_COST);
+                    entries.add(tagQuery);
+                    costSortedMap.put(GROUP_C_COST, entries);
```
jotak: Useless "put" ?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
```diff
@@ -0,0 +1,243 @@
...
+                } else {
+                    // TODO How to filter exact matching from regexp matching?
+                    List<Map.Entry<String, String>> entries = costSortedMap.get(GROUP_B_COST);
+                    entries.add(tagQuery);
+                    costSortedMap.put(GROUP_B_COST, entries);
```
jotak: Same here

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
```diff
@@ -0,0 +1,243 @@
...
+        if(groupBEntries.isEmpty() && !groupCEntries.isEmpty()) {
+            // Fetch all the available metrics for this tenant
+
+            groupMetrics = dataAccess.findAllMetricsFromTagsIndex()
+                    .compose(new TagsIndexRowTransformer<>(metricType))
+                    .filter(mId -> mId.getTenantId().equals(tenantId))
+                    .flatMap(metricsService::findMetric);
+
+            Observable<Metric<?>> dataMetrics = metricsService.findAllMetrics()
+                    .filter(metricTypeFilter(metricType));
```
jotak: I'm not sure to understand:
first, all tagged metrics are loaded ("findAllMetricsFromTagsIndex"), filtered with tenant id and type
then, all metrics are loaded (again?) ("findAllMetrics") filtered by type but not by tenant id
and finally they're concatenated? Did I miss something?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
```diff
@@ -0,0 +1,243 @@
...
+                                    tagValueMap.put(subKey, values);
+                                }
+                            }));
+
+                    return tagValueMap;
+                });
+    }
+
+    private Func1<Metric<?>, Boolean> tagNotExistsFilter(String unwantedTagName) {
+        return tMetric -> {
```
jotak: or just "!tMetric.getTags().keySet().contains(unwantedTagName);"

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -268,6 +270,10 @@ protected void initPreparedStatements() {
                 "SELECT DISTINCT tenant_id, type, metric, dpart " +
                         "FROM data_compressed");
 
+        findAllMetricsFromTagsIndex = session.prepare(
```
burmanm: Cassandra does not support partial filtering of partition key until 3.10

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
burmanm: It does not actually do the same thing, the latter function has a bad name. "findAllMetrics" fetches everything from the data & data_compressed tables (well, distinct partition keys), but it does not fetch everything from metrics_tags_idx (metricIds, which might not have any data - so no partition key).

So you need to combine the results.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
burmanm: Yep

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
burmanm: Yep

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
burmanm: Yeah, there used to be more functionality in it, but after removing those it would have made sense to refactor it.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TagQueryParser.java
jotak: Ok, thank you for the explanation.
But still, isn't it an issue that the second query (from "findAllMetrics") is not filtered by tenant id? And also, there are potential duplicate entries after concat, no? Shouldn't they be made unique?

jsanda: I am really starting to wonder if we should still store all metric ids in metrics_idx.

jotak:
LGTM

stefannegrea:
@burmanm, do you mind rebasing on master?

stefannegrea:
Thank you @burmanm!

PR #632, opened 2016-10-07
HWKALERTS-172 Migrate Metric External Alerter to Metrics Repo

jshaughn:
- migrate external alerter to metrics repo
- create new alerting-parent module to umbrella alerting jar and x-alerter
- add new x-alerter and alerting i-tests to rest-tests i-tests
- make sure we have proper alerting hooks in place in standalone deployment
- use the proper alerting artifact now that it has been fixed
- exclude transitive Halerting deps. it can cause CP problems
  to have alerting jars in the metrics war when there is
  already a deployment dependency declared from the metrics war to
  the alerting war, at the ear level.
- rename Manager to ExpressionManager in xalerter to be more specific
- trivial: spelling and inline doc fixes
- for safety, let findDataPoints default to ASC order when order is not specified.
- up to alerting src dep


jshaughn:
@stefannegrea Please review.  Note that this moves alerting to a SRC dep, but that should be fine for now, I think.

jshaughn:
@stefannegrea Please review when you have time, this change is needed for 0.21.0, Thanks!

stefannegrea:
Thank you @jshaughn!

PR #633, opened 2016-10-10
[HWKMETRICS-381] dynamically manage gc_grace_seconds for all tables

jsanda:


jsanda:
retest this please

jsanda:
As noted in the Javadocs, the checks and updates only happen on keyspace change events. Table change events are ignored because we don't dynamically add/alter tables at runtime. I am also wondering if there should be something in place to protect against concurrent updates.

stefannegrea:
Can this work be used for updating the replication factor on a keyspace later?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
```diff
@@ -0,0 +1,275 @@
...
+ */
+public class GCGraceSecondsManager implements SchemaChangeListener {
+
+    public static int DEFAULT_GC_GRACE_SECONDS = 604800;
+
+    private static Logger logger = Logger.getLogger(GCGraceSecondsManager.class);
+
+    private RxSession session;
+
+    private PreparedStatement resetGCGraceSeconds;
```
stefannegrea: This is not used. It is needed or left behind by mistake?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
```diff
@@ -0,0 +1,275 @@
...
+
+    private PreparedStatement getGCGraceSeconds;
```
stefannegrea: Should this be final?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
```diff
@@ -0,0 +1,275 @@
...
+     * until schema updates have finished.
+     */
+    public void maybeUpdateGCGraceSeconds() {
+        logger.info("Checking tables in " + keyspace + " to see if gc_grace_seconds needs to be updated");
+        Stopwatch stopwatch = Stopwatch.createStarted();
+        Map<String, String> replication = session.getCluster().getMetadata().getKeyspace(keyspace).getReplication();
+        String replicationFactor = replication.get("replication_factor");
+        Completable check;
+
+        if (getClusterSize() == 1 || replicationFactor.equals("1")) {
```
stefannegrea: Is the replication factor ever not equal to 1 now?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
```diff
@@ -0,0 +1,275 @@
...
+            check = updateAllGCGraceSeconds(0);
+        } else {
+            // Need to call Completable.merge in order for subscriptions to happen correctly. See https://goo.gl/l15CRV
+            check = Completable.merge(configurationService.load("org.hawkular.metrics").map(config -> {
```
stefannegrea: There is a method on the ConfigurationService that returns the value of a specific key. I think that would be better to use here.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/util/GCGraceSecondsManagerITest.java
```diff
@@ -0,0 +1,150 @@
...
+    public void initClass() {
+        String nodeAddresses = System.getProperty("nodes", "127.0.0.1");
+        Cluster cluster = new Cluster.Builder()
+                .addContactPoints(nodeAddresses.split(","))
+                .build();
+        configurationService = new ConfigurationService();
+        configurationService.init(rxSession);
+
+        SchemaService schemaService = new SchemaService();
+        schemaService.run(session, keyspace, Boolean.valueOf(System.getProperty("resetdb", "true")));
```
stefannegrea: Is it a good idea to reset the schema midway through this test set? Any possible side effects?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
jsanda: See MetricsServiceLifecyle.initGCGraceSecondsManager().

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
jsanda: Quite possibly yes. We don't manage Cassandra at all like we did with RHQ. See [this issue](https://github.com/openshift/origin-metrics/issues/213) in origin-metrics.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/util/GCGraceSecondsManagerITest.java
jsanda: These tests use a separate keyspace.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
stefannegrea: I do not see anywhere a reference to `resetGCGraceSeconds`

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/util/GCGraceSecondsManagerITest.java
stefannegrea: Thanks for the clarification; that makes sense.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/util/GCGraceSecondsManager.java
jsanda: Ah sorry Misunderstood :) Yeah, I need to remove it.

jsanda:
> Can this work be used for updating the replication factor on a keyspace later?

I don't think so, but is there something in particular you have in mind?

jsanda:
retest this please

stefannegrea:
Thank you @jsanda!

PR #634, opened 2016-10-10
HWKMETRICS-501 Fix

jshaughn:
The alerting war contained an older hawkular-metrics-model
jar. This was due to a circular dep forced by having the metrics
x-alerter in the alerting repo. The x-alerter is removed from
Halerts 1.2.2.Final, (to be migrated to Hmetrics in 0.21.0), so
change dep to 1.2.2.Final for all but the alerter, leave that at
1.2.1.Final for this micro release to bridge the gap between now
and the Hmetrics 0.21.0 release.
- Also, add a compression itest for avail data

@stefannegrea please review.


burmanm:
There's actually a test for compressed availability data already (and counters) in the CompressJobITest, but more tests of course never hurts.

stefannegrea:
Thank you @jshaughn!

PR #637, opened 2016-10-11
Compress the ccm commands to a single line using create options.

stefannegrea:


stefannegrea:
This is just a documentation change. No need to re-run perf testing.

PR #638, opened 2016-10-12
hwkmetrics-507 One Alerting send per list of inserted metrics

jshaughn:


alerting/alerting-jar/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -108,75 +109,75 @@ public void onMetricsServiceReady(@Observes @ServiceReady ServiceReadyEvent even
...
+            subscription = events.subscribe(list -> onInsertedData(list));
+        }
+    }
+
+    private void onInsertedData(List<Metric<?>> metrics) {
+        List<Data> data = new ArrayList<>(metrics.size());
+        metrics.stream().forEach(m -> addToData(m, data));
+        if (!data.isEmpty()) {
+            try {
+                log.tracef("Sending: %s", data);
```
pilhuhn: Should that be guarded with an if(trace_enabeld) ?

stefannegrea:
retest this please

alerting/alerting-jar/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
jshaughn: It could be but the difference should not be very much as that test should be performed under the covers via this call.   We're not doing any work preparing the parameter.  Unless there is something I don't understand about how these log 'f' calls work.

jshaughn:
@burmanm please take another look, changed things around such that filtering is back in metrics.  This is in addition to still optimizing to one sendData call per batch of metric data.

alerting/alerting-jar/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -86,100 +91,85 @@
...
-        MetricType<?> metricType = metric.getMetricId().getType();
+        MetricId<?> metricId = metric.getMetricId();
+        MetricType<?> metricType = metricId.getType();
         if (metricType == MetricType.UNDEFINED) {
             return;
         }
 
+        String tenantId = metricId.getTenantId();
+        String dataId = prefixMap.get(metricType) + metricId.getName();
+        if (!disablePublishFiltering && !containsKey(tenantId, dataId)) {
```
stefannegrea: Glad to see this change. Not creating objects for all incoming data removes the memory pressure we've seen recently.

alerting/alerting-jar/src/main/java/org/hawkular/metrics/alerting/InsertedDataSubscriber.java
```diff
@@ -85,97 +91,84 @@
...
-                        .buffer(publishPeriod, TimeUnit.MILLISECONDS, 100)
-                        .filter(list -> !list.isEmpty());
-            } else {
-                events = event.getInsertedData()
-                        .filter(m -> m.getType() != MetricType.STRING)
-                        .buffer(publishPeriod, TimeUnit.MILLISECONDS, 100)
-                        .filter(list -> !list.isEmpty());
-            }
+            events = event.getInsertedData()
+                    .buffer(publishPeriod, TimeUnit.MILLISECONDS, 100)
```
lucasponce: If this PR was motivated for a performance test scenario, the buffer should have more control from configuration side (something where we can pass some config to see how buffering might affect to different pressure scenarios IMHO).

jshaughn: I am +1 to making the 100 configurable.  This is unchanged by the PR but was something I noticed as well.

jshaughn:
retest this please

jshaughn:
@stefannegrea can you help me understand why the perf-stability-test has failed twice?  I don't see anything wrong in the log (other than the failed status).

jshaughn:
retest this please

jshaughn:
@stefannegrea  The test results seem to repeat, so perhaps this change is due to the filtering, maybe we need new baselines?

stefannegrea:
Thank you @jshaughn!

jshaughn:
as a note, the perf test actually showed slight gains, not losses, so that may be good.

PR #640, opened 2016-10-12
[HWKMETRICS-494] Retry on failures in compressJob

burmanm:
This PR is intended to fix some behavior that's still failing in the compressionJob.

* Add retry capabilities to the compressBlock, to ensure that the execution never fails.
  * TODO Requires still delayed retry
* Add mergeSort if out-of-order writes come very late

Metrics and manual running in different PR.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
@@ -49,6 +49,9 @@ public class CompressData implements Func1<JobDetails, Completable> {
     public static final String ENABLED_CONFIG = "compression.enabled";
     public static final String BLOCK_SIZE = "compression.block.size";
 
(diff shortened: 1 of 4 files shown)


jsanda:
With the retry() calls, I think it would be good to add some back off logic. I fear that retrying right away could make a bad situation worse in some cases.

I also have a question about the change to use concatMap. I know that emissions will be in order with concatMap whereas with flatMap they will be interleaved. Does that necessarily mean that execution is serial with concatMap, or can things still done concurrently?

jsanda:
I think I found my answer to the question about concatMap. It uses a SerializedSubscriber and the javadocs for it say:

> Enforces single-threaded, serialized, ordered execution of onNext, onCompleted, and onError.

so that is good. As load increases I think we will want some parallelism in the reads, but I think this is good for now.

burmanm:
Yes, as load increases we should have one worker per token range (so that each worker would targe different Cassandra instance).

jsanda:
Could the ```applyRetryPolicy()``` be implemented more easily without a Transformer using map or flatMap?

PR #641, opened 2016-10-13
[HWKMETRICS-505] Several changes to improve error handling

jsanda:
This commit includes a bunch of changes.

* Store job details in scheduled_jobs_idx
I previously only stored the job id in scheduled_jobs_idx and fetched the
details from the jobs table. This eliminates the extra read. I also am no
longer storing the job details in the jobs table. Right now it doesn't really
serve a purpose, and it has actually made things a little more complex because
there are times when the next trigger gets persisted and the job for the
current trigger still has to be retried.

* Do not updated finished_jobs_idx until other post-execution steps are done
I previously updated finished_jobs_idx right after the job was done executing.
If a subsequent post-execution step failed, namely reschulding, the job would
never run again because the computeRemainingJobs() method which determines
eligible jobs to execute filters out jobs that are already in finished_jobs_idx
for the time slice in question.

* Add status column to scheduled_jobs_idx
Rather than updating finished_jobs_idx right away, I now update this new status
column. If the job is then retried after a failure we first check to see if
the status flag is set. If it is we skip job execution and move right to the
post-execution steps.

* Add error logging
I have added a bunch of doOnError(Throwable) calls in the post-execution steps
for improved debugging to make it easier to see what exactly fails.

* Make sure active jobs cache is updated
Previously if there was an error during post-execution, the job id would not
get removed from the active jobs cache. This would prevent subsequent execution
of the job since we filter out those jobs in the active jobs cache when
determining the eligible jobs. A server restart would have been required. Now
I make sure the cache gets updated in a doOnError(Throwable) callback that is
at the end of the post-execution call chain.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
@@ -18,7 +18,9 @@ package org.hawkular.metrics.core.jobs;
 
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -38,7 +40,6 @@ import org.joda.time.Minutes;
 
 import com.google.common.collect.ImmutableMap;
 
-import rx.Observable;
 import rx.Single;
 import rx.functions.Func2;
 
@@ -80,9 +81,10 @@ public class JobsServiceImpl implements JobsService {
     }
 
     @Override
-    public void start() {
+    public List<JobDetails> start() {
         scheduler.start();
 
+        List<JobDetails> backgroundJobs = new ArrayList<>();
         Configuration configuration = configurationService.load("org.hawkular.metrics.jobs")
                 .toBlocking().lastOrDefault(null);
 
@@ -102,8 +104,10 @@ public class JobsServiceImpl implements JobsService {
         if(compressionEnabled) {
             CompressData compressDataJob = new CompressData(metricsService);
             scheduler.register(CompressData.JOB_NAME, compressDataJob);
-            maybeScheduleCompressData();
+            maybeScheduleCompressData(backgroundJobs);
         }
+
+        return backgroundJobs;
     }
 
     @Override
@@ -117,7 +121,7 @@ public class JobsServiceImpl implements JobsService {
                 new SingleExecutionTrigger.Builder().withDelay(1, TimeUnit.MINUTES).build());
     }
 
-    private void maybeScheduleCompressData() {
+    private void maybeScheduleCompressData(List<JobDetails> backgroundJobs) {
         String configId = "org.hawkular.metrics.jobs." + CompressData.JOB_NAME;
         Configuration config = configurationService.load(configId).toBlocking()
                 .firstOrDefault(new Configuration(configId, new HashMap<>()));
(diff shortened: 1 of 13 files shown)


jsanda:
@burmanm or @stefannegrea can you please review this? These changes are needed in conjunction with HWKMETRICS-494.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
```diff
@@ -102,8 +104,10 @@ public void start() {
         if(compressionEnabled) {
             CompressData compressDataJob = new CompressData(metricsService);
             scheduler.register(CompressData.JOB_NAME, compressDataJob);
-            maybeScheduleCompressData();
+            maybeScheduleCompressData(backgroundJobs);
```
burmanm: Wouldn't it make sense to have this method do the insert to the list and the scheduling parts just return the jobDetails?

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/CompressDataJobITest.java
```diff
@@ -193,12 +194,13 @@ public void testCompressJob() throws Exception {
...
         });
 
-        for (JobDetails jobDetails : jobsService.getJobDetails().toBlocking().toIterable()) {
-            if(CompressData.JOB_NAME.equals(jobDetails.getJobName())) {
-                jobScheduler.advanceTimeTo(jobDetails.getTrigger().getTriggerTime());
-                break;
-            }
-        }
+//        for (JobDetails jobDetails : jobsService.getJobDetails().toBlocking().toIterable()) {
+//            if(JOB_NAME.equals(jobDetails.getJobName())) {
```
burmanm: Remove commented stuff

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/api/JobDetails.java
```diff
@@ -21,11 +21,30 @@
...
  * @author jsanda
  */
 public class JobDetails {
 
+//    public enum Status {
+//
+//        FINISHED((byte) 1);
+//
+//        private byte code;
+//
```
burmanm: Remove commented code

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public Observable<ResultSet> updateStatusToFinished(Date timeSlice, UUID jobId) {
+        return session.execute(updateStatus.bind((byte) 1, timeSlice, jobId))
+                .doOnError(t -> logger.warn("There was an error updating the status to finished for [" + jobId +
+                        "] in time slice [" + timeSlice.getTime() + "]"));
+    }
+
+    Trigger getTrigger(UDTValue value) {
+        int type = value.getInt("type");
```
burmanm: Same applies to these.. type is not in any way described.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -737,20 +766,20 @@ public boolean isAcquired() {
     }
 
     private static class JobLock {
-        private UUID jobId;
+        JobDetails jobDetails;
```
burmanm: Missing visibility modifier? Intentional? You have getJobDetails() in any case and that's seems to be used also.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
jsanda: Not sure I understand. Can you explain what you are suggesting?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
jsanda: Actually, it is described in the switch statement that follows:

```java
switch (type) {
            case 0:
                return new SingleExecutionTrigger(value.getLong("trigger_time"));
            case 1:
                return new RepeatingTrigger(
                        value.getLong("interval"),
                        value.getLong("delay"),
                        value.getLong("trigger_time"),
                        value.getInt("repeat_count"),
                        value.getInt("execution_count")
                );
            default:
                throw new IllegalArgumentException("Trigger type [" + type + "] is not supported");
        }
```

I am not sure that an enum adds much value here beyond documentation.

PR #647, opened 2016-10-18
[HWKMETRICS-495] upgrade to cassalog 0.4.0 and add verification checks

jsanda:
The verification checks are used in restart scenarios to help determine whether
or not a change, such as creating a table, has already been applied. Applying
the schema change and updating the changelog table cannot be done atomically
so we need to do some extra work to make sure we are in a known, good state.

### DIFF
--- a/core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.15.0.groovy
+++ b/core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.15.0.groovy
@@ -27,6 +27,7 @@ CREATE TABLE sys_config (
     PRIMARY KEY (config_id, name)
 ) WITH compaction = { 'class': 'LeveledCompactionStrategy' }
 """
+  verify { tableExists(keyspace, 'sys_config') }
 }
 
 schemaChange {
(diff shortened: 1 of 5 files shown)


jshaughn:
@jsanda As we discussed, this will require a similar upgrade to alerting such that dependencies are not out of sync between the two wars.

stefannegrea:
@jsanda, I see this error in the build logs:

```
05:04:05,815 FATAL [org.hawkular.metrics.api.jaxrs.MetricsServiceLifecycle] (metricsservice-lifecycle-thread) HAWKMETRICS200006: An error occurred trying to connect to the Cassandra cluster: groovy.lang.MissingMethodException: No signature of method: org.cassalog.core.CqlChangeSet.verify() is applicable for argument types: (schema-0_15_0$_run_closure1$_closure6) values: [schema-0_15_0$_run_closure1$_closure6@39617ef7]
```

jsanda:
@stefannegrea This is the outstanding class loading issue I mentioned to you yesterday with the EAR. Metrics is picking up the Cassalog dependency from Alerts. For the short term we need to make common dependencies are in sync with respect to version. Alerts needs to upgrade Cassalog any way so I have already submitted [this PR](https://github.com/hawkular/hawkular-alerts/pull/233). Going forward after this upcoming release, we need to sort out the class loading issues. Once the Alerts PR is merged and we update the src dep, this problem should go away. I tested with only the Metrics WAR and everything works as expected.

stefannegrea:
Thank you @jsanda!

PR #652, opened 2016-10-20
[HWKMETRICS-510] add configurable retry logic for inserting data points

jsanda:
If insertion of data points fail, we will use an exponential back off delay
with caps on the delay and the number of retries. The max delay and max number
of retries are stored in the sys_config table and can be updated via system
properties or environment variables.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -75,6 +75,7 @@ import org.hawkular.metrics.model.exception.MetricAlreadyExistsException;
 import org.hawkular.metrics.model.exception.TenantAlreadyExistsException;
 import org.hawkular.metrics.model.param.BucketConfig;
 import org.hawkular.metrics.model.param.TimeRange;
+import org.hawkular.metrics.sysconfig.Configuration;
 import org.hawkular.metrics.sysconfig.ConfigurationService;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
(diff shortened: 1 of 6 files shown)


stefannegrea:
retest this please

FilipB:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -324,15 +329,20 @@ private void initMetrics() {
...
                 .lastOrDefault(null);
+        String configMaxStringSize = configuration.get("string-size");
         if (configMaxStringSize == null) {
             maxStringSize = -1;  // no size limit
         } else {
             maxStringSize = Integer.parseInt(configMaxStringSize);
         }
         log.infoMaxSizeStringMetrics(this.maxStringSize);
+
+        insertRetryMaxDelay = Long.parseLong(configuration.get("ingestion.retry.max-delay", "30000"));
```
stefannegrea: I do not think this logic is good because the value from the command line is overwritten with the value from the database every single time and there is no way to save value from the command line back into the db. I am missing something?

jsanda: The value from the cmd line is stored in the database. MetricsServiceImpl then fetches it from the database. We also set defaults in the database so you don't have to pass values via the command line, but command line values will always take precedence.

stefannegrea:
Thank you @jsanda!

FilipB:
This PR introduced perf drop 7% which was missed because of env issue. Investigation still in progress in https://github.com/hawkular/hawkular-metrics/pull/668

FilipB:
This PR introduced 7% perf drop probably caused by https://github.com/ReactiveX/RxJava/issues/3633

Agreed to update baselines and wait for 2.x version

PR #655, opened 2016-10-21
HWKMETRICS-515 Make external alerter cluster-aware

jshaughn:
- use ISPN CacheManager to inspect and react to cluster topology
- limit expression evaluation to only the cooridinater node

@lucasponce Please review as you are expert in cluster topology handling via CacheManager
@stefannegrea Please also optionally review

### DIFF
--- a/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ExpressionManager.java
+++ b/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ExpressionManager.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
+import javax.annotation.Resource;
 import javax.ejb.Singleton;
 import javax.ejb.Startup;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ExpressionManager.java
```diff
@@ -81,31 +94,91 @@
...
+            processTopologyChange();
+        }
+    }
+
+    private void processTopologyChange() {
+        boolean currentCoordinator = coordinator;
+        coordinator = distributed ? cacheManager.isCoordinator() : true;
+
+        if (coordinator && !currentCoordinator) {
+            start();
```
lucasponce: Logic looks good.
The only think I wonder is about the speed.
I mean, jgroups tends to move the coordinator, so on startups a real possibility is that the new node becomes coordinator.
So, you can count that stop() and start() can be launched separately.
My only doubt if those process are safe or if they need some kind of sync.
I.e. a start() can not happen until a stop() in other node is fully finished.
If that is ok, the code is correct.
If not, perhaps some locking is necessary.

jshaughn: That's a good observation and if it is an issue I think it will probably be solved later, when we convert to using Metrics' job scheduler (see the Jira for a comment about the two considered solutions).  For now, hopefully JGroups/ISPN informs listeners in a timely manner, because the stop() operation is quite fast, just issues any necessary cancels.   Although possible, if there is brief overlap, it is unlikely to generate duplicate alerts given that alerts should be rare.

stefannegrea:
Thank you @jshaughn!

PR #664, opened 2016-10-26
HWKMETRICS-515 Make external alerter cluster-aware

jshaughn:
- use ISPN CacheManager to inspect and react to cluster topology
- limit expression evaluation to only the cooridinater node


stefannegrea:
Thank you @jshaughn!

PR #665, opened 2016-10-26
[HWKMETRICS-520] make query page size configurable

jsanda:
The compression job defaults to a page size of 1,000. This only effects queries
on raw data.


jsanda:
retest this please

burmanm:
See my comments on the other PR for this one.

jsanda:
> If 5000 is causing us issues with reads in Cassandra 3.0.9, there's no point using this as default. Although you see it in compression job more easily (because it does reads), there's nothing preventing it from being triggered with other read requests also.
>
> Especially since those reads can be concurrent while the compression job does not do concurrent reads.

Not  sure I agree. We have not had any (at least reported) problems with OOMEs because the read patterns for user-facing queries do not introduce that much heap pressure. The decreased page size does slow down the compression job. It is running on my machine in a few minutes as opposed to 10 or 15 seconds. I would prefer to avoid slowing down user-facing queries if possible.

burmanm:
That itself is worrisome then. The compression job requests at most 2 hours of data at once -> processes it -> requests another batch. I can't imagine we wouldn't face queries that wouldn't read more than 2 hours of data.

This means two simultaneous queries should put more read pressure on the Cassandra than CompressionJob. A single read might read far more rows. Have you tried reading when using those insert-heavy-tests? If you read the same stuff as compressionjob, you should get the same results then.

The only thing that compression job does more than any other query is deletes.

jsanda:
> Have you tried reading when using those insert-heavy-tests?

I have not, but I will do just that.

jsanda:
I have not had a chance to do the testing, so I went ahead and changed the default page size as well to be 1000. I think this is now ready for final review/merge.

stefannegrea:
Thank you @jsanda!

PR #666, opened 2016-10-27
[HWKMETRICS-524] add flag for disabling compression job

jsanda:
The job can be enabled/disabled with either the system property

```
-Dhawkular.metrics.jobs.compression.enabled=false
```

or using the environment variable

`export COMPRESSION_JOB_ENABLED=false`

The job actually still runs but is just a no-op. This is important from a job scheduling standpoint. 
suppose the job is disabled just before 13:00, and it is re-enabled at 23:00. Raw data ingested
between 11:00 and 21:00 will not get compressed.

This has to be merged after [PR 665](https://github.com/hawkular/hawkular-metrics/pull/665).

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
@@ -27,6 +27,7 @@ import org.hawkular.metrics.datetime.DateTimeService;
 import org.hawkular.metrics.model.Metric;
 import org.hawkular.metrics.model.MetricId;
 import org.hawkular.metrics.scheduler.api.JobDetails;
+import org.hawkular.metrics.sysconfig.Configuration;
 import org.hawkular.metrics.sysconfig.ConfigurationService;
 import org.jboss.logging.Logger;
 import org.joda.time.DateTime;
@@ -60,19 +61,28 @@ public class CompressData implements Func1<JobDetails, Completable> {
 
     private int pageSize;
 
+    private boolean enabled;
+
     public CompressData(MetricsService service, ConfigurationService configurationService) {
         metricsService = service;
-        String pageSizeConfig =  configurationService.load(CONFIG_ID, "page-size")
-                .toBlocking().firstOrDefault(null);
-        if (pageSizeConfig == null) {
+        Configuration configuration = configurationService.load(CONFIG_ID).toSingle().toBlocking().value();
+        if (configuration.get("page-size") == null) {
             pageSize = DEFAULT_PAGE_SIZE;
         } else {
-            pageSize = Integer.parseInt(pageSizeConfig);
+            pageSize = Integer.parseInt(configuration.get("page-size"));
         }
+
+        String enabledConfig = configuration.get("enabled", "true");
+        enabled = Boolean.parseBoolean(enabledConfig);
+        logger.debugf("Job enabled? %b", enabled);
+
     }
 
     @Override
     public Completable call(JobDetails jobDetails) {
+        if (!enabled) {
+            return Completable.complete();
+        }
         // Rewind to previous timeslice
         Stopwatch stopwatch = Stopwatch.createStarted();
         logger.info("Starting execution");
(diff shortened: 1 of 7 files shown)


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -663,18 +667,70 @@ public void setDefaultTTL(int defaultTTL) {
...
+//                    break;
+//                case DESC:
+//                    comparator = (tDataPoint, t1) -> (int) (t1.getTimestamp() - tDataPoint.getTimestamp());
+//                    break;
+//                default:
+//                    throw new RuntimeException(safeOrder.toString() + " is not correct sorting order");
+//            }
+//
+//            Observable<DataPoint<T>> dataPoints = SortedMerge
+//                    .create(Arrays.asList(uncompressedPoints, compressedPoints), comparator, false, true);
```
burmanm: Remove commented out code

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/jobs/CompressDataJobITest.java
```diff
@@ -22,10 +22,14 @@
...
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
+import static org.testng.AssertJUnit.assertNull;
+import static org.testng.AssertJUnit.assertTrue;
+//import static org.junit.Assert.assertEquals;
+//import static org.junit.Assert.assertNotNull;
```
burmanm: Remove

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
```diff
@@ -48,18 +50,39 @@
...
         metricsService = service;
+        Configuration configuration = configurationService.load(CONFIG_ID).toSingle().toBlocking().value();
+        if (configuration.get("page-size") == null) {
+            pageSize = DEFAULT_PAGE_SIZE;
+        } else {
+            pageSize = Integer.parseInt(configuration.get("page-size"));
+        }
+
+        String enabledConfig = configuration.get("enabled", "true");
+        enabled = Boolean.parseBoolean(enabledConfig);
```
burmanm: Boolean.valueOf() is correct one, Boolean.parseBoolean() searches for System.property

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/JobsServiceImpl.java
```diff
@@ -102,7 +102,7 @@ public void setScheduler(Scheduler scheduler) {
 
         Boolean compressionEnabled = Boolean.valueOf(configuration.get(CompressData.ENABLED_CONFIG, "true"));
         if(compressionEnabled) {
-            CompressData compressDataJob = new CompressData(metricsService);
+            CompressData compressDataJob = new CompressData(metricsService, configurationService);
```
burmanm: This is now confusing.. we have two compress jobs enabled settings, doing different things.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/jobs/CompressData.java
```diff
@@ -48,18 +50,39 @@
...
     private MetricsService metricsService;
 
-    public CompressData(MetricsService service) {
+    private int pageSize;
+
+    private boolean enabled;
+
+    public CompressData(MetricsService service, ConfigurationService configurationService) {
         metricsService = service;
+        Configuration configuration = configurationService.load(CONFIG_ID).toSingle().toBlocking().value();
```
burmanm: Shouldn't this be in the call() method? Otherwise it can't be re-enabled without recreating the object.

jsanda: I intended to put it in the constructor. A server restart is required for changes to take effect. We could put it in the call() method, but since we do not currently expose any APIs for making the runtime change, I do not see any benefit. The user could directly update the database, but I do not think that is a good idea.

stefannegrea:
Thanks @jsanda!

PR #667, opened 2016-10-27
Migrate ptrans ITest to run against standalone ear

jshaughn:
Migrate ptrans ITest to run against standalone ear as opposed to the jaxrs war, which is no longer a formal dist.


stefannegrea:
Thank you @jshaughn!

PR #669, opened 2016-10-27
Thin Artifacts

jshaughn:
- up alerting version to use thinner wars src dep
- remove cassandra-driver as transitive dep in api-jaxrs war, and package it explicitly where needed.
  - for component war do not package cassandra-driver, it is provided by hawkular commons module.
  - for standalone and openshift ear, package cassandra driver
- add comments in ear dist poms explaining why cassalog/cassandra deps are not at ear level
- remove jboss-deployment-structure.xml from jaxrs-api war, it is superseded by the ear-level version.

Note - this PR does not significantly thin the dist ears because my attempts to share jars are ear level were not met with stability.  Nevertheless, there is some useful cleanup here, and wrt the component ear, there is a 9% size reduction.


jshaughn:
@stefannegrea this is failing perf because of the old baselines, I think.  This should have no effect on perf.

stefannegrea:
Thank you @jshaughn!

PR #671, opened 2016-10-31
[HWKMETRICS-528] Fix error code of missing admin filter

burmanm:
It currently returns 400 for all the error situations, this makes it 400 for missing code and 403 for incorrect code


stefannegrea:
Thank you @burmanm!

PR #679, opened 2016-11-03
[HWKMETRICS-514] a distributed lock to coordinate schema changes

jsanda:
This PR requires changes to standalone.xml and standalone-ha.xml. For standalone.xml we need:

```xml
<cache-container name="hawkular-alerts" default-cache="triggers" statistics-enabled="true">
    <local-cache name="partition"/>
    <local-cache name="triggers"/>
    <local-cache name="data"/>
    <local-cache name="publish"/>
    <local-cache name="schema"/>
    <local-cache name="locks"/>
```

Note the addition of the **locks** cache. And for standalone-ha.xml:

```xml
<cache-container name="hawkular-alerts" 
                              default-cache="triggers"
                              module="org.wildfly.clustering.server" 
                              statistics-enabled="true">
    <replicated-cache name="partition" mode="SYNC">
        <transaction mode="BATCH"/>
    </replicated-cache>
    <replicated-cache name="triggers" mode="ASYNC">
        <transaction mode="BATCH"/>
    </replicated-cache>
    <replicated-cache name="data" mode="ASYNC">
        <transaction mode="BATCH"/>
    </replicated-cache>
    <replicated-cache name="publish" mode="ASYNC">
        <transaction mode="BATCH"/>
    </replicated-cache>
    <replicated-cache name="schema" mode="SYNC">
        <transaction mode="NON_XA"/>
    </replicated-cache>
    <replicated-cache name="locks" mode="SYNC">
        <transaction mode="NON_XA" locking="PESSIMISTIC"/>
    </replicated-cache>
```

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -110,7 +110,6 @@ import com.google.common.util.concurrent.Uninterruptibles;
  * @author Thomas Segismont
  */
 @ApplicationScoped
-//@Eager
 public class MetricsServiceLifecycle {
     private static final RestLogger log = RestLogging.getRestLogger(MetricsServiceLifecycle.class);
 
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -231,9 +231,6 @@ public class MetricsServiceLifecycle {
     @ServiceReady
     Event<ServiceReadyEvent> metricsServiceReady;
 
-//    @Resource(lookup = "java:jboss/infinispan/container/hawkular-alerts")
-//    private EmbeddedCacheManager cacheManager;
-
     @Resource(lookup = "java:jboss/infinispan/cache/hawkular-alerts/locks")
     private Cache<String, String> locksCache;
 
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -53,8 +53,11 @@ import java.util.concurrent.ThreadFactory;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
+import javax.annotation.Resource;
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Initialized;
 import javax.enterprise.event.Event;
+import javax.enterprise.event.Observes;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 import javax.net.ssl.SSLContext;
@@ -63,7 +66,6 @@ import org.hawkular.metrics.api.jaxrs.config.Configurable;
 import org.hawkular.metrics.api.jaxrs.config.ConfigurationProperty;
 import org.hawkular.metrics.api.jaxrs.log.RestLogger;
 import org.hawkular.metrics.api.jaxrs.log.RestLogging;
-import org.hawkular.metrics.api.jaxrs.util.Eager;
 import org.hawkular.metrics.api.jaxrs.util.JobSchedulerFactory;
 import org.hawkular.metrics.api.jaxrs.util.MetricRegistryProvider;
 import org.hawkular.metrics.core.jobs.CompressData;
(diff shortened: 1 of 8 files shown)


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -227,6 +231,12 @@
     @ServiceReady
     Event<ServiceReadyEvent> metricsServiceReady;
 
+//    @Resource(lookup = "java:jboss/infinispan/container/hawkular-alerts")
```
stefannegrea: Do you mind removing commented code?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -106,7 +110,7 @@
  * @author Thomas Segismont
  */
 @ApplicationScoped
-@Eager
+//@Eager
```
stefannegrea: Do you mind removing if not needed?

stefannegrea:
We will need to also update the WF distribution too to add the new caches.

FilipB:
retest this please

stefannegrea:
Thank you @jsanda!

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -255,6 +261,10 @@ public State getState() {
         return state;
     }
 
+    void eagerInit(@Observes @Initialized(ApplicationScoped.class) Object event) {
+        // init
+    }
+
```
jshaughn: What effect does this have?

api/metrics-api-jaxrs/src/main/webapp/WEB-INF/web.xml
```diff
@@ -24,6 +24,16 @@
...
+  <resource-env-ref>
+    <resource-env-ref-name>container/hawkular-alerts</resource-env-ref-name>
+    <lookup-name>java:jboss/infinispan/container/hawkular-alerts</lookup-name>
+  </resource-env-ref>
+
+  <resource-env-ref>
+    <resource-env-ref-name>cache/locks</resource-env-ref-name>
+    <lookup-name>java:jboss/infinispan/cache/hawkular-alerts/locks</lookup-name>
+  </resource-env-ref>
+
```
jshaughn: So, prior to this change Metrics nodes did not require clustering, right?  Although, non-clustered nodes meant that Alerting would likely have had to be disabled.   Now Metrics requires clustering and ISPN for schame handling, right?  This seems OK to me given the Alerting needs and the overall strategy to have clustered nodes, but I just wanted to confirm that this was basically true.

dist/component/component-war/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -27,5 +27,7 @@
       <module name="org.jboss.resteasy.resteasy-jackson2-provider" services="import"/>
       <module name="org.hawkular.bus"/>
     </dependencies>
+    <module name="org.infinispan" export="true" />
+    <module name="org.infinispan.commons" export="true"/>
```
jshaughn: This is not going to have any effect.  In fact, I thought I had removed the WAR-level jboss-deployment-structure.xml.  The WAR-level instances of this file are ignored, only the EAR-level version is relevant.

jshaughn:
I believe you will need a PR for Hservices as well, to add the cache in /hawkular-services/feature-pack/src/main/xsl/subsystem-templates/hawkular-services-infinispan.xsl ?

jshaughn:
Thinking further, should this mechanism perhaps be made available via commons as opposed to metrics?  You've used the hawkular-alerts cache-container, which is admittedly more convenient than adding a hawkular-metrics cache-container, but alerting can't leverage it without actually duplicating the DistributedLock stuff, I think.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
jsanda: The ```@Eager``` annotation once again was causing problems. Until eager initialization makes its way into CDI, this is the way to do it.

dist/component/component-war/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
jsanda: Since you have been handling these changes and are more familiar, I will defer to you on what needs to be done. I'd like to get any other changes needed done ASAP so @mwringe can build/test.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
jshaughn: That's good to know.  You may want to add that as an inline comment if you end up going back into this file.

jsanda:
> Thinking further, should this mechanism perhaps be made available via commons as opposed to metrics?

If this is going to be used outside of Metric, then yes. There is more work to do with this, but that the PR was good enough to go into the release and further changes could happen later. I may even wind up replacing the locking that the job scheduler uses with this.

dist/component/component-war/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
jshaughn: The EARs already have an infinspan dep declared.  They don't have org.infinispan.commons declared, but iiuc this is not needed, because as I mentioned, this file should be getting ignored anyway.  Take it away and ensure that it still works as expected.

PR #683, opened 2016-11-04
[HWKMETRICS-536] check if jobs in finished state are already rescheduled

jsanda:
When we go to execute a job that is already in the finished state we need to
check if the job has already been rescheduled. If so, we skip rescheduling to
avoid an execution loop.


jsanda:
retest this please

jsanda:
retest this please

stefannegrea:
Thank you @jsanda!

PR #687, opened 2016-11-07
[HWKMETRICS-535] use configurable, parallel queries for compression job

jsanda:
The main part of this commit is using flatMap instead of concactMap in
MetricsService.compressBlocks(). By default, we are not introducing any
parallelism; however, it is configurable with the system property ```hawkular.metrics.jobs.compression.parallel-reads``` or the environment variable ```COMPRESSION_PARALLEL_READS```.


jsanda:
@FilipB can you run the overnight job with increased parallelism, e.g., 

```-Dhawkular.metrics.jobs.compression.parallel-reads=8```

FilipB:
retest this please

burmanm:
Actually, this should follow the idea I had documented in the code. We should do the parallelism based on the token ranges (targeting the nodes) and then let that be configured. This was one of the reasons for the original "tokenAware" stuff which I removed in the final commit.

PR #692, opened 2016-11-10
Perf debug

stefannegrea:


burmanm:
..why?

FilipB:
To find out if perf test failures are caused by driver version. There will be jira once it's clear what's causing it.

FilipB:
The issue is still visible - https://issues.jboss.org/browse/HWKMETRICS-542

stefannegrea:
I will close this since @jsanda  is working on a fix with the current driver version.

PR #695, opened 2016-11-16
[HWKMETRICS-542] increase max queue size to work around BusyPoolException

jsanda:


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -439,6 +439,7 @@ private Session createSession() {
                 .setMaxConnectionsPerHost(HostDistance.REMOTE, newMaxConnections)
                 .setMaxRequestsPerConnection(HostDistance.LOCAL, newMaxRequests)
                 .setMaxRequestsPerConnection(HostDistance.REMOTE, newMaxRequests)
+                .setMaxQueueSize(1024)
```
jshaughn: @jsanda It may be useful to make this configurable.

@lucasponce As an aside, there are a variety of options used here on the clusterBuilder that perhaps we should consider?  Or, perhaps commons would be a place for some common driver settings, that could then be overridden in a component?

jsanda: I have hard coded the changed for now simply to see it gets past the error in the perf test. There are a number of driver settings that are worth exposing. Metrics and alerts should probably share the same Session instance.

FilipB:
The exception is still there for bigger msgs. Now with increased pool number - no available connection and the queue has reached its max size 1024

PR #696, opened 2016-11-18
[HWKMETRICS-545] [HWKMETRICS-502] Compression job control

burmanm:
This PR adds the ability of manually adjusting the blockSize of the compression (accepting the ISO8601 format as parameter) as well as creating MBean for the compression job triggering for given timestamp.

Calculation of the compressed range is moved from MetricsServiceImpl to CompressData to allow customized ranges to be triggered.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsService.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsService.java
@@ -171,7 +171,17 @@ public interface MetricsService {
     <T> Observable<DataPoint<T>> findDataPoints(MetricId<T> id, long start, long end, int limit, Order order,
             int pageSize);
 
-    Observable<Void> compressBlock(Observable<? extends MetricId<?>> metrics, long timeSlice, int pageSize);
+    /**
+     * Compresses the given range between timestamps to a single block.
+     *
+     * @param metrics Compressed metric definitions
+     * @param startTimeSlice Start time of the block
+     * @param endTimeSlice End time of the block
+     * @param pageSize Cassandra query parameter
+     * @return onComplete when job is done
+     */
+    Observable<Void> compressBlock(Observable<? extends MetricId<?>> metrics, long startTimeSlice, long endTimeSlice,
+                                   int pageSize);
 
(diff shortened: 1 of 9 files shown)


burmanm:
@jsanda I'm seeing some issue with scheduler occasionally (does not happen every time), where it does not actually schedule the manual run at all.

https://gist.github.com/burmanm/56af20a0ae2b22a98791e291c8481b62

jsanda:
Can you change ```submitCompressJob(Map<String, String> parameters)``` to subscribe to the returned Single and log the results? If an error occurs scheduling the job, it is just getting swallowed.

burmanm:
@jsanda Yes, I can try debugging that way. Should I however modify the original scheduleJob to give it .doOnError() instead of making each caller do the subscribe + logging just to get stats?

jsanda:
> Should I however modify the original scheduleJob to give it .doOnError() instead of making each caller do the subscribe + logging just to get stats?

That sounds good. We could use some INFO or DEBUG statements in SchedulerImpl.scheduleJob() that logs the outcome.

FilipB:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsService.java
```diff
@@ -171,7 +171,17 @@
     <T> Observable<DataPoint<T>> findDataPoints(MetricId<T> id, long start, long end, int limit, Order order,
             int pageSize);
 
-    Observable<Void> compressBlock(Observable<? extends MetricId<?>> metrics, long timeSlice, int pageSize);
```
jsanda: Why not return Completable here? One of its primary use cases was to address Observable<Void>. I know that we have plenty of places where we use Observable<Void>, but I would like to eventually replace them with Completable provided we only care about a terminal event which is almost always the case.

jsanda:
Could you add a test that demonstrates scheduling a compression job via JMX?

burmanm:
That would require creating some sort of access methods to the job scheduler. As the JMX is initialized in the Wildfly phase, it requires the whole Wildfly to be running and only then can I start doing the JMX requests. So same sort of packaging as the rest-tests and then request the scheduler's statistics I guess..

Or Arquillian perhaps. Ugh.

burmanm:
With blocking processing I haven't seen missed schedulings.

PR #697, opened 2016-11-21
[HWKMETRICS-543] Missing type in exact id filtering should return 400

burmanm:
The id filtering without tags allows only exact matching, so the type is required. If it's missing, return 400 instead of 500.


burmanm:
retest this please

PR #698, opened 2016-11-21
[HWKMETRICS-550] persist internally collected metrics

jsanda:
This commit adds support for persisting internally collected metrics provided that they follow the naming scheme specified in MetricNameService.


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
```diff
@@ -0,0 +1,81 @@
...
+ * Provides some basic functions for DropWizard metric names. In order to avoid naming collisions, each metric name
+ * is prefixed with the hostname, which defaults to the hostname of the local host. The format is,
+ * <code>hostname:metric_name</code>. All metric names should be created using {@link #createMetricName(String)} in
+ * order to ensure a consistent format.
+ *
+ * @author jsanda
+ */
+public class MetricNameService implements MetricFilter {
+
+    public static final String TENANT_ID = "org.hawkular.metrics";
```
stefannegrea: This should be the standard admin user for Metrics. Ultimately these metrics will need to be protected with the admin token, but that can be part of another PR/JIRA ticket.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricsInitializer.java
```diff
@@ -0,0 +1,134 @@
...
+
+        List<Metric<Double>> gaugeMetrics = new ArrayList<>();
+        List<Metric<Long>> counterMetrics = new ArrayList<>();
+
+        metricRegistry.getMeters(metricNameService).entrySet().forEach(entry -> {
+            String tenantId = MetricNameService.TENANT_ID;
+            String metricName = metricNameService.getMetricName(entry.getKey());
+            Map<String, String> tags = getTags(metricName);
+
+            gaugeMetrics.add(new Metric<>(new MetricId<>(tenantId, GAUGE, metricName + "-1min"), tags));
```
jotak: I think we should adopt the same naming convention than with the hawkular-reporter, see here: https://github.com/hawkular/hawkular-dropwizard-reporter/blob/master/hawkular-dropwizard-reporter/src/main/java/org/hawkular/client/dropwizard/HawkularReporter.java#L171

So, it should be changed either here or on the reporter side, no? As I understand, parts of a metric name are quite often (but maybe not conventionally) separated with dots.

FilipB:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
jsanda: What should be the standard admin user?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricsInitializer.java
jsanda: We can change the names, but I do not really think it is that important. Hawkular Metrics does not (yet) impose any restrictions on metric names. Each client may wind up having different naming conventions because of this. The hawkular-dropwizard-client is a separate REST client which can change independently of any internal reporting within hawkular-metrics.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
stefannegrea: The original design document for admin features does not name an admin tenant. Two obvious choices are `admin` and `metrics-admin`. I prefer first since it is shorter and simpler, but the second one is more useful if Alerting will want their own admin tenant too.

jsanda:
I have changed the tenant id to admin,  and I have also added a tag named component which has a value of org.hawkular.metrics. For the Cassandra driver related metrics, the component tag would have a value of com.datastax.driver. For hawkular-alerts' metrics, the component tag value would be org.hawkular.alerts.

I was considering adding a description tag as well, but I decided to hold off for now. Some of the DropWizard metrics map to multiple hawkular-metrics metrics. I am not sure a description should be duplicated for each of those metrics.

stefannegrea:
Thank you @jsanda!

PR #701, opened 2016-11-28
[HWKMETRICS-526] report status of Cassandra nodes

jsanda:
The output of the status endpoint now looks like:

```json
{
  "MetricsService": "STARTED", 
  "Built-From-Git-SHA": "291fabc1214149653ee3df14e0f2c60b10c64360", 
  "Implementation-Version": "0.22.0-SNAPSHOT", 
  "Cassandra": [
    {
      "address": "127.0.0.1", 
      "status": "up"
    }, 
    {
      "address": "127.0.0.2", 
      "status": "up"
    }, 
    {
        "address": "127.0.0.3",
        "status": "up"
    }
  ]
}
```

@mwringe does this work for you?

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -20,6 +20,7 @@ package org.hawkular.metrics.api.jaxrs;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
+import static java.util.stream.Collectors.toList;
 
 import static org.hawkular.metrics.api.jaxrs.config.ConfigurationKey.ADMIN_TOKEN;
 import static org.hawkular.metrics.api.jaxrs.config.ConfigurationKey.CASSANDRA_CONNECTION_TIMEOUT;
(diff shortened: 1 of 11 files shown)


mwringe:
Do we want to expose the ip address of the nodes? The status endpoint currently is public and I don't know if we want the ip address of the Cassandra nodes should be considered more private information.

jsanda:
If a node is down, we won't be able to see which one is down without including the addresses. What is the security concern/risk with exposes the addresses?

stefannegrea:
@jsanda, exposing the information about the infrastructure is always a security concern. For example, it allows an attacker to estimate the resources needed for a successful DDoS attack.

@jsanda, @mwringe, how about a separate protected admin status endpoint (`admin/status`) that contains the existing status info plus change proposed in this PR. Later we could expand the admin status to container other sensitive information such as WF topology.

mwringe:
A separate /admin/status endpoint would work here

jsanda:
What do we want to for the existing status endpoint with respect to Cassandra? Here is one idea if we still want to include Cassandra info on the existing status endpoint. If all nodes are up, we report up. If all nodes are down we report down. If some nodes are down we report something like partially up. Thoughts?

stefannegrea:
If some nodes are down, it is possible the cluster is  operational but have reduced performance or not operational at all. @jsanda, how about calling the third state `degraded`?

jsanda:
> If some nodes are down, it is possible the cluster is operational but have reduced performance or not operational at all. 

The cluster is still operational, but some requests will fail. If we have nodes A, B, and C with RF = 1, then each node owns roughly 33.3% of keys. If node B goes down, requests targeted for B will fail, but requests targeted for A and C will succeed.

> how about calling the third state degraded?

Works for me.

stefannegrea:
Thanks @jsanda!

PR #707, opened 2016-12-07
[HWKMETRICS-554] Fix several issues with CORS

jotak:
- payload was sent even on error 400 on invalid origin. This was because CORS validation occured at response filtering => move validation to request filter

- CORS related headers not sent for some endpoints. Related to how NamedDataPointObserver was handling the async response, i think the output stream was closed and subsequent response filter adding headers were ineffective; now use AsyncResponse::resume to complete the response.


rhqci:

metlos:
Other components use https://github.com/hawkular/hawkular-commons/tree/master/hawkular-cors-jaxrs-filter. It would be great if metrics used it, too... Note that the commons filter is a response filter, too, so maybe it will need fixing, too, along with the components that use it.

jotak:
I'm updating this PR to use code from commons... (and add tests) please don't merge yet

jotak:
For the time being it uses commons' SNAPSHOT, until a new release is made

jshaughn:
@jotak I think you can update the commons version from src dep to 0.9.1.Final.  Given that change I think @stefannegrea perhaps may want to merge in these CORS fixes for the next release.

jotak:
yep I'll update it

jotak:
Test failed: TenantITest.deleteTenantTwiceConcurrently:572 expected:<204> but was:<200>
Seems unrelated to this PR.

jsanda:
I agree that is unrelated. I restarted travis.

stefannegrea:
Thank you @jotak!

PR #708, opened 2016-12-07
[HWKMETRICS-555] include hostname in id when creating metrics

jsanda:


FilipB:
@jsanda does it mean that metric ids used in alert triggers will have to be renamed as well?

jsanda:
> does it mean that metric ids used in alert triggers will have to be renamed as well?

This does not impact alerting in any way. This is only for internally generated metrics to ensure metrics from different hosts do not overwrite one another.

stefannegrea:
Thank you @jsanda!

PR #713, opened 2016-12-16
[HAWKULAR-1168] make replication_factor configurable

jsanda:
It can be set at start up using the hawkular.metrics.cassandra.replication-factor
system property or the CASSANDRA_REPLICATION_FACTOR environment variable.


jsanda:
retest this please

FilipB:
retest this please

stefannegrea:
@jsanda, what happens if the user restarts with a different replication factor?

jsanda:
> what happens if the user restarts with a different replication factor?

Nothing because the change is only applied on initial keyspace creation. This will only work for new installations, not upgrades.

stefannegrea:
@jsanda, would't be a good idea to have a warning in case users try to use this flag beyond the initial installation? Maybe in cases where the new replication factor is different than the one used.

jsanda:
If the property is set I will log an INFO message with the current/new replication_factor. That is the easiest thing to do because in our start up code we do not do anything special to determine whether it is a new install or update. That install vs update logic is limited to bootstrap.groovy. I'd prefer to keep it that way to the extent possible.

jsanda:
retest this please

stefannegrea:
Thank you @jsanda!

PR #717, opened 2017-01-01
[HWKMETRICS-563] Added ability to group by datapoint tag key

garethahealy:
This is primarily for enabling grafana (will raise a PR/JIRA) to display multiple metrics that are collected from a prometheus endpoint.

FYI: Day1 of using Java Lambda/Rx, so happy for input if this isn't the best solution.

https://issues.jboss.org/browse/HWKMETRICS-563


rhqci:

FilipB:
ok to test

FilipB:
retest this please

PR #718, opened 2017-01-04
Update RxJava to 1.2.4

burmanm:
Updates RxJava version 1.2.4 (some API changes and performance improvements compared to 1.1.7)


metlos:
Since multiple hawkular components use RxJava, wouldn't it be better to consolidate this in the parent pom (which currently has rxjava 1.1.7)?

stefannegrea:
@metlos, no reason to stay behind; this is just a small but needed update. Also, we have an upgrade to 2.0 planned.

PR #719, opened 2017-01-06
[HWKMETRICS-564] fix several bugs

jsanda:
When jobs fall behind and are rescheduled for immediate execution, we first
renew the job lock. The job should only be executed again if the lock is
successfully renewed. We were always executing the job. This is fixed.

When there are jobs to execute across multiple time slices, we were executing
things out of order. This is fixed simply by replacing flatMap with concatMap
in a few places.

When retrying a job execution, we were updating finished_jobs_idx with the
wrong timestamp. We were using the trigger time instead of the time in which
the job is scheduled. In this scenario, the scheduled time != the trigger time.

### DIFF
--- a/job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
+++ b/job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -140,7 +140,8 @@ public class SchedulerImpl implements Scheduler {
     static final int SCHEDULING_LOCK_TIMEOUT_IN_SEC = 5;
 
     // TODO We probably want this to be configurable
-    static final int JOB_EXECUTION_LOCK_TIMEOUT_IN_SEC = 3600;
+//    static final int JOB_EXECUTION_LOCK_TIMEOUT_IN_SEC = 3600;
+    static final int JOB_EXECUTION_LOCK_TIMEOUT_IN_SEC = 60;
 
     /**
      * Test hook. See {@link #setTimeSlicesSubject(PublishSubject)}.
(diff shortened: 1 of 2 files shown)


job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -140,7 +140,8 @@
     static final int SCHEDULING_LOCK_TIMEOUT_IN_SEC = 5;
 
     // TODO We probably want this to be configurable
-    static final int JOB_EXECUTION_LOCK_TIMEOUT_IN_SEC = 3600;
+//    static final int JOB_EXECUTION_LOCK_TIMEOUT_IN_SEC = 3600;
```
stefannegrea: This was confusing change, but from what I understand needs to be reverted.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
```diff
@@ -269,27 +272,42 @@ public void start() {
...
                             .flatMap(this::acquireJobLock)
                             .filter(jobLock -> jobLock.acquired)
                             .map(jobLock -> jobLock.jobDetails)
-                            .doOnNext(details -> logger.debugf("Acquired job lock for %s in time slice %s", details,
-                                    timeSliceLock.timeSlice))
-                            .flatMap(details -> executeJob(details, timeSliceLock.timeSlice, activeJobs).toObservable()
+                            .doOnNext(details -> {
+                                logger.debugf("Acquired job lock for %s in time slice %s", details,
+                                        timeSliceLock.timeSlice);
+                                synchronized (activeJobsLock) {
```
stefannegrea: Why is this lock needed with a ConcurrentSkipListSet?

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: Good catch. Yes, it needs to be reverted. Right now it is hard coded, but that will change with [HWKMETRICS-551](https://issues.jboss.org/browse/HWKMETRICS-551).

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/SchedulerImpl.java
jsanda: I don't think the lock is needed.

stefannegrea:
Thank you @jsanda!

PR #725, opened 2017-01-12
[HWKMETRICS-523] Implement expressions for tag queries

stefannegrea:
The syntax supports:
- logical operators: AND OR
- equality operators: =  !=
- value in array operators: IN, NOT IN 
- existential conditions:
  - tag without any operator is equivalent to `= '*'`
  - tag preceeded by the NOT operator matches only instances without the tag defined
- all the values in between single quotes are treated as regex expressions
- spaces before and after equality operators are not necessary

Sample queries:
```
a1 = 'bcd' OR a2 != 'efg'
a1='bcd' OR a2!='efg'
a1 = 'efg' AND ( a2 = 'hijk' OR a2 = 'xyz' )
a1 = 'efg' AND ( a2 IN ['hijk', 'xyz'] )
a1 = 'efg' AND a2 NOTIN ['hijk']
a1 = 'd' OR ( a1 != 'ab' AND ( c1 = '*' ) )
a1 OR a2
NOT a1 AND a2
a1 = 'a' AND NOT b2
```


core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
```diff
@@ -0,0 +1,120 @@
...
+  ;
+
+
+//parser
+OR: O R;
+AND: A N D;
+EQUAL: '==';
+NOTEQUAL: '!=';
+IN: I N;
+NOTIN: N I N;
```
burmanm: I'm not huge fan of ``NIN``. Why reinvent the wheel when we could have the ``NOT IN``? Like similar commands, such as ``NOT EXISTS``. Makes the query syntax more logical to me..

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
```diff
@@ -0,0 +1,120 @@
...
+  : '[' VALUE (',' VALUE)* ']'
+  | '[' ']' // empty array
+  ;
+
+
+//parser
+OR: O R;
+AND: A N D;
+EQUAL: '==';
+NOTEQUAL: '!=';
```
burmanm: Should we have an alias with ``<>`` ?

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
```diff
@@ -0,0 +1,120 @@
...
+array
+  : '[' VALUE (',' VALUE)* ']'
+  | '[' ']' // empty array
+  ;
+
+
+//parser
+OR: O R;
+AND: A N D;
+EQUAL: '==';
```
burmanm: ``=``is often used in query languages when it comes to the equality check (it's not an assignment operator in these languages)

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
mwringe: Yeah, I think this makes more sense. I wouldn't necessarily know what NIN means if you were to put a query in front of me the first time.

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
mwringe: @stefannegrea See, I am not the only one who prefers `=` to `==` :)

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
stefannegrea: Good point that it might be confusing to new users. I will expand NIN to NOTIN, not warranted to have two tokens with space.

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
stefannegrea: I would like to keep it simple. Upper and lower case aliasing makes sense, but I would avoid having aliases outside of that.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
```diff
@@ -137,11 +136,17 @@ public void createMetric(
...
             @Suspended AsyncResponse asyncResponse,
-            @ApiParam(value = "List of tags filters", required = false) @QueryParam("tags") Tags tags) {
+            @ApiParam(value = "List of tags filters", required = false) @QueryParam("tags") Tags tags,
+            @ApiParam(value = "Tags query expression", required = false) @QueryParam("tagsQuery") String tagsQuery) {
 
-        Observable<Metric<AvailabilityType>> metricObservable = (tags == null)
-                ? metricsService.findMetrics(getTenant(), AVAILABILITY)
-                : metricsService.findMetricsWithFilters(getTenant(), AVAILABILITY, tags.getTags());
+        Observable<Metric<AvailabilityType>> metricObservable = null;
+        if (tags != null && tagsQuery == null) {
```
burmanm: What if user sends both ? Tags and tagsQuery?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricsServiceHandler.java
```diff
@@ -66,34 +66,43 @@ protected String getTenant() {
...
-            return Observable.error(new RuntimeApiError("Either metrics or tags parameter must be used"));
+        if (metricNames.isEmpty() && tags.isEmpty() && tagsQuery == null) {
+            return Observable.error(new RuntimeApiError("Either metrics or tags query parameters must be used"));
         }
-        if (!metricNames.isEmpty() && !tags.isEmpty()) {
-            return Observable.error(new RuntimeApiError("Cannot use both the metrics and tags parameters"));
+        if (!metricNames.isEmpty() && (!tags.isEmpty() || tagsQuery != null)) {
+            return Observable.error(new RuntimeApiError("Cannot use both the metrics and tags query parameters"));
+        }
+        if (metricNames.isEmpty() && (!tags.isEmpty() && tagsQuery != null)) {
```
burmanm: For the previous question, I guess it's catched here. But if it's a REST-interface limitation then it should be catched earlier.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
jshaughn: Is it really required to support both?  I guess for backward compatibility?  Perhaps 'tags' should support both formats.

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
jshaughn: +1, as much as I'm used to '==', I think '=' is more natural for most people, given just math or some DB/SQL experience.

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
jsanda: My personal preference would be ```not in``` as opposed to ```notin```

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/AvailabilityHandler.java
mwringe: Is there any way to easily determine from the format that we are using one format or another? If there could be a conflict where some queries could be matched to both formats, then I think we need to have two different named parameters

jotak:
Hey,
Just wondering about the EXISTS / NOT EXISTS that we talked earlier this week: in which order would it be placed next to the tag name, before or after? "EXISTS mytag" or "mytag EXISTS" ?

lucasponce:
In alerting we are proposing this for exists / not exists:
https://github.com/hawkular/hawkular-alerts/pull/282/files#diff-daed480b842abc261c043dfd06c7a006R262
https://github.com/hawkular/hawkular-alerts/pull/282/files#diff-5dfa51900f24fdd2f517482ac27f050fR50

I find the syntax simpler, and I don't think it clashes with any other syntax that could potentially create ambiguity in the parser.

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/metrics/ExpressionTagQueryITest.java
```diff
@@ -0,0 +1,173 @@
...
+
+        gauges = test.parse(tenantId, GAUGE, "a1 ='2' AND (a2 notin ['3'])").toList().toBlocking()
+                .lastOrDefault(null);
+        assertMetricListById(gauges, "m5");
+
+        gauges = test.parse(tenantId, GAUGE, "hostname ='web.*'").toList().toBlocking()
+                .lastOrDefault(null);
+        assertMetricListById(gauges, "mA", "mB");
+
+        gauges = test.parse(tenantId, GAUGE, "hostname ='web.*' or a1='*'").toList().toBlocking()
```
lucasponce: Minor thing but just in case it escaped into the review.

> a1 = '*'

The '*' is not a valid regexp in java, I guess you would mean '.*' to get all possible values, unless we want to shortcut this in a customized way.

I commented, because I saw it also in the first comment on the PR.

burmanm: Currently in our implementation we give the ability of using ``*`` instead of ``.*`` , there's an alias in the processor that accepts it.

lucasponce: @burmanm this '*' -> '.*' pre-preprocessor is general ? I mean 'tag*' -> 'tag.*' or just when you have the '*' -> '.*' case ?

burmanm: In current impl it's just the special case of ``*`` (as we don't do any regexp matching in that case)

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/AggregatedStatsQueryRequest.java
```diff
@@ -43,6 +43,8 @@
 
     private String tags;
 
+    private String tagsQuery;
```
jotak: tagsQuery should be added to equals & hashcode below (and eventually toString)

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
```diff
@@ -0,0 +1,129 @@
...
+  ;
+
+boolean_operator
+  : EQUAL
+  | NOTEQUAL
+  ;
+
+array_operator
+  : IN
+  | NOTIN
```
lucasponce: I suggest to maintain the NOTIN separated to simplify the management of NOT operator.
I referenced this here, too.
https://github.com/hawkular/hawkular-alerts/pull/282#issuecomment-274758744

core/metrics-core-service/src/main/antlr4/org/hawkular/metrics/core/service/tags/parser/TagQuery.g4
```diff
@@ -67,6 +69,10 @@ array_operator
   | NOT IN
   ;
 
+existence_operator
+  : NOT
```
jotak: So, is it acted that we won't use "exists" / "not exists" as existence operators, but instead just no operator for existence and just "not" for non-existence?
I'm asking because I'm implementing it in Grafana.

lucasponce: I think that's correct.
(Now both metrics and alerting are sync-ed with the expression language).
I.e.:
tagA and (not tagB or (tagC in [a, 'bc.*'] and tagD not in [1, 2]))

jotak: cool, thanks Lucas!

stefannegrea: @jotak, I updated the PR documentation to reflect the existential operators.

jshaughn:
@stefannegrea Is there anything holding the merge of this PR?  I want it merged so I can make some final changes to the Alerter PR so we can get that merged as well.

stefannegrea:
@jshaughn, I am working on one last change for this PR. Right now the PR still has two parameters for the REST API, I am working on using a single tag parameter to supports both the old and new syntax.

jshaughn:
@stefannegrea For what it's worth, in alerting we maintained the two fields in the REST API, deprecating the old one.  At the handler level we convert the old style into the new style and so the lower levels deal only with the new style.

stefannegrea:
@jshaughn, the syntax (old vs new) is very different in Metrics so it is easy to try to parse for new and revert to old if that fails.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/tags/TagsConverter.java
```diff
@@ -0,0 +1,83 @@
...
+import org.hawkular.metrics.model.param.Tags;
+
+/**
+ * The string format is a list of tags in the {@code
+ * name:value} form, comma-separated.
+ *
+ * @author Thomas Segismont
+ * @author Stefan Negrea
+ */
+public class TagsConverter {
```
jotak: That makes this class duplicated in 2 places, is the other one still needed?

stefannegrea: @jotak, good catch; the one in the REST layer is not needed anymore. I will remove it.

stefannegrea: @jotak, I made a mistake in the last comment. The converter in the API is still needed because there is still a considerable number of APIs that still accept Tags as a parameter.

jshaughn:
@stefannegrea This LGTM. I think it is ready to be squashed and then merged.  What is the plan for documenting the new query expressions?

stefannegrea:
retest this please

stefannegrea:
retest this please

stefannegrea:
retest this please

jotak:
Fine for me too

jsanda:
Minor knit. Lines 557 - 559 in MetricsServiceImpl.java can be changed from

```java
return expresssionTagQueryParser
         .parse(tenantId, metricType, tags)
         .map(tMetric -> (Metric<T>) tMetric);
```

to

```java
return expresssionTagQueryParser.parse(tenantId, metricType, tags);
```

PR #726, opened 2017-01-17
[HWKMETRICS-599] Improve performance with automatic batching

burmanm:


jsanda:
A few comments/questions. If I read correctly it looks like you are doing a batch size of 20. Have you tried different sizes? And if so, did you see worse performance with larger sizes?

Secondly, did you consider storing a map of Token --> TokenRange to avoid iterating over all token ranges, which would be 256 in the worst case.

Lastly, A single C* is a very common scenario for us. If we are going to make optimizations we might as well make optimizations for the common case too. For a single C* node, I do not see the need group statements by token range. Just do the batching. We could probably also introduce a no-op load balancing policy for single node deployments.

burmanm:
I'll answer here also..

[Token,TokenRange] map with the key being Token could result in 2^128 keys. The TokenRange.contains(x) calculates if there's something in the range (it does not store all the possible tokens).

As for single C*, I did not see a difference in excluding this counting and keeping it. Or they were too small in my tests.

stefannegrea:
@burmanm, why is the improvement in performance not identical across all metric types?

burmanm:
The processing is identical for each metric type and different PRs show different improvements (the RxJava2 PR was based on this PR also and it got different results also even for the old RxJava1 paths). @FilipB can you get stable results for the performance metrics?

FilipB:
It's described in the report. Gauges and counters msgs are sent as chunks containing 10 metrics with one datapoint. Avail and strings msgs contain 500 metrics. Bigger msgs -> bigger improvements in case of this PR. Unstability 2% is the best what I was able to get.

jsanda:
>It's described in the report. Gauges and counters msgs are sent as chunks containing 10 metrics with one datapoint. Avail and strings msgs contain 500 metrics. Bigger msgs -> bigger improvements in case of this PR. Unstability 2% is the best what I was able to get.

Thanks for the explanation. Better throughput is definitely expected with bigger messages assuming bigger means multiple data points per metric. The reason this should yield better throughput is because we batch multiple data points belonging to the same metric since they are written to the same partition. This is a very efficient way to utilize batching as Cassandra performs all of the writes atomically.

jsanda:
@burmanm how many Cassandra nodes did you test with?

@FilipB do you have any perf tests that run against multiple Cassandra nodes? That's what we really need here to see what kind of benefit this (i.e., the micro batching) yields. As mentioned in previous comments, there shouldn't be much of an appreciable difference when running with a single C* node. We should however, see gains when running with larger clusters.

jsanda:
Just a follow up on my previous comment. This PR is a performance enhancement that really is *not* relevant for single node scenarios. In order to see what kind of performance gains we get, we need tests with clusters of 2, 3, ..., N nodes. If for whatever reason the changes do not yield an appreciable difference, then it does not make sense to merge the changes because it just introduces unnecessary complexity. I am in favor of the changes, but I would like to see some numbers before we merge.

FilipB:
@jsanda it's actually 500 metric where each contains just one datapoint so that is probably why the improvement is not as big as expected. I don't have any env with more cassandra nodes prepared. It would take some time + I need to prepare tests with lot of datapoints per metrics to see real benefit.

jsanda:
> I need to prepare tests with lot of datapoints per metrics to see real benefit.

We don't need a lot of data points per metric. To be honest, I think it is better, at least for now, to focus on testing with a single data point per metric. That is how metrics are inserted in OpenShift so it makes sense to follow that same pattern. I think it is the same for MiQ as well, but I need to double check.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -632,86 +655,165 @@ protected void initPreparedStatements() {
                 .concatWith(rxSession.executeAndFetch(findAllMetricsInDataCompressed.bind()));
     }
 
-    @Override
-    public Observable<Integer> insertGaugeData(Metric<Double> gauge) {
-        return insertGaugeData(gauge, -1);
+    /*
+     * Applies micro-batching capabilities by taking advantage of token ranges in the Cassandra
+     */
+    private Observable.Transformer<BoundStatement, Integer> applyMicroBatching() {
```
jsanda: Right now our insert patterns are such that we do one data point per metric.  In the event that we start doing multiple data points per metric per request, will this method group the corresponding inserts into the same batch statement?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -632,86 +655,165 @@ protected void initPreparedStatements() {
...
+                .groupBy(b -> {
+                    ByteBuffer routingKey = b.getRoutingKey(ProtocolVersion.NEWEST_SUPPORTED,
+                            codecRegistry);
+                    Token token = metadata.newToken(routingKey);
+                    for (TokenRange tokenRange : session.getCluster().getMetadata().getTokenRanges()) {
+                        if (tokenRange.contains(token)) {
+                            return tokenRange;
+                        }
+                    }
+                    throw new RuntimeException("Unable to find any Cassandra node to insert token " + token.toString());
```
jsanda: I realize that this situation shouldn't happen, but wouldn't it be better to log a warning and write to any node?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
burmanm: It should not matter, because it's working at the level of BoundStatements.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
burmanm: Yeah, I'll change to that.

burmanm:
I tested this PR with 1 and 2 nodes, but running on the same machine. In any case, this applies to the single node scenario as well and is the performance difference I talked about in the mailing list. So without REST-interface the performance was >5x in a single node case.

FilipB:
I did testing on following set up: 3 cassandra nodes each 2CPUs, 4GB of memory, replication factor 1. h-metrics with 2 CPUs and 4 GB of memory.
Tested following msg sizes: 1, 10, 500 and 5000. Each metric in the json contains only one datapoint.
This PR improvements against current master are: 6%, 11%, 42%, 164%. Bigger msgs -> bigger improvement

jsanda:
Thanks for sharing the test results Filip! I think we are good to go.

PR #727, opened 2017-01-17
HWKMETRICS-566 Power up the external alerter

jshaughn:
- Introduces a new mechanism for defining conditions using
  flexible expressions with embedded stats queries.
- Replaces the previous expression syntax.

Note:
- Introduces use of a new 3rd party component, EvalEx.  A nifty
  tool for handling expression evaluation.
  - https://github.com/uklimaschewski/EvalEx
- Added 'samples' to AvailabilityBucketPoint
  - good for consistency, and useful for determining upCount
- Add Percentile.toString() for better debugging


FilipB:
retest this please

jshaughn:
@stefannegrea @lucasponce This is now ready, I think, for a code review prior to merge.  Since the meeting I've cleaned things up a bit and incorporated feedback:
- Added support for EvalType.EACH which performs evaluations against each metric individually, as opposed to EvalType.ALL (the default) which combines the datapoints from all metrics and performs a single evaluation.
-- Note, EvalType was formerly named ResultType.
-- Note, I moved EvalType from Query level to ConditionExpression level, which I think is more straightforward and less confusing than supporting different approaches for different queries used in the eval expression.
- I migrated from having the Alerter generate Data to generating (non-persisted) Events when an evaluation is true.  As discussed with @lucasponce , this is a better semantic for External Alerters.  It also avoids an issue with alerting tossing out Data for the same DataId, when timestamps are too near together.
- I also added better handling for when queries return no data.  In this situation the evaluation simply resolves to false and things continue (debug logging as well, if enabled).

Please take a look, time permitting.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
```diff
@@ -0,0 +1,312 @@
...
+        log.debugf("eval [%s] produced [%s] with variables %s", eval, expression.getOriginalExpression(), queryVars);
+
+        // Do a test evaluation to validate the expression
+        try {
+            for (String var : queryVars.keySet()) {
+                this.expression.setVariable(var, "1");
+            }
+            this.expression.eval();
+        } catch (Exception e) {
+            throw new ExpressionException("Invalid eval [" + eval + "]: " + e.getMessage());
```
lucasponce: Perhaps a IllegalArgumentException instead to propagate this one com.udojava.evalex.Expression.ExpressionException ?

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
```diff
@@ -0,0 +1,312 @@
...
+import org.hawkular.metrics.model.AvailabilityBucketPoint;
+import org.hawkular.metrics.model.BucketPoint;
+import org.hawkular.metrics.model.NumericBucketPoint;
+import org.hawkular.metrics.model.Percentile;
+import org.jboss.logging.Logger;
+
+import com.udojava.evalex.Expression;
+import com.udojava.evalex.Expression.ExpressionException;
+
+public class ConditionEvaluator {
```
lucasponce: If ConditionEvaluator is dealing with the expression, perhaps a BNF representation of valid expressions here is good.
It has not to be complete but some lines would definitely help.
For example, I put a similar one in this other alerter:
https://github.com/hawkular/hawkular-alerts/blob/master/hawkular-alerts-engine-extensions/hawkular-alerts-events-aggregation/src/main/java/org/hawkular/alerts/extensions/Expression.java#L31

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
```diff
@@ -0,0 +1,312 @@
...
+                case notUpCount:
+                    return data.getNotUpCount().doubleValue();
+                case notUpDuration:
+                    return data.getNotUpDuration().doubleValue();
+                case upCount:
+                    return data.getUpCount().doubleValue();
+                case uptimeRatio:
+                    return data.getUptimeRatio().doubleValue();
+                case samples:
+                    return data.getSamples() * 1.0;
```
lucasponce: I normally used the cast operator for this. I guess this is for converting to double, isn't it ?

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
```diff
@@ -0,0 +1,312 @@
...
+
+    private String replaceQueriesWithVariables(String eval) {
+        if (null == queryVars) {
+            queryVars = new HashMap<>();
+        }
+        queryVars.clear();
+
+        // First, collect the queries
+        List<String> queries = new ArrayList<String>();
+        Matcher m = Pattern.compile("q\\(.*?\\)").matcher(eval);
```
lucasponce: Minor optimization.
If this one pattern is going to be used often you could make it static as part of the class.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
```diff
@@ -0,0 +1,312 @@
...
+
+        public QueryFunc(String queryName, String function) {
+            super();
+            this.queryName = queryName.trim();
+            this.function = function.trim();
+        }
+
+        public QueryFunc(String query) {
+            super();
+            Matcher m = Pattern.compile("q\\((.*?),\\s*(.*?)\\)").matcher(query);
```
lucasponce: Same here. Pattern can be static to the class.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionExpression.java
```diff
@@ -0,0 +1,657 @@
...
+
+public class ConditionExpression {
+
+    /**
+     * Determines whether the eval expression is performed on ALL or EACH of the metrics (either specified
+     * explicitly or via the tag query.  ALL means that the metric data points will be combined, then aggregated, and
+     * the evaluation is performed once (per run of the ConditionExpression).  EACH means that data for each
+     * metric will be aggregated separately and the evaluation is performed once per metric.  Note that if multiple
+     * queries are involved in the eval expression that only metrics common to each query will be evaluated.
+     * The default is ALL.
```
lucasponce: I have read the comment, but I have some fear that perhaps I could misinterpreted the behavior, perhaps a couple of lines with examples of ALL/EACH to clarify could help.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionExpression.java
```diff
@@ -0,0 +1,657 @@
...
+     */
+    public enum EvalType {
+        ALL, EACH
+    }
+
+    private static final Logger log = Logger.getLogger(ConditionExpression.class);
+
+    /** 10mn */
+    public static final String DEFAULT_FREQUENCY = "10mn";
```
lucasponce: "mn" means "minutes" here ? Why don't use just "m" ? In other DSL, the some-sort-standard prefixes are d/h/m/s etc.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
jshaughn: +1, replaced in two places so as not to propagate the 3rd party Exception class.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
```diff
@@ -0,0 +1,554 @@
...
+@Singleton
+public class ConditionManager {
+    private final Logger log = Logger.getLogger(ConditionManager.class);
+
+    private static final String TAG_EXTERNAL_CONDITION_NAME = "HawkularMetrics";
+    private static final String TAG_EXTERNAL_CONDITION_VALUE = "MetricsCondition";
+
+    private static final Integer THREAD_POOL_SIZE = 20;
+
+    ScheduledThreadPoolExecutor expressionExecutor;
```
lucasponce: Note to self.
ThreadPools are growing.
I guess that at some point it could be nice to have a wrapper for these classes and put more descriptive names to them.
Then when we dump a JVM we can trace better the threads involved.

Just a side comment, no action required, but it is not first time that I think on it.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
jshaughn: Good point, a couple of the new classes were missing class-level jdoc.  The supported Expression syntax is more fully defined at the EvalEx page, I have created class jdoc that points to it, as well as adds more info.  I also added class jdoc for ConditionExpression.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
jshaughn: Multiplying by a double literal always returns a double, so a cast is not necessary.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
jshaughn: +1, create a static Pattern.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionEvaluator.java
jshaughn: +1, created a static Pattern.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
```diff
@@ -0,0 +1,554 @@
...
+            Map<String, String> preparedCondition = null;
+            try {
+                preparedCondition = evaluator.prepare(queryResults);
+            } catch (Exception e) {
+                log.debugf("Could not prepare evaluation, Skipping due to [%s]", e.getMessage());
+            }
+
+            if (evaluator.evaluate()) {
+                try {
+                    Data externalData = Data.forString(externalCondition.getTenantId(),
```
lucasponce: So, I am not 100% sure if I understand well some details of the logic.
But in EACH mode, the ExternalCondition will expect and Event and in ALL mode a Data to trigger the Alert.
Am I right ?

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/alerting/ExternalAlerterITest.groovy
```diff
@@ -61,10 +68,12 @@ class ExternalAlerterITest extends AlertingITestBase {
         resp = client.post(path: "triggers", body: triggerTestAvg)
         assertEquals(200, resp.status)
 
-        // ADD external metrics avg condition. Note: alerterId must be "HawkularMetrics"
+        // Note: alerterId must be "HawkularMetrics"
         // Average over the last 1 minute > 50, check every minute
+        Query qNow = Query.builder("qNow").metric(metricId).duration("1mn").build();
```
lucasponce: I would also add tests over the REST API through plain JSON too.
Perhaps the same but with plain text.
There are not support for extensions with Java clients, and also, I am not 100% if python/ruby clients are tested against ExternalCondition (probably not).
So, these JSON tests will be the closest example to a live example.
It would help to point people into the feature.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionExpression.java
jshaughn: I have added a more verbose comment to hopefully make things more clear.  This same information will be in the documentation.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionExpression.java
jshaughn: "mn" is the suffix used by Metrics' Duration. I question their decision as well but  I am only conforming to their documented Duration type: http://www.hawkular.org/docs/rest/rest-metrics.html#_custom_string_formats.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
jshaughn: +1, they are everywhere.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
jshaughn: No, the reason you are confused is that this is a bug, leftover from the conversion from Data to Event sending.  Thanks for pointing it out. Now the code sends Events consistently.

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/alerting/ExternalAlerterITest.groovy
jshaughn: I'm not exactly sure what that test would exercise.  This code is already using REST to post the ExternalCondition.  Are you concerned that python/ruby clients may not be able to post an ExternalCondition using a JSON string for the expression value?

lucasponce: I know that groovy is unbeatable for these tests where Java objects are "JSON-lized" under the hood in a pretty straightforward way.
But experience showed us when someone else needs to use this feature, and we point them to these examples, they are not very useful, as they need to access via Java API.
I think these tests are great and we should continue using but what I do is to introduce some additional test where instead of an object we define the json representation in raw format.
That helps to catch up with the technology.

jshaughn: OK, you are talking about this from the perspective of a code example.  OK, I can create one like that.  In the formal documentation I have a JSON example as well.

jshaughn:
@lucasponce Muchas gracias for the review!   I have pushed a commit with the changes (all but the requested JSON ITest example).

jshaughn:
With the review from @lucasponce and the follow-up commits I think I can remove the DoNotMerge label.  @stefannegrea, or anyone, it is still open for more review comments or questions if you have them.

jshaughn:
Re-added DNM, I want to use only the the new tag query string, so waiting on PR 725...

jshaughn:
@stefannegrea @lucasponce, I have removed the DNM label.  I have rebased the PR on master, applied final changes wrt the new tag query work, and sqashed.  Assuming travis is happy and if there are no further comments, I think this can be merged along with the associated documentation PR (https://github.com/hawkular/hawkular.github.io/pull/265).

stefannegrea:
Thank you @jshaughn!

PR #728, opened 2017-01-17
[HWKMETRICS-572] Make admin tenant configurable. The admin tenant is …

stefannegrea:
…used for persisting internal metrics.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -33,11 +33,11 @@ import com.google.common.base.Preconditions;
  */
 public class MetricNameService implements MetricFilter {
 
-    public static final String TENANT_ID = "admin";
 
     private static final String separator = ":";
 
-    private String hostname;
+    private final String adminTenant;
+    private final String hostname;
 
     public MetricNameService() {
         try {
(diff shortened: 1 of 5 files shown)


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
```diff
@@ -33,28 +33,45 @@
...
+    private final String hostname;
 
     public MetricNameService() {
         try {
             hostname = InetAddress.getLocalHost().getHostName();
         } catch (UnknownHostException e) {
             throw new RuntimeException("Failed to get hostname", e);
         }
+
+        this.adminTenant = "admin";
```
mwringe: Is there any reason why we need to hard code the 'admin' default here when we are already setting the default in ConfigurationKey.java?

Do we need this constructor? or can we get rid of it and rely on the other constructions which require the admin tenant to be passed?

mwringe:
Other than my one comment, LGTM

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/MetricNameService.java
stefannegrea: I think it might be needed for testing purposes. @jsanda, is this constructor important?

jsanda: @stefannegrea you might want to rebase on master because the version of this class is out of date a bit. The constructor is used for testing, but admin tenant is still hard coded, 

```java
public static final String TENANT_ID = "admin";
```

That will have to change.

stefannegrea: @jsanda, do you mind clarifying the last comment? The line you referenced has been completely removed by this PR. Is it hardcoded somewhere else?

jsanda: Sorry, I misunderstood the previous comments and diff. The no-arg constructor gets called if neither the ```hawkular.metrics.hostname``` system property nor the ```HOSTNAME``` environment variable are set.

jsanda: I just realized that my previous comment is based on changes in a branch I am working on. Sorry for the confusion. In master, the no-arg constructor is only used in tests; however, that is going to change.

PR #740, opened 2017-01-19
Move to Alerting 1.3.3 for bug fixes.

jshaughn:


jshaughn:
@stefannegrea the build failure does not seem related to the update of alerting, what do you think?

burmanm:
That's a bug in the MixedMetricsITest, it's leaking resources from one test to another (I've had that same happen on my machine and I did quickly debug it earlier.. but then it stopped showing up on my machine)

stefannegrea:
Thank you @jshaughn!

PR #742, opened 2017-01-20
[HWKMETRICS-583] some small changes to job locks

jsanda:
The job lock name now includes the hostname and the queries for acquiring and
renewing locks have been modified so that a job scheduler instance can
reacquire a lock it already holds.

This commit also removes the call to reschedule a job immediately when it falls
behind schedule. I removed it because it is 1) it an optimization that is not
necessary and 2) it was causing some inconsistent behavior due to a race
condition when the indexes are read and updated.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
@@ -399,7 +399,7 @@ public class MetricsServiceLifecycle {
             if (metricsReportingHostname == null) {
                 metricNameService = new MetricNameService(adminTenant);
             } else {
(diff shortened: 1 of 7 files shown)


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/MetricsServiceLifecycle.java
```diff
@@ -399,7 +399,7 @@ private void startMetricsService() {
             if (metricsReportingHostname == null) {
                 metricNameService = new MetricNameService(adminTenant);
             } else {
-                metricNameService = new MetricNameService(metricsReportingHostname, adminTenant);
+                metricNameService = new MetricNameService(metricsReportingHostname);
```
stefannegrea: We need the admin tenant here to make sure metrics are stored with the correct tenant.

stefannegrea:
Thank you @jsanda!

PR #744, opened 2017-01-23
[HWKMETRICS-585] Merge dropwizard reporter in metrics

jotak:
And use hawkular-metrics-parent as new maven parent, adapt pom to remove items already managed in parent.

Note about travis: I've let the existing travis config, but I'm not sure if it's going to work as is since it's now  in a subfolder. Basically, this config pulls & start container for running metrics in order to execute integration tests. It was copied from the ruby client travis scripts.


rhqci:

clients/common/pom.xml
```diff
@@ -32,6 +32,19 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.glassfish</groupId>
```
jotak: I'm not 100% sure if it's the right place for that (in clients/common, scope provided)... is it ok?

jotak:
@stefannegrea I've switched to embedded hawkular/wildfly for integration tests

jsanda:
Sorry for the delay in getting to this. Thanks for doing this @jotak!

PR #749, opened 2017-01-31
[HWMETRICS-570] initial support for automatic lock renewal

jsanda:
The idea is to prevent a job execution lock from expiring while the job is
running without placing too much extra load on Cassandra. job execution locks
will automatically renew once more than half the time to their expiration has
passed. If a lock is set to expire in one hour for example, then it will renew
in 30 minutes.

Single execution jobs locks have a hard coded timeout of 30 minutes. That value
is completely arbitrary. I did not want a value to small resulting in lots of
lock renewal queries.

Repepating job lock timeouts are based on the job interval. The timeout is 25%
more than the interval. The compression job runs every two hours, so the lock
timeout will be 2.5 hours and will renew in 75 minutes.


stefannegrea:
Thank you @jsanda!

PR #752, opened 2017-02-02
[HWKMETRICS-591] Incorrect token authorization takes longer than correct..

burmanm:
Do not retry token authorization if connection was successful, but the response was not what we wanted.

@mwringe

### DIFF
--- a/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/TokenAuthenticator.java
+++ b/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/TokenAuthenticator.java
@@ -197,9 +197,7 @@ class TokenAuthenticator implements Authenticator {
         // this method returns, but we need to wait for Kubernetes' master response.
         serverExchange.dispatch();
         XnioIoThread ioThread = serverExchange.getIoThread();
-        ConnectionPool connectionPool = connectionPools.computeIfAbsent(ioThread, t -> {
-            return new ConnectionPool(connectionFactory, componentName);
-        });
+        ConnectionPool connectionPool = connectionPools.computeIfAbsent(ioThread, t -> new ConnectionPool(connectionFactory, componentName));
         PooledConnectionWaiter waiter = createWaiter(serverExchange);
         if (!connectionPool.offer(waiter)) {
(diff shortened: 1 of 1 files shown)


containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/TokenAuthenticator.java
```diff
@@ -490,17 +487,18 @@ protected void stringDone(String body) {
...
                     onRequestResult(serverExchange, connection, allowed);
                 } catch (IOException e) {
-                    onRequestFailure(serverExchange, connection, e);
+                    onRequestFailure(serverExchange, connection, e, true);
                 }
             } else {
                 IOException e = new IOException(StatusCodes.getReason(clientExchange.getResponse().getResponseCode()));
-                onRequestFailure(serverExchange, connection, e);
+                // Retries will not help to solve the issue
+                onRequestFailure(serverExchange, connection, e, false);
```
mwringe: This gets call if the response is anything other than 201. And there are some responses in which we may want to retry (eg 503) and certain ones in which we don't want to retry (eg 400, 401, 403, etc).

I think it makes sense to still retry in some situations which would need to be addressed here.

burmanm: I tried to search the Kubernetes documentation on which return codes are available, but the documentation in this sense is useless. 503 is service unavailable, transforming it to 500 sounds like a proper choice in my opinion? It's not necessarily quickly available again.

mwringe: I don't think we should worry too much about what the current kubernetes implementation is doing, we might not even be hitting the exact endpoint but some proxy or a load balancer.

I think this should just be done with best effort and based on the definitions of what those status codes typically means (eg 503 typically is a temporary problem, but not always). Even if we just retry with a 503 and not retry with the others I think would be a good solution.

If we have a need to tweak this later if deemed necessary.

mwringe:
Awesome, LGTM

stefannegrea:
Thanks @burmanm!

PR #754, opened 2017-02-03
Minor README verbiage updates.

stefannegrea:


README.adoc
```diff
@@ -13,21 +13,20 @@ endif::[]
 
 == About
 
-Hawkular Metrics is the metric data store for the http://www.hawkular.org/[Hawkular] project.
-It can also be used independently.
+Hawkular Metrics is the metric data store engine part of http://www.hawkular.org/[Hawkular] community.
```
burmanm: `store` or `storage` ? Or do you want to separate the fact that Cassandra is the "real storage" ?

README.adoc
```diff
@@ -112,7 +111,7 @@ https://lists.jboss.org/mailman/listinfo/hawkular-dev[Hawkular developer mailing
 
 We love contributions and pull-requests :-)
 
-To meet the guys, join us on irc://irc.freenode.net/#hawkular[Freenode IRC] in channel #hawkular.
+To chat, join us on irc://irc.freenode.net/#hawkular[Freenode IRC] in channel #hawkular.
```
burmanm: If possible, instead of me creating one more PR about these changes, could you modify this PR to include the following changes also:

-It can also be used independently.
+It can also be used independently or with https://github.com/hawkular/hawkular-alerts[Alerts].

And to contribute part:

+To file an issue, use the http://issues.jboss.org/browse/HWKMETRICS[Hawkular-Metrics JIRA].

(this was actually asked in the IRC, where to report bugs and we don't have a link anywhere)

stefannegrea: @burmanm, I will add the JIRA link, good suggestions. However, I am not sure I want to add the independently statement that you propose because we no longer advertise Metrics without Alerting.

README.adoc
stefannegrea: @burmanm, good catch; will fix.

PR #755, opened 2017-02-05
[HWKMETRICS-556] use TWCS for data_compressed table

jsanda:
We are also switching back to STCS for the data table because we are now doing
range deletes as part of the compression job. Eventually I think we'll switch
over to using what are essentially temporary tables for raw data that we just
drop once we are done writing to them and have compressed the data.


jsanda:
retest this please

stefannegrea:
Thank you @jsanda!

PR #757, opened 2017-02-05
[HWMETRICS-595] do not query past TTL to avoid reading tombstones

jsanda:


jsanda:
I need to fix/add the logic in the code I added in MetricsServiceImpl.java. I need to look at both the ```start``` and ```end``` parameters. There are also a number of failing tests that need to be fixed. These tests are broken because they use timestamps start at offsets from zero with the default data retention. The tests need to be updated to use meaningful timestamps.

jsanda:
retest this please

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -1045,6 +1051,22 @@ private int getTTL(MetricId<?> metricId) {
...
     }
 
+    /**
+     * We do not want to query past the TTL. Doing so may cause Cassandra to read and load into heap space a large
+     * number of tombstones that could substantial increase read latencies and even lead to OOMEs in Cassandra. See
+     * https://bugzilla.redhat.com/show_bug.cgi?id=1416850 for more info.
+     */
+    private DateRange getDateRange(MetricId<?> metricId, long start, long end) {
+        long ttl = Duration.standardSeconds(getTTL(metricId)).getMillis();
+        if ((System.currentTimeMillis() - end) >= ttl) {
```
burmanm: Could we use the DateTime tools instead? To account for proper timezone handling. This uses current time, while we use UTC in the storage + TTL.

jsanda: The code here uses unix timestamps - the number of milliseconds since the unix epoch. Timestamps are not affected by time zones, so this should be fine, right?

PR #762, opened 2017-02-10
Prototype of using RxJava2 in Gauge write path

burmanm:
Branched from PR #726


burmanm:
(don't close this, I will next try RxJava 2.1 version)

PR #765, opened 2017-02-12
Fix JMH benchmarks to work with current version of Hawkular-Metrics.

burmanm:
Refactoring of the packages in Hawkular-Metrics did not fix all the packages (such as JMH benchmarks) which are not compiled in the normal clean install process.


stefannegrea:
Thank you @burmanm!

PR #769, opened 2017-02-15
HWKMETRICS-608

jsanda:
Update to TWCS if it is available on Cassandra's classpath.

### DIFF
--- a/.travis.install.cassandra.sh
+++ b/.travis.install.cassandra.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+# Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
 # and other contributors as indicated by the @author tags.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -22,7 +22,7 @@ set -xe
 cd "${HOME}"
 
 C_MAJOR="3"
-C_MINOR="7"
+C_MINOR="9"
 
 # Find the closest Apache mirror
 APACHE_MIRROR="$(curl -sL https://www.apache.org/dyn/closer.cgi?asjson=1 | python -c 'import sys, json; print json.load(sys.stdin)["preferred"]')"
--- a/core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
+++ b/core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
@@ -92,9 +92,6 @@ public class SchemaService {
             if (resultSet.one().getString(0).equals(TWCS_CLASS)) {
                 return;
             }
-            if (resultSet.one().getString(0).equals(TWCS_CLASS)) {
-                return;
-            }
         } else {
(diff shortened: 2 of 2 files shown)


.travis.install.cassandra.sh
```diff
@@ -22,7 +22,7 @@ set -xe
 cd "${HOME}"
 
 C_MAJOR="3"
-C_MINOR="7"
+C_MINOR="9"
```
mwringe: are we skipping 8? or does this include both what is in 8 and 9?

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -61,6 +66,62 @@ public void run(Session session, String keyspace, boolean resetDB) {
...
+     * so fails, we just log a warning so that start up is not disrupted.
+     *
+     * See HWKMETRICS-608 and HWKMETRICS-603 for more info.
+     */
+    private void maybeUseTWCS(Session session, String keyspace) {
+        String cassandraVersion = getCassandraVersion(session);
+        String getCompactionClass;
+        ResultSet resultSet;
+
+        if (cassandraVersion.startsWith("2")) {
```
mwringe: Will this bite us when Cassandra 20 is released?

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -61,6 +66,62 @@ public void run(Session session, String keyspace, boolean resetDB) {
...
+            resultSet = session.execute(getCompactionClass);
+            if (resultSet.isExhausted()) {
+                logger.warn("Unable to determine the compaction strategy for the data table due to empty result set");
+                return;
+            }
+            if (resultSet.one().getString(0).equals(TWCS_CLASS)) {
+                return;
+            }
+            if (resultSet.one().getString(0).equals(TWCS_CLASS)) {
+                return;
```
mwringe: Why is this code being run twice?

jsanda: I need to fix that.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
jsanda: I don't think so because this change is only being made in the release/0.18.0 branch. That version of metrics is compatible with Cassandra 3.x, Cassandra 2.2.x, and that's it.

.travis.install.cassandra.sh
burmanm: Cassandra 3.8 was never released (installation package for 3.8 is the same as 3.9) as it had some serious bugs.

mwringe: Ah, sorry, I wasn't paying attention to that this was doing before.

@jsanda is there any reason why we are using 3.9 here instead of 3.0.x? And another question, if this is based on release/0.18.0, shouldn't we be using Cassandra 2.2.x?

PR #770, opened 2017-02-28
[HWKMETRICS-589] Add REST API request logging

jsanda:
There are two properties that control request logging one of which is for writes. I felt it necessary to have a separate property for logging writes because it could result in lots of extra logging.

The output looks like:

INFO  [org.hawkular.metrics.api.jaxrs.util.RequestLoggingFilter] (default task-49)
REST API request:
--------------------------------------
path: /metrics
segments: [metrics]
method: GET
query parameters: {type=[availability]}
headers: [Accept=application/json, application/javascript, text/javascript,Connection=Keep-Alive,Hawkular-Tenant=T9a116f18-28cf-41b3-8ff8-c9752ac60e26232,Host=127.0.0.1:55977]


stefannegrea:
Looks like a configured log level will trigger logging. It is prudent to turn this off by default and let users explicitly turn this on for security reasons.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/RequestLoggingFilter.java
```diff
@@ -0,0 +1,115 @@
...
+    private void logRequest(Logger.Level level, ContainerRequestContext requestContext) {
+        String request =
+                "\n" +
+                "REST API request:\n" +
+                "--------------------------------------\n" +
+                "path: " + requestContext.getUriInfo().getPath() + "\n" +
+                "segments: " + requestContext.getUriInfo().getPathSegments() + "\n" +
+                "method: " + requestContext.getMethod() + "\n" +
+                "query parameters: " + requestContext.getUriInfo().getQueryParameters() + "\n" +
+                "headers: " + requestContext.getHeaders() + "\n" +
```
stefannegrea: I think we need to exclude the admin token here any other authentication headers.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/RequestLoggingFilter.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private boolean isInsertDataRequest(ContainerRequestContext requestContext) {
+        List<PathSegment> pathSegments = requestContext.getUriInfo().getPathSegments();
+        return (requestContext.getMethod().equals("POST")) && !pathSegments.get(pathSegments.size() - 1).getPath()
+                .equals("query");
+    }
+
+    private void logRequest(Logger.Level level, ContainerRequestContext requestContext) {
+        String request =
```
stefannegrea: Because this is going to be a large string it is better to use something more efficient for constructing the string.

jsanda: String concatenation with ```+``` is done using StringBuilder.

stefannegrea: Good observation, no looping involved here. Please ignore my comment.

PR #771, opened 2017-02-28
[HWKMETRICS-611] Add dot to the list of supported characters for simp…

stefannegrea:
…le text tokens. This allows using the query language with tags that have dot characters in the tag name.


core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/metrics/ExpressionTagQueryITest.java
```diff
@@ -121,6 +121,38 @@ public void tagValueSearch() throws Exception {
...
+
+        counters = test.parse(tenantId, COUNTER, "a2.label1 = 5.6.7").toList().toBlocking()
+                .lastOrDefault(null);
+        assertMetricListById(counters, "c3");
+
+        counters = test.parse(tenantId, COUNTER, "a2.label1").toList().toBlocking()
+                .lastOrDefault(null);
+        assertMetricListById(counters, "c2", "c3");
+
+        counters = test.parse(tenantId, COUNTER, "a2.label1 = '5.*'").toList().toBlocking()
```
burmanm: This would match 5abc also as ``.`` is a regexp match for anything but not existant.

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/rest/ExpressionTagQueryITest.groovy
```diff
@@ -170,6 +170,30 @@ class ExpressionTagQueryITest extends RESTTest {
...
+      assertMetricListById(response.data, 'm6')
+
+      response = hawkularMetrics.get(path: "metrics",
+        query: [tags: "a.b = 'c.d'"],
+        headers: [(tenantHeaderName): tenantId])
+      assertEquals(200, response.status)
+      assertMetricListById(response.data, 'm6')
+
+      response = hawkularMetrics.get(path: "metrics",
+        query: [tags: "a.b = 'c.*'"],
```
burmanm: Same here, if the intention was to test ``startsWith(c.)`` then this doesn't do that, but instead of ``startsWith(c)``

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/metrics/ExpressionTagQueryITest.java
stefannegrea: The intention here was to match anything that starts with `5`, including `5` as value.

integration-tests/rest-tests-jaxrs/src/test/groovy/org/hawkular/metrics/rest/ExpressionTagQueryITest.groovy
stefannegrea: The intention here was to match anything that starts with `c`. 

This is a negative test to make sure that dot is still interpreted correctly for the complex value part of the ANTLR syntax. I added the dot only to the simple value parsing, so wanted to make sure that complex values are still parsed correctly.

PR #772, opened 2017-03-01
[HWKMETRICS-589] Add REST API request logging

jsanda:
There are two properties that control request logging one of which is for writes. I felt it necessary to have a separate property for logging writes because it could result in lots of extra logging.

The output looks like:

```
INFO  [org.hawkular.metrics.api.jaxrs.util.RequestLoggingFilter] (default task-49)
REST API request:
--------------------------------------
path: /metrics
segments: [metrics]
method: GET
query parameters: {type=[availability]}
Tenant: T9a116f18-28cf-41b3-8ff8-c9752ac60e26232
```


stefannegrea:
Thank you @jsanda!

PR #777, opened 2017-03-14
[HWKMETRICS-623] Add GZIP annotation to all REST endpoints to enable …

stefannegrea:
…content encoding when the client supports it. The annotation enables content encoding by default.


jsanda:
retest this please

PR #778, opened 2017-03-15
[HWKMETRICS-625] Change timestamp fetching to be an optional parameter

burmanm:
Currently we fetch all the min and max timestamps of data for each requested metric definition. This changes the behavior for multi-metric-definition requests by making it optional parameter and requiring "timestamps: true" to be set as a query parameter should you wish to get them.


jshaughn:
Any issue with changing the current API behavior?  It's fine by me, just noticed in passing that this seems to mildly change behavior.

stefannegrea:
Do we know who requested the feature and if it is actively in use?

stefannegrea:
@jshaughn, the call to get the list of metrics is not fast enough when hundreds of thousands of metrics are stored in a single tenant.

burmanm:
@stefannegrea I think this is used by the ManageIQ (Ruby client). So we need to make a change there.

jsanda:
I will create a ticket for ruby client.

jsanda:
I don't know why MinMaxTimestampTransformer lives in the metrics-api-jaxrs modules instead of metrics-core-service. We can eliminate a bunch of duplicate code in the JAX-RS handlers if we push the ```timestamps``` parameter and corresponding logic down into MetricsService.findMetrics. @burmanm would you mind making the change?

PR #779, opened 2017-03-16
[HWKMETRICS-565]

jsanda:
Update endpoints for fetching metric definitions to chunk responses rather than building the entire response in memory before sending it to the client.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
@@ -16,14 +16,15 @@
  */
 package org.hawkular.metrics.api.jaxrs.handler.observer;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import javax.servlet.AsyncContext;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.hawkular.jaxrs.filter.cors.Headers;
 import org.hawkular.metrics.model.ApiError;
 import org.hawkular.metrics.model.Metric;
 import org.jboss.logging.Logger;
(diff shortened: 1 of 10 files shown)


jsanda:
I am not explicitly setting the ```Transfer-Encoding:chunked``` header, but I read that if neither the Content-Length nor the Transfer-Encoding header is set, then the HTTP server will implicitly add the chunked header. I have not checked the HTTP spec to see if that is required; so, it might be worth it for me to update this PR to explicitly set the Transfer-encoding header.

jsanda:
@jotak I guess I missed your [comment/question](https://issues.jboss.org/browse/HWKMETRICS-554?focusedCommentId=13334853&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13334853) on HWKMETRICS-554. I'll be pretty blunt here. I am not a big fan of JAX-RS. It does not suit our needs, i.e., async and streaming, very well, but we are forced to use it. JAX-RS does not support streaming with AsyncResponse, at least not directly. My original streaming implementation with NamedDataPointObserver involved some blocking, which we want to avoid. @tsegismont refactored it to the version up to what we had prior to HWKMETRICS-554. It made the streaming blocking. The changes you made in HWKMETRICS-554 undid the streaming/chunking. I guess we need to sync up to make sure we have the proper CORS information while still streaming.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public MetricObserver(HttpServletRequest request, HttpServletResponse response, ObjectMapper mapper) {
+        this.request = request;
+        this.response = response;
+        this.mapper = mapper;
+        count = new AtomicInteger();
+        jsonOutputStream = new ByteArrayOutputStream();
+        try {
+            generator = mapper.getFactory().createGenerator(response.getOutputStream(), JsonEncoding.UTF8);
+            generator.writeStartArray();
```
burmanm: Shouldn't this happen at the onNext (when the count.incrementAndGet() == 1) ? Now we write to this generator even if we're not going to write anything else to it (only 204 code).

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
```diff
@@ -0,0 +1,142 @@
...
+    @FunctionalInterface
+    private interface WriteValue<T> {
+        void call(Metric<T> dataPoint) throws IOException;
+    }
+
+    private final HttpServletRequest request;
+    private final HttpServletResponse response;
+    private final JsonGenerator generator;
+    private final ObjectMapper mapper;
+    private AtomicInteger count;
```
burmanm: Why the AtomicInteger when there's nothing protecting the onNext from writing concurrently to the same generator?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public void onCompleted() {
+        try {
+            if (count.get() == 0) {
+                response.setStatus(HttpServletResponse.SC_NO_CONTENT);
+            } else {
+                generator.writeEndArray();
+            }
+            generator.close();
```
burmanm: This is never called if we encounter an error (onError terminates the stream, so onCompleted is never called)

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
```diff
@@ -0,0 +1,142 @@
...
+    public void onError(Throwable e) {
+        logger.warn("Fetching data failed", e);
+        try {
+            if (count.get() == 0) {
+                response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+                response.setHeader("Content-Type", "application/json");
+                ApiError apiError = new ApiError(Throwables.getRootCause(e).getMessage());
+                mapper.writeValue(response.getOutputStream(), apiError);
+            } else {
+                generator.close();
```
burmanm: Should this be on the finally block?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/observer/MetricObserver.java
burmanm: Same here, finally?

PR #780, opened 2017-03-16
[HWKMETRICS-623] Add integration tests to verify that the content ret…

stefannegrea:
…urned is GZIP encoded.


jsanda:
retest this please

PR #781, opened 2017-03-20
HWKMETRICS-630 Alert quietCount option

jshaughn:
@lucasponce  This feature may be OK as is but we should keep it open until things are clear for hwkalerts-240.

### DIFF
--- a/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
+++ b/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
@@ -279,6 +279,9 @@ public class ConditionManager {
         private ExternalCondition externalCondition;
         private ConditionExpression expression;
 
+        // Map of quiet metric names to count of remaining quiet evals
+        private transient Map<String, Integer> quietMap;
+
         public ExpressionRunner(MetricsService metrics, AlertsService alerts, Trigger trigger,
                 ExternalCondition externalCondition,
                 ConditionExpression expression) {
@@ -288,6 +291,9 @@ public class ConditionManager {
             this.trigger = trigger;
             this.externalCondition = externalCondition;
             this.expression = expression;
+            if (expression.getQuietCount() > 0) {
+                quietMap = new HashMap<>();
+            }
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


lucasponce:
As commented in the JIRA, I am skeptical about this option, as basically it introduces a sort of dampening effect.
Perhaps I don't have enough examples/use cases for it.
Could you ellaborate more use cases to discuss it ?

jshaughn:
@lucasponce, I responded in HWKALERTS-240.  Hopefully that post explains well why I think this is still a useful option for the alerter, even if we don't use the alerter as the solution for that issue.

Please review as just an additional alerter feature.

jshaughn:
retest this please

jshaughn:
@lucasponce Unless you have an issue, I think this can be merged.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
```diff
@@ -489,15 +495,17 @@ private void evaluateEach(String target, Map<String, Map<String, List<? extends
...
             Map<String, String> preparedCondition = null;
             try {
                 preparedCondition = evaluator.prepare(queryResults);
             } catch (Exception e) {
                 log.debugf("Could not prepare evaluation, Skipping due to [%s]", e.getMessage());
             }
 
-            if (evaluator.evaluate()) {
+            boolean evalResult = evaluator.evaluate();
+            if (!isQuiet("_ALL_", expression.getQuietCount(), evalResult)) {
```
lucasponce: Minor one: I wonder if instead to treat "_ALL_" as a metricName, if it is not better just to pass a boolean flag to indicate all/each.

lucasponce:
@jshaughn sorry for the delay, overall LGTM, I have a minor comment but no blocker.
This feature can get merged.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/ConditionManager.java
jshaughn: I see your point, but I guess for now I'd prefer to leave things unchanged.  By using a special metricName the underlying mechanism can just use it as a Map key and not have to know it's something special.

stefannegrea:
Thank you @jshaughn!

PR #782, opened 2017-03-21
[HWKMETRICS-612] Add DELETE APIs to all metric types. The API is appl…

stefannegrea:
…icable to individual metrics and it removes the metric information from all the indexes (metric, tag, and retention) and all the metric's data points.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
@@ -193,7 +193,8 @@ public class CounterHandler extends MetricsServiceHandler implements IMetricsHan
 
     @DELETE
     @Path("/{id}")
-    @ApiOperation(value = "Deletes the metric and associated data points, and updates internal indexes.")
+    @ApiOperation(value = "Deletes the metric and associated data points, and updates internal indexes."
+            + " Note: compressed data will not be deleted and still be accessible.")
     @ApiResponses(value = {
             @ApiResponse(code = 200, message = "Metric deletion was successful."),
             @ApiResponse(code = 500, message = "Unexpected error occurred trying to delete the metric.")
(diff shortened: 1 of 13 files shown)


jsanda:
It doesn't look like the PR does anything with compressed data points, nor should it IMO. The data_compressed table now uses TWCS, and the guidelines for using it are pretty specific. It is best for append-only data that uses a global TTL. While on that basis we should avoid explicitly deleting compressed data points, I do think it needs to be documented that deleting a metric does **not** delete the compressed data. It will instead expire according to data retention settings.

stefannegrea:
I think it will be really confusing to users when/if they query for data and get data points that happen to be compressed.

jsanda:
It may be confusing for users, hence the need for it to be well-documented. We need make it very clear that we do append-only writes for data. Mixing deletes with the append-only writes will make compaction less efficient and actually result in space not getting reclaimed. See [TWCW Part 1](http://thelastpickle.com/blog/2016/12/08/TWCS-part1.html) for an explanation on this.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/CounterHandler.java
```diff
@@ -193,7 +193,8 @@ public void getMetric(@Suspended final AsyncResponse asyncResponse, @PathParam("
 
     @DELETE
     @Path("/{id}")
-    @ApiOperation(value = "Deletes the metric and associated data points, and updates internal indexes.")
+    @ApiOperation(value = "Deletes the metric and associated data points, and updates internal indexes."
+            + " Note: compressed data will not be deleted and still be accessible.")
```
jsanda: I would make one more minor change, something along the lines of, "Note: compressed data will not be deleted. It is instead deleted as part of the normal data expiration as defined by the data retention settings; consequently, compressed data will still be accessible until it expires."

stefannegrea: That is a good suggestion. I will update the PR.

PR #784, opened 2017-03-26
HWKALERTS-240 Add Metrics Group Trigger support in the metrics alerter

jshaughn:
MGT support allows users to define a group trigger whose member triggers are
managed automatically by the alerter.

@lucasponce This is ready for review although it is marked DNM while I continue testing and I need to dial back some logging.

### DIFF
--- a/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
+++ b/alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
@@ -1,0 +1,593 @@
+/*
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.metrics.alerter.groups;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.annotation.Resource;
+import javax.ejb.Singleton;
+import javax.ejb.Startup;
+import javax.inject.Inject;
+
+import org.hawkular.alerts.api.json.GroupMemberInfo;
+import org.hawkular.alerts.api.model.condition.CompareCondition;
+import org.hawkular.alerts.api.model.condition.Condition;
+import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.services.DefinitionsEvent;
+import org.hawkular.alerts.api.services.DefinitionsListener;
+import org.hawkular.alerts.api.services.DefinitionsService;
+import org.hawkular.metrics.core.service.MetricsService;
+import org.hawkular.metrics.model.Metric;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.notifications.Listener;
+import org.infinispan.notifications.cachemanagerlistener.annotation.ViewChanged;
+import org.infinispan.notifications.cachemanagerlistener.event.ViewChangedEvent;
+import org.jboss.logging.Logger;
+
+/**
+ * Another feature of the Metrics Alerter is an ability to manage a Metrics Group Trigger.  A Metrics Group Trigger
+ * is an hAlerting group trigger that has a member trigger for each metric is a set scoped by a metrics tag query.
+ * The concept is similar to data-driven triggers but because metrics does not lend itself to the required
+ * source+dataId approach used by data-driven triggers, we use this approach.
+ * <p>
+ * We interact with hAlerting to pull properly tagged group triggers.  We set up a job for each to query for
+ * metrics, and maintain member triggers for each.
+ * </p><p>
+ * Each Metrics Group Trigger has several tags to guide its handling:
+ * </p><pre>
+ *   "HawkularMetrics" : "GroupTrigger"       // Required to be processed as Metrics Group Trigger
+ *   "DataIds"         : DataId1..DataIdN     // The DataIds used in the trigger conditions, required
+ *   DataId1           : tag query            // Metrics tag query for dataId1 metrics
+ *     ...
+ *   DataIdN           : tag query            // Metrics tag query for dataIdN metrics
+ *   "SourceBy"        : TagName1..TagNameN   // *see below*
+ * </pre>
+ * <h4>SourceBy</h4>
+ * The SourceBy tag specifies the metric tag names used to determine the member trigger population.  As an example,
+ * consider a group trigger with one CompareCondition:  HeapUsed > 80% HeapMax.  We want to compare HeapUsed to HeapMax
+ * but of course each comparison should be done on metrics from the same machine.  We don't want to compare all
+ * HeapUsed to all HeapMax for every machine in the data center.  We'd want a group trigger tagged something like this:
+ * <pre>
+ *   "HawkularMetrics" : "GroupTrigger"
+ *   "DataIds"         : "HeapUsed,HeapMax"
+ *   "HeapUsed"        : name = "HeapUsed"
+ *   "HeapMax"         : name = "HeapMax"
+ *   "SourceBy"        : Machine
+ * </pre>
+ * This would result in one member trigger per machine, for machine names common to HeapUsed and HeapMax metrics. So,
+ * if we had the following metrics in the database, tagged as specified:
+ * <pre>
+ *   /machine0/HeapUsed    {name=HeapUsed, Machine=machine0}
+ *   /machine0/HeapMax     {name=HeapMax, Machine=machine0}
+ *   /machine1/HeapUsed    {name=HeapUsed, Machine=machine1}
+ *   /machine1/HeapMax     {name=HeapMax, Machine=machine1}
+ *   /machine2/HeapUsed    {name=HeapUsed}
+ *   /machine2/HeapMax     {name=HeapMax}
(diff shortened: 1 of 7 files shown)


alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+ * </p><p>
+ * Each Metrics Group Trigger has several tags to guide its handling:
+ * </p><pre>
+ *   "HawkularMetrics" : "GroupTrigger"       // Required to be processed as Metrics Group Trigger
+ *   "DataIds"         : DataId1..DataIdN     // The DataIds used in the trigger conditions, required
+ *   DataId1           : tag query            // Metrics tag query for dataId1 metrics
+ *     ...
+ *   DataIdN           : tag query            // Metrics tag query for dataIdN metrics
+ *   "SourceBy"        : TagName1..TagNameN   // *see below*
+ * </pre>
```
lucasponce: Some feedback as a reader.
Jumping to the "SourceBy" examples without describing simple examples (i.e. conditions with only one dataId) created some initial confusion.
I think, some initial example before the SourceBy would help.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+
+        mgtExecutor = new ScheduledThreadPoolExecutor(THREAD_POOL_SIZE);
+
+        refresh();
+
+        if (null == definitionsListener) {
+            log.info("Registering Trigger UPDATE/REMOVE listener");
+            definitionsListener = new DefinitionsListener() {
+                @Override
+                public void onChange(Set<DefinitionsEvent> event) {
```
lucasponce: I think at some point the metrics alerter needs to move to DistributedEvent instead DefinitionsEvent.
Perhaps it is fine to maintain the singleton behaviour until the features get maturity, just to not introduce more variables in potential debugging issues.
But just leaving a note here.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+                            log.warnf("Tagged as Metrics Group Trigger is not a group trigger, skipping %s", t);
+                            return false;
+                        }
+                        return true;
+                    })
+                    .collect(Collectors.toSet());
+
+            // for each metrics group trigger run a job to manage its member trigger set
+            for (Trigger mgt : mgts) {
+                if (mgtFutures.containsKey(mgt)) {
```
lucasponce: Not sure if I am right, but what happen if you want to modify an existing trigger ?
How to detect that a mgt trigger has changed a tag expression ?
It is not skipped on this check ?

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+                    log.warnf("Metrics Group Trigger is not a group trigger, skipping %s", mgt);
+                    return;
+                }
+                Map<String, String> tags = mgt.getTags();
+                if (isEmpty(tags.getOrDefault(TAG_DATA_IDS, ""))) {
+                    log.warnf("Required Metrics Group Trigger tag [%s] missing or null, skipping %s",
+                            TAG_DATA_IDS, mgt);
+                    return;
+                }
+                if (isEmpty(tags.getOrDefault(TAG_SOURCE_BY, ""))) {
```
lucasponce: shouldn't be consider an empty SourceBy valid ? or similar as wildcard * ?

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: I agree that more documentation is needed.  It's not written yet but this feature needs an entry in the Alerting section of the Metrics' User Guide.   I'll work on that and have it prior to the metrics release.  Tutorial lessons for both the existing external alerter support and this new feaqture would also be worthwhile.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: +1 on both points.  I think we should move to DistributedEvent, but after this is released using the single-node, coordinator approach, just to reduce variables at the moment.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: >  but what happen if you want to modify an existing trigger

Right, good question.  There are two things going on.  The change listener is responsible for identifying the set of Metrics Group Triggers (MGTs) and ensuring that each has a recurring job running.  Each MGT job is responsible for maintaining the set of member triggers for that MGT.  So, if an MGT comes or goes the change listener will add/remove a job.  If the MGT is updated with different tag values, or if its metric set changes,  the member set will be updated to reflect those changes.   If this makes it more clear let me know and I'll add this blurb to the class doc.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: Good question.  SourceBy is required, and does not default to "*", because I want to make sure that the client explicitly sets "*" when desired.  The reason is to prevent an undesired generation of a large member trigger set.  The wildcard SourceBy basically means use all combinations of the metrics resulting from the dataId tag queries.   For single-dataId-conditions it is likely that a user would use SourceBy="*", but this is just a guard against a potentially bad mistake, particularly in the CompareCondition scenario.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
lucasponce: Ok, so it is the job who is periodically query the backend to synchronize the trigger and metric definitions.
For both sides: new metrics or changes in the tags.
If I correctly understand the Job's logic.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+                log.warnf("doomedMemberIds        : %s", existingMemberIds);
+                for (String doomedMemberId : existingMemberIds) {
+                    try {
+                        log.warnf("Removing trigger %s", doomedMemberId);
+                        definitionsService.removeTrigger(mgt.getTenantId(), doomedMemberId);
+                    } catch (Exception e) {
+                        log.warnf("Failed to delete member %s: %s", doomedMemberId, e.getMessage());
+                    }
+                }
+            } catch (
```
lucasponce: Minor, wrong formatting code.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+                        .map(m -> m.getMemberId())
+                        .collect(Collectors.toSet());
+                Set<String> existingMemberIds = existingMembers.stream()
+                        .map(t -> t.getId())
+                        .collect(Collectors.toSet());
+                log.warnf("members: %s", members);
+                log.warnf("memberIds        : %s", memberIds);
+                log.warnf("existingMemberIds: %s", existingMemberIds);
+                for (GroupMemberInfo member : members) {
+                    if (existingMemberIds.contains(member.getMemberId())) {
```
lucasponce: So, if follow the logic. The Job read the group trigger, it calculates the new members, then check with existing ones, to maintain the existing ones, create new ones and removed older one.
I have one question, is the memberId just enough to check this ?
I mean, imagine that the same trigger can be modified in all DataIds or SourceBy values, the memberId will protect this ?

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
```diff
@@ -0,0 +1,597 @@
...
+                }
+
+                Map<String, String> memberContext = new HashMap<>();
+                memberContext.put("source", source.toString());
+
+                for (Map<String, String> memberDataIdMap : memberDataIdMaps) {
+                    // use the memberDataIdMap hashcode as the member trigger id. We'll use this id to
+                    // determine whether the member already exists.
+                    GroupMemberInfo member = new GroupMemberInfo(mgt.getId(),
+                            String.valueOf(memberDataIdMap.hashCode()), null, null, memberContext,
```
lucasponce: Ok, now I see where the member id is generated.

lucasponce: Another comment at this point, I don't see bad to suply a name and description, as this info is used by actions plugins to show where an alert is generated. If not, an email plugin will show undesired null/empty fields.

lucasponce:
Ok @jshaughn I finished a primer review on it with several comments.
Probably more clarifications rather than changes.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: >Ok, so it is the job who is periodically query the backend to synchronize the trigger and metric definitions.

That's correct.  It's a repeating job, one per MGT, and on each run it synchronizes the member triggers for that group trigger, based on the current MGT tag values and the resulting metrics from the dataId tag queries.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: Interesting, Eclipse formatter did not catch that.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: This wass another good observation, but I think this is OK, as you saw below, the member ID reflects the whole memberDataIdMapping by being the hashcode of the whole Map.

alerting/alerter-war/src/main/java/org/hawkular/metrics/alerter/groups/GroupTriggerManager.java
jshaughn: It defaults to the MGT, so I don't think it is necessary here.

lucasponce:
LGTM, as soon as it passes the CI it can get merged.

jshaughn:
@stefannegrea please merge.

stefannegrea:
Thank you @jshaughn!

PR #787, opened 2017-03-29
HWKMETRICS-636 Move release/0.23.0 to Alerting fix version 1.5.4.Final

jshaughn:


jshaughn:
retest this please

jshaughn:
@stefannegrea I can't explain the perf failures. I don't see how it can be related.

FilipB:
@jshaughn @stefannegrea the performance drop is visible because the baselines for release/0.23.0 and master have diverged and the check only uses baselines for master. In other words master is faster than release/0.23.0

stefannegrea:
Will merge the PR as-is based on the comments from @FilipB.

PR #788, opened 2017-03-30
Adds a new itest for the Alerter MGT feature

jshaughn:


stefannegrea:
Thank you @jshaughn!

stefannegrea:
Merging without perf tests since only integration test code was modified.

PR #789, opened 2017-03-30
HWKMETRICS-638 Add Elasticsearch alerter to Metrics deployments

jshaughn:
@lucasponce This PR adds the elasticsearch stuff into the ear deployment.  It should be sanity checked using the most recent alerting master (or, actually https://github.com/hawkular/hawkular-alerts/pull/314 if it hs not yet been merged).

We'll hold the merge until we update this PR with the 1.6 release.

### DIFF
--- a/dist/component/component-ear/pom.xml
+++ b/dist/component/component-ear/pom.xml
@@ -55,6 +55,12 @@
     </dependency>
     <dependency>
       <groupId>org.hawkular.alerts</groupId>
+      <artifactId>hawkular-alerts-actions-elasticsearch-metrics</artifactId>
+      <version>${version.org.hawkular.alerts}</version>
+      <type>war</type>
+    </dependency>
+    <dependency>
+      <groupId>org.hawkular.alerts</groupId>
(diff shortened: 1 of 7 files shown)


lucasponce:
I have built locally the 1.6.0.Final-SNAPSHOT and this PR and a smoke tests (creating a Elasticsearch Trigger and check that alerter conects) passed.

I used the generated
```
dist/standalone/wildfly-standalone/target/wildfly/wildfly-10.0.0.Final
```
for the smoke tests.
Not tested the containers one but I guess to validate the connectivity inside the .ear both should be enough.

Let me know if there is any additional test to peform on this.

A side comment, I found that overlays files are inside the license checker plugin, I had not expected fun.
Why don't add them into the exclude them at root pom.xml ?
Something like this:
```
            <!-- Exclude overlays -->
            <exclude>overlays/**</exclude>
```

jshaughn:
retest this please

stefannegrea:
Thank you @jshaughn!

PR #790, opened 2017-03-31
[HWKMETRICS-637] Wait for entire C* cluster to be up for schema updates

jsanda:
In OpenShift, Hawkular may start before Cassandra. We already have checks in
place in start up code to wait until we connect to Cassandra. Things get more
complicated though with multiple C* nodes. If we have nodes A and B, only A is
up, we connection, and start trying to apply schema updates. Some updates may
fail if they target node B. This also assume replicaiton_factor = 1 since we do
not support anything else right now. The schema update will fail, start up
fails, and then the hawkular pod goes into a restart loop.

This commit changes start up logic so that we wait until all nodes are up
before we attempt to apply schema updates.


stefannegrea:
Thank you @jsanda!

PR #794, opened 2017-04-05
update getting started with release build section

jsanda:


FilipB:
retest this please

README.adoc
```diff
@@ -43,18 +43,81 @@ The http://www.hawkular.org/docs/rest/rest-metrics.html[REST API] should be easy
...
+* WildFly distribution
+* EAR distribution
 
-Check out the https://github.com/hawkular/hawkular-metrics/releases[Metrics releases] page and download the latest
-version of the web application archive: `hawkular-metrics-api-jaxrs-X.Y.Z.war`
+IMPORTANT: Earlier versions of Hawkular Metrics could be run by deploying `hawkular-metrics-api-jaxrs-X.Y.Z.war`. This
+is longer supported.
 
-Copy this file to the `standalone/deployments` directory of the Wildfly server.
+The WildFly distribution is a pre-configured WildFly 10 server that includes Hawkular Alerts in addition to Hawkular
```
stefannegrea: I would not mention what version of WF is included because it is packaged. That will make it easy when we upgrade WF versions, we do not need to update the documentation.

jsanda: I think we do want to mention it to be very clear what the supported runtime is, just like we mention the version of Cassnandra. If anything, I think we ought to make it more prominent.

stefannegrea: Sounds good; then let's see what feedback we get ...

stefannegrea:
Thank you @jsanda!

PR #795, opened 2017-04-05
[HWKMETRICS-643] Disable the job for deleting expired metrics if the …

stefannegrea:
…run interval configuration is zero or negative.


jsanda:
Looks like an alerts test is failing:

```
Failed tests: 
  GroupTriggerAlerterITest.t02_mgtAutoResolveTest:278 expected:<2> but was:<1>
```

@jshaughn would you mind taking a look?

jsanda:
The ```metricExpirationJobFrequencyInDays > 0``` check to decide whether or not to schedule the job makes sense, but I think an ```enabled``` property is more clear and more consistent with what we already do. For the compression job, we have ```hawkular.metrics.jobs.compression.enabled``` and ```COMPRESSION_JOB_ENABLED```. How about something similar for  the deletion job?

FilipB:
retest this please

PR #801, opened 2017-04-10
[HWKMETRICS-606] change compaction strategy of metrics_idx to STCS

jsanda:
Note that this is for the release/0.23.0 branch.


stefannegrea:
Thank you @jsanda!

PR #808, opened 2017-04-11
HWKMETRICS-649 GroupTriggerAlerterITest fails in slow environments

lucasponce:
cc @stefannegrea


stefannegrea:
Merging this without perf tests passing because it is just a change in the integration tests code.

stefannegrea:
Thank you @lucasponce!

PR #809, opened 2017-04-11
[HWKMETRICS-650] Upgrade rxjava 1.2.6 -> 1.2.7

pgier:
Matches version with vert.x platform


burmanm:
1.2.9 is already out, I don't think it makes sense to update to older version.

stefannegrea:
This was done to align project dependencies with other projects.  @burmanm, do you mind opening a JIRA and PR to update master to the latest version?

stefannegrea:
Thank you @pgier!

burmanm:
I'm not sure it makes any sense to align with projects that we're not dependant on. What's the point of that process?

burmanm:
Especially since this was targeting master.

PR #810, opened 2017-04-11
[HWKMETRICS-650] Upgrade rxjava 1.2.6 -> 1.2.7

pgier:
Matches version with vert.x platform


burmanm:
1.2.9 is already out, I don't think it makes sense to update to older version.

stefannegrea:
Thank you @pgier!

PR #818, opened 2017-04-18
[HWKMETRICS-614] Temporary tables for writes

burmanm:
This is WIP. At the time of writing this, gauge inserts / fetches work correctly, but for example metrics_idx queries rely on the partition keys and that needs to be fetched from the temp tables also (otherwise our unit tests fail because they test two things and not just insert / delete like they claim to)


core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.27.0.groovy
```diff
@@ -0,0 +1,263 @@
...
+    availability blob,
+    l_value bigint,
+    tags map<text,text>,
+    PRIMARY KEY ((tenant_id, type, metric, dpart), time)
+) WITH CLUSTERING ORDER BY (time DESC)
+"""
+}
+
+// Lets hope this will never be compacted.. just to be on the safe-side though..
+schemaChange {
```
burmanm: I left this unimplemented for tables - what would be the best idea? We ideally want something that's quick to flush / minor compact. I selected TimeWindow with 1 day size as the table should live for only ~3 hours.

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.27.0.groovy
```diff
@@ -0,0 +1,263 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+schemaChange {
+  version '8.0'
+  author 'burmanm'
+  tags '0.27.x'
+  cql """
+CREATE TABLE data_temp_0 (
```
jsanda: I would make the tags column ```frozen```.  It will result in the map being serialized into a single blob column vs being stored in multiple columns. It is more efficient both in terms of writes and in terms of space. The current data table does **not** use ```frozen``` which is an oversight on my part. The caveat with a ```frozen``` collection is that you cannot add/append/remove a single element. You have to update the entire collection. For data point tags I think this is perfectly fine.

core/schema/src/main/resources/org/hawkular/schema/updates/schema-0.27.0.groovy
jsanda: I would probably stick with STCS especially considering that within a given time bucket TWCS  uses STCS.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -1053,16 +1057,25 @@ private BoundStatement bindDataPoint(PreparedStatement statement, Metric<?> metr
 
         // Max time back is <24 hours.
         if(startZone.isBefore(endZone.minus(23, ChronoUnit.HOURS))) {
+            log.infof("Adjusting the startZone\n");
```
jsanda: I would be inclined to change some of these info level log statements to debug or trace.

jsanda:
@burmanm when we last spoke, you mentioned that you are going to go with dropping instead of truncating tables since truncation requires all C* nodes to be up (why I do not know). In that case I propose using the timestamp for the corresponding compression block as the table name suffix to ensure we avoid any naming collision issues. 

I am also not so sure Cassalog makes sense here. We will need to create tables at start up (as necessary), and then again when the compression job finishes. We will also need to generate prepared statements for the tables as we create them.

Lastly, the dpart column should just be the start time of the corresponding compression block, so maybe rename it to something better.

jsanda:
We have collectively said that MetricsServiceImpl and DataAccess is overdue for some major refactoring. When I started on a prototype for this, I started moving all the logic for dealing the raw tables into its own class. I called something like RawDataManager or RawDataService. Stuff like this is often rather subjective, but I think this new functionality is a good candidate for being separated out into its own class (or classes).

burmanm:
If we don't use cassalog then we will need to ensure that multiple hawkular-metrics instances don't have a race creating the temp tables (like we saw with the cassalog). I'd like to avoid creating that multiple times.

Problems with these drop & recreate is that if nodes are down we might get some silly situations as it seems there's multiple bugs open in Cassandra JIRA for these situations. So one option could be to stop the CompressionJob if there are nodes down (but that brings some other problems also - lets say the node has been down for over 24 hours..).

jsanda:
Cassalog does not offer any functionality in terms of preventing races conditions. I have seen issues reported involving drop/create as well. It is going to require some investigation and testing. As for the race condition, I think we have a way to handle that. We have a lock service that internal to the job scheduler right now. I think it could be used for this situation.

burmanm:
My roadmap of missing features:
* Fix DeleteTenant job to use DataAccess (and with that, temp tables also)
* Add support for reading and writing out-of-order writes to the old data table for GAUGE/COUNTER/AVAILABILITY
  * Add configuration option for the data-table to disable the feature
  * Create modifications to certain insert statements that data table still uses timeuuids
    * This is to ensure we don't need any data migration tools

Testing fixes and additions:
* MixedMetricsITest delete tests
* CompressJob tests
* Add two new tests:
  * Write n-datapoints to n-metrics in random order and ensure that all of them are correctly read back using TokenRange operations
  * Push invalid dates (such as 1970s) to all the methods that use temporary tables and ensure that there's no Exceptions (such as NPE) coming
    * This is with the "out-of-order support" disabled

Any other future work to another PR in my opinion, such as:
  * Split DataAccess to multiple parts
  * Experimental support for using "WAL log in Cassandra" and in-memory Gorilla streams to write data to (which are used for compression then) - single node featureset for first PR.

burmanm:
Well, something odd in the scheduling of the job (parts I have not changed). The job isn't executed correctly at all (the logging is quite different from what I see on my machines and also there's no doOnComplete logging in Travis logs, I wonder how CountDownLatch was freed..)

burmanm:
retest this please

gbaufake:
@burmanm 
I tested this and I got (complete log > https://paste.fedoraproject.org/paste/69I38o6F2sQVv6NJu8pOVA)

It seems to be an error on CompressDataJobITest.testCompressJob and GCGraceSecondsManagerITest

caused by
can't execute it on hawkulartest (INSERT INTO tenants (id, retentions) VALUES (?, ?) IF NOT EXISTS)

and  

can't execute it on hawkulartest (SELECT table_name, gc_grace_seconds FROM system_schema.tables WHERE keyspace_name = ?)

burmanm:
Well, the GCGraceManager wasn't modified in this PR. I think the problem is that connections are shared and so some of the commands go to connections that are using different keyspace. 

This isn't a real world problem as we only use "hawkular_metrics" keyspace, but something in the testing.  @jsanda knows more about the GCGraceManager and how it tries to test things. Might be that GCGraceManager does not like newer Cassandra drivers.

Caused by: java.lang.IllegalStateException: Statement was prepared on keyspace gc_grace_test, can't execute it on hawkulartest (INSERT INTO tenants (id, retentions) VALUES (?, ?) IF NOT EXISTS)

burmanm:
That same reason is the failure for both tests actually.

jsanda:
I don't want to do it for this PR, but making the DataStax driver's SchemaChangeListener reactive would be a nice addition to the rx-java-driver module.

burmanm:
You mean this way?

https://gist.github.com/burmanm/01183fdbcd069759c6d07dc7dbca2f2d

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -72,13 +96,157 @@
...
+
+    private static String scanTableBase =
+            "SELECT tenant_id, type, metric, time, n_value, availability, l_value, tags, token(tenant_id, type, " +
+                    "metric, dpart) FROM %s " +
+                    "WHERE token(tenant_id, type, metric, dpart) > ? AND token(tenant_id, type, metric, dpart) <=" +
+                    " ?";
+
+    // Create statement prototype
+
+    private static String TEMP_TABLE_BASE_CREATE = "CREATE TABLE %s ( " +
```
burmanm: It made no difference to store bigint or smallint, I tested this at some point in the past. DPART would never be used, it was just a copy from the original structure - and to keep functions the same. 

Different table schema than data_0 (the overflow) requires extra structure in the code and makes it far more complicated to maintain, so I would rather keep this quite close to the one that's being kept on the disk as well.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -223,17 +341,185 @@
...
+            metadata.getKeyspace(session.getLoggedKeyspace()).getTables().stream()
+                    .map(AbstractTableMetadata::getName)
+                    .filter(t -> t.startsWith(TEMP_TABLE_NAME_PROTOTYPE))
+                    .forEach(tables::remove);
+
+            return tables;
+        })
+
+                .flatMapIterable(s -> s)
+                .zipWith(Observable.interval(300, TimeUnit.MILLISECONDS), (st, l) -> st)
```
burmanm: No, the delay is there to make sure we don't bomb too many schema changes in a short period of time.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -755,53 +967,110 @@ protected void initPreparedStatements() {
...
     }
 
+    Long tableToMapKey(String tableName) {
+        LocalDateTime parsed = LocalDateTime
+                .parse(tableName.substring(TEMP_TABLE_NAME_PROTOTYPE.length()),
+                        TEMP_TABLE_DATEFORMATTER);
+        return Long.valueOf(parsed.toInstant(ZoneOffset.UTC).toEpochMilli());
+    }
+
+    String getTempTableName(long timestamp) {
```
burmanm: Yeah, I think I mentioned that in some earlier comment, but decided that I would rather move all that work to a separate PR that would refactor the DataAccessImpl / MetricsServiceImpl

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
```diff
@@ -1046,39 +1310,14 @@ private BoundStatement bindDataPoint(PreparedStatement statement, Metric<?> metr
...
-    @Override
-    public Observable<Row> findAvailabilityData(MetricId<AvailabilityType> id, long timestamp) {
-        return rxSession.executeAndFetch(findAvailabilityByDateRangeInclusive.bind(id.getTenantId(),
-                AVAILABILITY.getCode(), id.getName(), DPART, UUIDs.startOf(timestamp), UUIDs.endOf(timestamp)));
-    }
 
-    @Override
-    public <T> Observable<ResultSet> deleteMetricData(MetricId<T> id) {
-        return rxSession.execute(deleteMetricData.bind(id.getTenantId(), id.getType().getCode(), id.getName(), DPART));
+        return getPrepForAllTempTables(TempStatement.DELETE_DATA)
```
burmanm: If we don't delete the data, then the compress job will reinsert them to the data_compressed and then the jobs "DeleteTenant" or any other delete don't really do much anymore. As there's no place they would delete from.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
jsanda: > It made no difference to store bigint or smallint

Are you talking about on size on disk? It should make a difference for heap memory usage when reading rows.

jsanda: > Different table schema than data_0 (the overflow) requires extra structure in the code and makes it far more complicated to maintain

I have not gotten to the overflow stuff yet, but why do we need the DPART column for the overflow table?

gbaufake:
@burmanm 
 
The tests are still failing:

`Caused by: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Statement was prepared on keyspace gc_grace_test, can't execute it on hawkulartest (SELECT table_name, gc_grace_seconds FROM system_schema.tables WHERE keyspace_name = ?)
Caused by: java.lang.IllegalStateException: Statement was prepared on keyspace gc_grace_test, can't execute it on hawkulartest (SELECT table_name, gc_grace_seconds FROM system_schema.tables WHERE keyspace_name = ?)
Results :`

`Failed tests:
  CompressDataJobITest.testCompressJob:196->BaseITest.doAction:126 Unexpected onError events (0 completions) (+1 error)
  GCGraceSecondsManagerITest.multiNodeClusterWithRF2:130 Updates failed`

`Tests run: 64, Failures: 2, Errors: 0, Skipped: 5`

Does c3a42d7 correct it somehow?

burmanm:
No, c3a42d7 does not touch session sharing in the tests.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -703,7 +720,67 @@ public void setDefaultTTL(int defaultTTL) {
...
+            startTime = startTime.plus(2, ChronoUnit.HOURS);
+        }
+
+        return Completable.fromObservable(dataAccess.createTempTablesIfNotExists(timestamps));
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Completable compressBlock(long startTimeSlice, int pageSize, int maxConcurrency) {
+        return Completable.fromObservable(
```
jsanda: There is a lot going on in this method both in terms of the work it is doing as well as the RxJava operators being used. It would be really helpful to do a BlueJeans session explain what is happen and the operators being used.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -703,7 +720,67 @@ public void setDefaultTTL(int defaultTTL) {
...
+                                        return dataAccess
+                                                .insertCompressedData(metricId, startTimeSlice, cpc, getTTL(metricId))
+                                                .mergeWith(updateMetricExpiration(metricId).map(rs -> null));
+                                    });
+                                }), maxConcurrency)
+                        .flatMap(rs -> rs)
+                        .doOnError(Throwable::printStackTrace)
+                        .doOnCompleted(() -> {
+                            log.infof("Compress part completed");
+                            dataAccess.dropTempTable(startTimeSlice)
```
jsanda: I noticed that ```dropTempTable``` has no retry logic. We need to make sure these tables get deleted. What happens if the query fails?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
jsanda: 300 ms is still completely arbitrary. The slowest and safest way is probably to use ```Metadata.checkSchemaAgreement```.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
burmanm: I'll add comments to the code to explain what is happening

FilipB:
Soak test failed after ~3 hours with following exceptions in server.log:

2017-06-27 08:15:50,736 INFO  [org.hawkular.metrics.core.jobs.TempTableCreator] (RxComputationScheduler-3) Temporary tables are valid until 2017-06-28T12:00Z
2017-06-27 09:00:48,686 INFO  [org.hawkular.metrics.core.jobs.TempDataCompressor] (RxIoScheduler-6) Starting to process temp table for starting time of 2017-06-27T10:00:00.000Z
2017-06-27 09:02:15,533 ERROR [org.hawkular.metrics.api.jaxrs.util.ApiUtils] (RxComputationScheduler-2) HAWKMETRICS200010: Failed to process request: java.nio.BufferUnderflowException
        at java.nio.Buffer.nextGetIndex(Buffer.java:500)
        at java.nio.HeapByteBuffer.get(HeapByteBuffer.java:135)
        at fi.iki.yak.ts.compression.gorilla.ByteBufferBitInput.flipByte(ByteBufferBitInput.java:87)
        at fi.iki.yak.ts.compression.gorilla.ByteBufferBitInput.getLong(ByteBufferBitInput.java:63)
        at fi.iki.yak.ts.compression.gorilla.Decompressor.nextValue(Decompressor.java:125)
        at fi.iki.yak.ts.compression.gorilla.Decompressor.next(Decompressor.java:56)
        at fi.iki.yak.ts.compression.gorilla.Decompressor.readPair(Decompressor.java:37)
        at org.hawkular.metrics.core.service.transformers.DataPointDecompressTransformer.lambda$call$1(DataPointDecompressTransformer.java:87)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:69)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.emitScalar(OperatorMerge.java:395)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.tryEmit(OperatorMerge.java:355)
        at rx.internal.operators.OperatorMerge$InnerSubscriber.onNext(OperatorMerge.java:846)
        at org.hawkular.rx.cassandra.driver.ResultSetToRowsTransformer$RowProducer.produce(ResultSetToRowsTransformer.java:111)
        at org.hawkular.rx.cassandra.driver.ResultSetToRowsTransformer$RowProducer.lambda$execute$0(ResultSetToRowsTransformer.java:154)
        at rx.internal.schedulers.EventLoopsScheduler$EventLoopWorker$1.call(EventLoopsScheduler.java:172)
        at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: rx.exceptions.OnErrorThrowable$OnNextValue: OnError while emitting onNext value: com.datastax.driver.core.ArrayBackedRow.class
        at rx.exceptions.OnErrorThrowable.addValueAsLastCause(OnErrorThrowable.java:118)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:73)
        ... 14 more

burmanm:
@FilipB Well, you managed to hit an interesting edge case in HWKMETRICS (it's not a bug introduced in this PR) where a certain bit-length is exact. I'll fix it in this PR and send a separate fix to the master also.

burmanm:
retest this please

gbaufake:
The integrations tests still failing such as the stack trace that I made on the last comment.

FilipB:
BufferUnderflowException from previous comment is still visible

core/metrics-core-service/src/test/java/org/hawkular/metrics/core/service/metrics/GaugeITest.java
```diff
@@ -346,68 +346,128 @@ public void addAndFetchGaugeDataAggregates() throws Exception {
...
-                new DataPoint<>(start.plusMinutes(2).getMillis(), 2.2),
-                new DataPoint<>(start.plusMinutes(4).getMillis(), 3.3),
-                new DataPoint<>(end.getMillis(), 4.4)));
-
-        Observable<Void> insertObservable = metricsService.addDataPoints(GAUGE, Observable.just(m1));
-        insertObservable.toBlocking().lastOrDefault(null);
-
-        DateTime startSlice = DateTimeService.getTimeSlice(start, CompressData.DEFAULT_BLOCK_SIZE);
-        DateTime endSlice = startSlice.plus(CompressData.DEFAULT_BLOCK_SIZE);
+//    @Test
```
jsanda: Looks like this test needs to be updated. When I uncommented and ran, it fails with:

```
java.lang.AssertionError: Not completed! (0 completions) (+1 error)
	at org.hawkular.metrics.core.service.metrics.GaugeITest.addAndCompressData(GaugeITest.java:397)
Caused by: java.util.concurrent.ExecutionException: com.datastax.driver.core.exceptions.InvalidQueryException: unconfigured table data_temp_2017063020
Caused by: com.datastax.driver.core.exceptions.InvalidQueryException: unconfigured table data_temp_2017063020
```

jsanda:
I think there is a window where we can lose data points. Let's say the current temp table is is data_14:00. Any data points we ingest between 14:00 and 16:00 having a timestamp in that range will get written to data_14:00. More precisely, the data points will get written to data_14:00 as long as the table exists and as long its prepared statements are found in ```DataAccessImpl.preMap```. Let's say that the compressionn job kicks off. If we receive data points for the 14:00 block after ```DataAccessImpl.findAllDataFromBucket``` returns, I think we can wind up losing those data points. The data points will get written to data_14:00, they will be excluded from compression, and then the table will be dropped. To avoid data loss, I think we need to stop writing to data_14:00 when the compression job starts. That presents a problem of coordination since the job only runs on a single hawkular-metrics node.

burmanm:
Yes, that does happen (during the compression job). Although the window is not that large actually, since findAllData uses TokenRanges, so even if the compression job has started but hasn't processed that partition yet then nothing will be lost.

This could be solved however with some sort of broadcast / job that runs on every node. It would remove the temporary statements from the skip list before the compression job starts. That way all the writes during compression job would go to the out-of-order table already. This expire timer could run like ~5-10 minutes before the compression job also.

Should we do such now or later is another discussion. We don't have all the nodes connected at the moment, for single node this is simpler to solve.

burmanm:
Unless we plan to cluster Hawkular-Metrics instances. Alert's Infinispan connection could be one thing that kicks the job also.

jsanda:
My example was a bit screwed up. I was incorrectly thinking that the compression job would start at 16:00, but the job that processes the 14:00 - 16:00 block will start at 17:00. While there is a window where we could lose data points, we still support an hour of out of order writes. I think we are good. If we need to support more than an hour at any point, we can deal with it later.

jsanda:
When does data get purged from the data_0 table?

burmanm:
TODO
* Add TTL and TWCS to data_0 table

jsanda:
retest this please

gbaufake:
@jsanda and @burmanm 

Although soak test on this PR still ongoing with no critical failures so far I got this error at the beginning of the test 

`2017-07-06 09:00:21,715 ERROR [org.hawkular.metrics.core.service.MetricsServiceImpl] (RxComputationScheduler-2) Failure while trying to apply compression, skipping block: java.util.concurrent.ExecutionException: com.datastax.driver.core.exceptions.InvalidQueryException: unconfigured table data_temp_2017070610
        at com.google.common.util.concurrent.AbstractFuture$Sync.getValue(AbstractFuture.java:299)
        at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:286)
        at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:116)
        at rx.observable.ListenableFutureObservable$2$1.run(ListenableFutureObservable.java:78)
        at rx.observable.ListenableFutureObservable$1$1.call(ListenableFutureObservable.java:50)
        at rx.internal.schedulers.EventLoopsScheduler$EventLoopWorker$1.call(EventLoopsScheduler.java:172)
        at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: com.datastax.driver.core.exceptions.InvalidQueryException: unconfigured table data_temp_2017070610`

burmanm:
@gbaufake How old build are you using? In other PR you used a build older than June 19th. The latest commit in this PR for example has a test that tries to replicate that error (and fixes it).

gbaufake:
@burmanm it was a previous build as you pointed (my bad)

I restarted the soak test like @FilipB said the `2017-07-06 12:36:59,375 ERROR [org.hawkular.metrics.api.jaxrs.util.ApiUtils] (RxComputationScheduler-1) HAWKMETRICS200010: Failed to process request: java.nio.BufferUnderflowException` is present on 05712b0

and I also found 

`2017-07-06 12:44:58,729 ERROR [org.hawkular.metrics.api.jaxrs.util.ApiUtils] (RxComputationScheduler-4) HAWKMETRICS200010: Failed to process request: java.lang.IllegalArgumentException: -1 is not a recognized availability type
        at org.hawkular.metrics.model.AvailabilityType.fromByte(AvailabilityType.java:90)
        at org.hawkular.metrics.core.service.transformers.DataPointDecompressTransformer.lambda$call$1(DataPointDecompressTransformer.java:96)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:69)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.emitScalar(OperatorMerge.java:395)
        at rx.internal.operators.OperatorMerge$MergeSubscriber.tryEmit(OperatorMerge.java:355)
        at rx.internal.operators.OperatorMerge$InnerSubscriber.onNext(OperatorMerge.java:846)
        at org.hawkular.rx.cassandra.driver.ResultSetToRowsTransformer$RowProducer.produce(ResultSetToRowsTransformer.java:111)
        at org.hawkular.rx.cassandra.driver.ResultSetToRowsTransformer$RowProducer.lambda$execute$0(ResultSetToRowsTransformer.java:154)
        at rx.internal.schedulers.EventLoopsScheduler$EventLoopWorker$1.call(EventLoopsScheduler.java:172)
        at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: rx.exceptions.OnErrorThrowable$OnNextValue: OnError while emitting onNext value: com.datastax.driver.core.ArrayBackedRow.class
        at rx.exceptions.OnErrorThrowable.addValueAsLastCause(OnErrorThrowable.java:118)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:73)
        ... 14 more`

a partial server log is here: https://paste.fedoraproject.org/paste/0JZWOQrwcnWS0mPPheqxAg

gbaufake:
~ 4 h later soak test started Cassandra get write timeouts

Short Log: `2017-07-06 16:19:15,378 ERROR [org.hawkular.metrics.api.jaxrs.util.ApiUtils] (RxComputationScheduler-4) HAWKMETRICS200010: Failed to process request: java.lang.RuntimeException: java.util.concurrent.ExecutionException: com.datastax.driver.core.exceptions.WriteTimeoutException: Cassandra timeout during write query at consistency LOCAL_ONE (1 replica were required but only 0 acknowledged the write)`

jsanda:
What kind of write load is there on Cassandra?

FilipB:
Created JIRAs for those two exceptions: https://issues.jboss.org/browse/HWKMETRICS-700 and https://issues.jboss.org/browse/HWKMETRICS-701

PR #819, opened 2017-04-23
[HWKMETRICS-546] initial commit for automated management of C* cluster

jsanda:


cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
```diff
@@ -0,0 +1,130 @@
...
+                    try {
+                        String hostname = event.getDetails().get("hostname");
+                        InetAddress address = InetAddress.getByName(hostname);
+                        state.setLastNodeAdded(address);
+                    } catch (UnknownHostException e) {
+                        throw new RuntimeException(e);
+                    }
+                })
+                .flatMap(state -> {
+                    if (state.getLastNodeAdded() == null) {
```
burmanm: Shouldn't this be:

```java
.filter(state.getLastNodeAdded() == null)
.flatMap(state -> getNodes().filter(address -> !address.equals(state.getLastNodeAdded())))
```

? Instead of using if inside the flatMap and returning empty result. The processing stops there in any case.

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
```diff
@@ -0,0 +1,130 @@
...
+                .flatMap(context -> new CleanupCommand().run(context))
+                .toCompletable()
+                .concatWith(Completable.defer(() -> configurationService.save(CONFIG_ID, "lastRun",
+                        Long.toString(jobDetails.getTrigger().getTriggerTime())).toCompletable()))
+                .doOnCompleted(() -> {
+                    stopwatch.stop();
+                    logger.infof("Finished Cassandra cluster maintenance in %d ms",
+                            stopwatch.elapsed(TimeUnit.MILLISECONDS));
+                })
+                .doOnError(t -> logger.warn("Cassandra cluster maintenance has been aborted to to an unexpected error",
```
burmanm: No retry based on the exception?

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private Observable<InetAddress> getNodes() {
+        return Observable.from(session.getCluster().getMetadata().getAllHosts()).map(Host::getAddress);
+    }
+
+    private Observable<String> getTables(String keyspace) {
+        return Observable.from(session.getCluster().getMetadata().getKeyspace(keyspace).getTables())
+                .map(TableMetadata::getName)
+                .cache();
```
burmanm: Is this cache() in the correct place? The caller does not know some result is cached and this isn't actually cached for multiple callers.

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
```diff
@@ -0,0 +1,130 @@
...
+                .flatMap(state -> {
+                    if (state.getLastNodeAdded() == null) {
+                        return Observable.empty();
+                    }
+                    return getNodes().filter(address -> !address.equals(state.getLastNodeAdded()));
+                })
+                .concatMap(address -> createExecutionContexts(address, getTables(keyspace), jmxPort, keyspace))
+                .flatMap(context -> new CleanupCommand().run(context))
+                .toCompletable()
+                .concatWith(Completable.defer(() -> configurationService.save(CONFIG_ID, "lastRun",
```
burmanm: Maybe update ConfigurationService to always return ``Completable`` instead of ``Observable<Void>`` and replicating this logic?

jsanda: That actually won't help here. Our queries are basically hot observables since they are executed immediately without any subscription. The reason I use ```defer``` here is to ensure that ```configurationService.save``` is only called when the other Completable completes successfully.

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
jsanda: No, it is not in the right place.

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
jsanda: This initial commit is basically just a PoC. There will definitely be retry/failure logic.

cassandra-management/src/main/java/org/hawkular/cassandra/management/CassandraMaintenanceJob.java
jsanda: Yeah, I think it can be refactored, but the first filter needs to be,

```state.getLastNodeAddded() != null```

That would mean a node has been added to the cluster and maintenance needs to be performed.

PR #821, opened 2017-04-25

jsanda:


jsanda:
@burmanm can you review this please?

mwringe:
@jsanda do we need for QE to spin up additional tests to see how performance could be affected?

jsanda:
Yes, additional tests would be good. Note that this PR is for master. I assume those tests will want to work with one of the release branches. I will have PRs submitted to the necessary release branches tomorrow.

PR #822, opened 2017-05-01
try increasing max requests per connection for better throughput

jsanda:
I want to see how this does with the perf tests.


burmanm:
I think in the RxJava2 benchmarks I increased all the values to get better performance (or was it better stability.. in any case). The JMH was the best indicator what happens when you modify those values and it was quick to locally / with remote Cassandra.

PR #823, opened 2017-05-05
HWKMETRICS-482 Query by tag the restful way

jotak:
Adds endpoints GET /*/tags/{tags}/raw
where * is gauges|counters|strings|availability

Performs the same as POST /*/raw/query, restricted to querying by tag


stefannegrea:
Thank you @jotak!

PR #828, opened 2017-05-25
[HWKMETRICS-676] When a simple text is supplied for tag value in a qu…

stefannegrea:
…ery, use a simple query to get data from C*.


jsanda:
I was talking with @burmanm about this and he raised a good question. How can we be certain that the value is **not** a regex?

stefannegrea:
@jsanda, I limited the scope a little for this initial PR. There are two types of expressions supported by the parser, simple and complex text. The parsing expression for simple text is this: `SIMPLETEXT  : [a-zA-Z_0-9.]+ ;` . This PR only treats the simple text expression as a non-regex expression; and this is inline with actual language expectations.

stefannegrea:
retest this please

pilhuhn:
@stefannegrea Couldn't `.` be part of a regex?

stefannegrea:
@pilhuhn, you are right about the dot being part of regex, however the intent was to allow regex only in the complext text (delimited by ").

pilhuhn:
@stefannegrea Yes I understand. Which is why I asked about the dot - I would not expect it to be part of the simple expression (unlike e.g. underscore or even a dash)

stefannegrea:
@pilhuhn, thanks for the clarification. We need the dot in the simple expression because we allow and have clients that push tag names with the dot character (eg. label.pod).

burmanm:
Don't we have clients that push ``/`` also? And who knows what other classes as well. As we never limited the available characters, we have no safe way of telling which characters are used in the tag names.

Would it make more sense to have a parameter (or a function) that tells if the regexp is used or not?

For example: ``a1 = regexp(a.b)`` or ``a1 = exact(a.b)``. Problem is, we've never really documented the language behavior so I'm not sure what the "contract" we have currently is. We could go either way though, I might prefer the ``regexp(searchKey)`` if we consider breaking the current behavior.

FilipB:
Performance results:
Average response times for read tests are 10-30 times better depending on type of query.
E.g. following "gauges/stats?tags=userID${userID} AND group_id=groupID${groupID}&bucketDuration=600000ms" is 30 times faster now.

jshaughn:
The suggestion from @burmanm makes sense to me.  This interim solution is a weak band-aid that will likely hit issues given the limited character set being used for simple text.  As soon as possible the language should be extended with some sort of explicit regex decl when regex matching is desired.  The syntax should likely be agreed on and shared with alerting

stefannegrea:
Should we extend the proposal from @burmanm to `SIMPLETEXT` expressions too? Or that proposal is only for 'COMPLEXTEXT` expressions? Right now, the semantics of `SIMPLETEXT` expression for tag names are very clear, it is a strict equality and never regex. 

This PR will stand as is if the new regex notation (`regex("asb.sdf*")`) is only done for "COMPLEXTEXT".

burmanm:
I don't think we should have a complexText / simpleText detection at all since the simpletext does not match all the exact values.

We should probably just change the semantics to be:

``a = ^b$`` -> exact match always, no regexp (it wants a tag value ``^b$``, not ``b``)
``= regexp(^b$)`` -> regexp evaluate, in this case the match is ``b``

jshaughn:
Right, unless flagged as a regex match it would always be exact match, there would be no categorization of the match string based on the characters in the string.  As an aside, alerting is doing the same thing with the limited character set, although it is already performing exact match (not regex) when the value is appropriate.  Both metrics and alerting should support an explicit regex directive.  There are some back compat issues here, although I think the change in API is required.  If required we could support some sort of sysprop that forces regex always, just to potentially cover the back compat situation.

As for the syntax, instead of function-like expression that adds more parentheses, perhaps we should consider a new operator.  Keep '=' for exact match and introduce either "MATCHES" or maybe one of "~" or "~=" for regex.

a = ^b$               -> exact match
a MATCHES ^b$  -> regex match
a ~ ^b$             -> alternative regex (I prefer the keyword)

@stefannegrea @burmanm @lucasponce , thoughts?

lucasponce:
We would need another operator.
Today we use 
```
EQUAL: '=';
NOTEQUAL: '!=';
```
So, we would need the same both operators but for regex
```
EQUAL: '=';
NOTEQUAL: '!=';
EQUALREGEX: '~';
NOTEQUAL: '!~';
```
I prefer the shorter way, less verbosity the better.
At the moment that you chained several expressions, the string could be unmanegable.

jshaughn:
@lucasponce +1, looking at it again, I agree. I'd just change the enum:

REGEXMATCH: '\~';
NOTREGEXMATCH: '!\~';

Actually, regex could do the negation itself, but I like having the second operator.

stefannegrea:
@jshaughn, @lucasponce, @burmanm , I will rework this PR with new operators.

jshaughn:
Great. LGTM.  In alerting we will also make this change for the next version.

jsanda:
@stefannegrea can you add a test for the ```!~``` operator?

jotak:
For the record I'll have to add ~ and !~ in grafana as well

PR #829, opened 2017-05-25

jsanda:

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
@@ -482,7 +482,7 @@ public class MetricsServiceImpl implements MetricsService {
                 // eventually want to implement more fine-grained error handling where we can
                 // notify the subscriber of what exactly fails.
                 List<Observable<ResultSet>> updates = new ArrayList<>();
-                updates.add(dataAccess.insertIntoMetricsTagsIndex(metric, metric.getTags()));
+                updates.add(dataAccess.addTags(metric, metric.getTags()));
 
                 if (metric.getDataRetention() != null) {
                     updates.add(updateRetentionsIndex(metric));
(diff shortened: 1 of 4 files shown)


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -595,24 +595,23 @@ public void setDefaultTTL(int defaultTTL) {
...
-                .toList().map(l -> null);
+        return dataAccess.addTags(metric, tags).map(l -> null);
     }
 
     @Override
     public Observable<Void> deleteTags(Metric<?> metric, Set<String> tags) {
         return getMetricTags(metric.getMetricId())
                 .map(loadedTags -> {
-                    loadedTags.keySet().retainAll(tags);
+                    if (tags != null) {
```
jotak: If `tags` is null, then we should return an empty observable from the beginning. Or better: not call this function at all?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -1007,23 +1006,17 @@ public void shutdown() {
...
-
         //NOTE: compressed data is not deleted due to the using TWCS compaction strategy
         //      for the compressed data table.
-        result.mergeWith(dataAccess.deleteMetricData(id).flatMap(r -> null));
-
-        result.mergeWith(dataAccess.deleteMetricFromRetentionIndex(id).flatMap(r -> null));
-        result.mergeWith(dataAccess.deleteFromMetricExpirationIndex(id).flatMap(r -> null));
 
-        return result;
+        return getMetricTags(id)
```
jotak: What happens here if the metrics has no tag? Won't it skip the `flatMap` and thus not be deleted?

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jsanda: If tags is null, then I believe```retainAll``` throws an NPE.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jsanda: I think flatMap is still called because the observable returned from getMetricTags still emits an empty map.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jotak: Yes, but my point was more about: why would we call "deleteTags" with null tags? And if really we want to do so, then we could just return an empty observable at the beginning of "deleteTags" and thus avoid an unnecessary call to "getMetricsTags".

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
jotak: right :)
Btw I've just seen that it's a backport patch, so I suppose you can ignore my comments!

stefannegrea:
Closing since it is not needed now in this branch.

PR #830, opened 2017-05-25

jsanda:


jsanda:
retest this please

FilipB:
Just a reminder that haw-metrics-perf-stability-test is failing because this goes to hawkular:release/0.21.0. haw-metrics-perf-stability-test is only relevant for PRs for master

stefannegrea:
Thank you @jsanda!

PR #831, opened 2017-05-25

jsanda:


stefannegrea:
Thank you @jsanda!

PR #832, opened 2017-05-30
[HWKMETRICS-681] Expose internal metrics over Prometheus endpoint

rubenvp8510:
@jsanda  Could you please review? Thank you!

### DIFF
--- a/api/metrics-api-jaxrs/pom.xml
+++ b/api/metrics-api-jaxrs/pom.xml
@@ -52,6 +52,24 @@
   <dependencies>
 
     <dependency>
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient</artifactId>
+      <version>${version.io.prometheus}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.prometheus</groupId>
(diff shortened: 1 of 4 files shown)


rhqci:

jsanda:
Overall, the PR looks really good. There are a couple minor things I found.

I think that we only need to do the CollectorRegistry initialization in PrometheusHandler.java once and not per request. You can probably refactor that initialization code into a ```@PostConstruct``` method which of course also means that the CollectorRegistry will need to be an instance/member variable.

For the new maven dependencies that you have added in the metrics-api-jaxrs module, you need to add a variable in the hawkular-metrics/pom.xml file for the version number. Its name should be ```version.io.prometheus```. 

We need to figure out how we can configure which metrics to include/exclude. With the JMX exporter it can be done via configuration. So far it looks like we might have to implement something. If that turns out to be the case, then we can address it in a separate ticket/PR.

rubenvp8510:
This is ready for another review.

jsanda:
We can close since it was replaced with the work done for HWKMETRICS-682.

PR #837, opened 2017-06-28
[HWKMETRICS-688] Docker Container

gbaufake:
Hey Guys, 

It is a working progress of a docker container. 
It compiles the latest version of hawkular metrics. 

I think this might improve the way we deliver the latest code for new developers and might also flat the curve learning on delivering hawkular metrics + Cassandra on the same box.

Another benefits which I see:
- Most Actual Hawkular Metrics Code (Master Branch)
- Increase the testability by converting to a docker container

When next branch comes, it will need modifications for get vertex working there.

Best Regards,
Guilherme Baufaker Rêgo

### DIFF
--- a/ansible-build/README.md
+++ b/ansible-build/README.md
@@ -1,0 +1,45 @@
+
+# Hawkular Docker Container Ansible Playbook 
+
+This playbook generates a Hawkular Docker Image
+
+## Requirements
+
+- ``Git``
+- ``Maven``
+- ``Java``
+- ``Docker and Docker-py``
+- ``Ansible and Ansible-Playbook``
+
+
+## Usage
+
+* Clone the hawkular-metrics on a specific branch/tag/commit (If the desired commit/branch/tag doesn't have the `ansible-build` folder, it should be copied from another commit/branch/tag).
+
+* Run the Standard building command:
+
+`mvn-clean-install -DskipTests` 
+
+> if you want to make any change on the code, you can do it then run the `mvn command`
+
+
+* Once the building process finalized with success, it is necessary to export the following environment variables:
+
+ ``export HAWKULAR_METRICS_TAG='my-tag'``
+ ``export CASSANDRA_VERSION='3.0.14'``
+
+
+* If you need to deploy the image on Docker Hub, you will need to change the hawkular_metrics_image_name on ``roles/create_image/defaults/main.yml``, eg:
+
+``hawkular_metrics_image_name: "usename_on_docker_hub/hawkular-metrics"`` 
+
+
+* Finally, Run ``ansible-playbook -i hosts playbooks/hawkular.yml  -vvv`` to create a hawkular metrics docker container (Make sure that Docker is running and that you have all the requirements)
+
+
+* At the end of playbook running, a file called ``docker-compose.yml`` will be generated on ``ansible-build`` folder.
+
+
+* You may start Hawkular Metrics and Cassandra running `docker-compose up` inside ``ansible-build`` folder
+
+
--- a/ansible-build/ansible.cfg
+++ b/ansible-build/ansible.cfg
@@ -1,0 +1,2 @@
+[defaults]
+roles_path = roles/
--- a/ansible-build/hawkular.yml
+++ b/ansible-build/hawkular.yml
@@ -1,0 +1,21 @@
+#
+# Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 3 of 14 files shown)


rhqci:

docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install pip
+RUN curl  -o /opt/pip.py "https://bootstrap.pypa.io/get-pip.py"
+RUN python /opt/pip.py
+
+# Install CCM
+RUN pip install ccm
+
+# Start Hawkular Metrics Cassandra Cluster
+CMD ccm create -v 3.0.12 -n 1 -s hawkular
```
jotak: I'm not sure if cassandra should be part of the docker image. In other components, like hawkular-services, the docker image only contains the hawkular runnable, and C* is linked in a side docker-compose file : https://github.com/hawkular/hawkular-services/blob/master/docker-dist/docker-compose-template.yml

docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Clone the repository
+RUN git clone https://github.com/hawkular/hawkular-metrics.git /opt/hawkular-metrics
+
+# http, https, management, debug ports, Cassandra
+EXPOSE 8080 9990 8787 9042
+
+CMD cd /opt/hawkular-metrics/ && mvn clean install -DskipTests
```
jotak: Shouldn't be RUN instead of CMD? I thought only 1 CMD per dockerfile was allowed.

docker/Dockerfile
jsanda: Agreed. We do not want Cassandra to be part of the image. Ultimately, it would be great if we can update the build and release process to publish images. Overall, very glad to see this being worked on!

gbaufake: @jotak and @jsanda If we separate cassandra images, then I will need to migrate to docker-compose.

Since we have cassandra images for specific versions, I thought perhaps "embedding cassandra" may reduce the compatibility issues or enable test compatibility testing.

docker/Dockerfile
gbaufake: @jotak corrected on c43e5e0

gbaufake:
On point of view:

One requirement that is need to fulfill is the possibility to check to a specific version of Hawkular Metrics inside the container.

What do you think? @jotak, @jsanda

burmanm:
Like this?

```
[miburman@miranda hawkular_metrics]$ curl -s localhost:8080/hawkular/metrics/status | python -c 'import sys,json;print json.load(sys.stdin)["Implementation-Version"]'
0.28.0-SNAPSHOT
[miburman@miranda hawkular_metrics]$
```

jsanda:
We also write the version to the log file at start up, e.g.,

INFO  [org.hawkular.metrics.api.jaxrs.MetricsServiceLifecycle]
(ServerService Thread Pool -- 61) Hawkular Metrics version: 0.28.0-SNAPSHOT

gbaufake:
@burmanm Yes like this 
and to specific commit, tag or version inside the docker.

docker/Dockerfile
```diff
@@ -28,15 +28,13 @@ RUN python /opt/pip.py
...
 RUN git clone https://github.com/hawkular/hawkular-metrics.git /opt/hawkular-metrics
 
 # http, https, management, debug ports, Cassandra
 EXPOSE 8080 9990 8787 9042
 
-CMD cd /opt/hawkular-metrics/ && mvn clean install -DskipTests
-
 # Starting Hawkular Metrics
-CMD /opt/hawkular-metrics/integration-tests/rest-tests-jaxrs/target/wildfly-run/wildfly-10.0.0.Final/bin/standalone.sh
+CMD   cd /opt/hawkular-metrics/ && mvn clean install -DskipTests && /opt/hawkular-metrics/integration-tests/rest-tests-jaxrs/target/wildfly-run/wildfly-10.0.0.Final/bin/standalone.sh
```
jotak: I will test it on Monday, but I can see another issue here: `mvn install` will be performed every time the container is started, which will slow down the startup considerably.

burmanm: Not to mention the ccm creation, which will download Cassandra, install it (or compile it) and then initialize it..

This container will probably take ~10 minutes to start. Isn't this kinda defeating the whole idea? Also, the hawkular-metrics should probably be started from dist/standalone/wildfly-standalone/target/wildfly/wildfly-10.0.0.Final/bin/standalone.sh

jotak: For Cassandra, it's ok: "RUN" is executed at build-time, so will result as a layer in the image. It won't be executed at startup (as opposed to "CMD").

jotak: (just realized that it's a problem :-) )

jsanda: Does it make more sense to just copy artifacts from a maven repo? It could be local repo for snapshot/dev builds or from nexus. I do not understand why we would want to build hawkular-metrics here.

gbaufake: Thank for the comments. I will submit a new version of it on Monday!

gbaufake:
Please review this new version based on Ansible.

gbaufake:
@stefannegrea Please review it

stefannegrea:
@gbaufake, can you please leave Hawkular Alerting at version 1.7.0?

gbaufake:
@stefannegrea Done! I squashed the commits to a single one as well (rebased from latest master), in order to review it better.

Any other change, please tell me!
Best Regards,
Guilherme Baufaker Rêgo

stefannegrea:
Thank you @gbaufake!

PR #838, opened 2017-06-28
[HWKMETRICS-689] make collection interval configurable

jsanda:
The collection interval for internal metrics is interpreted in seconds and
defaults to 3 minutes.

I also fixed the flags for enabling internal metrics reporting so that they
are disabled by default.


stefannegrea:
Thank you @jsanda!

PR #841, opened 2017-07-11
[HWKMETRICS-699] Angular ui

jotak:
See https://issues.jboss.org/browse/HWKMETRICS-699

- Moved from React to Angular 4
- Use hawkular-charts
- Maven build
- Currently (temporarily) it's served as a side http service on 8081


jotak:
Note: to try it, the verticle named "UIVerticle" must be manually started. For the time being it runs on 8081.

mtho11:
@jotak for any UI related PRs can we get screenshots (or gif videos) so we can get a feel without having to install and run it ourselves.

mtho11:
This PR is getting pretty large now. Probably best to add the remaining tasks as separate PRs.

mtho11:
@jotak Hey, why don't we split off the UI into its own repo since it is using different technologies than metrics. We will be able to better manage the UI code this way (like eventually automated Travis builds/linting/etc...)

Heck, we can even repurpose the old UI repo: https://github.com/hawkular/hawkular-ui (or we can create a new one. I really don't care, but it should be in its own repo).

jotak:
Screenshots (it's actually almost the same as my previous PR with react)
http://imgur.com/ghRTJL8
http://imgur.com/1X4yT4R

jotak:
@mtho11 I'm fine to merge it without completing the task list. I will move that list to a JIRA

pilhuhn:
+1 on merge early, merge often.
Having a separate repo would have the advantage that commits of the UI code do not trigger full CI runs for the hawkular-metrics server code.

jotak:
Yes, but on the other hand, commits on the UI would require constantly to bump / npm-publish / update package.json here, so a more fastidious process.

jotak:
So now the webapp static content is pulled from another repo (or cloned, to be accurate). This approach has the advantage to be much faster: no need for node & npm local install, pulling all dependencies, building, etc. : everything is ready-to-use upstream.

Git clone can be configured in the pom.xml, with specific branch and/or tag/commit, quite similar to the srcdeps plugin, but it's the standard scm plugin used here.

So, upstream in hawkular-ui, the publish scenario would be:

1. push any changes to master
2. switch to release branch (which has a "dist" directory, unlike master where it's gitignored) and merge master in
3. cd metrics/ngapp && ng build -prod
4. commit and push to release (eventually tag)

So basically, the "release" branch would be just "master + dist".
Another option is to include dist in master so we don't have "release" at all, but it makes code review more painful.

mtho11:
@jotak I like your first proposal (to have a separate release branch) -- It does make the code reviews easier ;)

PR #842, opened 2017-07-11
[HWKMETRICS-696] Added local webfonts to prevent Internet connections on private enviroments

rubenvp8510:
Fixes HWKMETRICS-696


rhqci:

PR #846, opened 2017-07-13
Update the tenant filter to use project id's as tenant name

mwringe:
This updates the current filter to use ${OPENSHIFT_PROJECT_NAME}:${OPENSHIFT_PROJECT_ID} as the tenant name in Hawkular Metrics, instead of relying just on the project name

### DIFF
--- a/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
+++ b/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.openshift.auth.org.hawkular.openshift.namespace;
+
+import static io.undertow.util.StatusCodes.INTERNAL_SERVER_ERROR;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.StringJoiner;
+
+import org.hawkular.openshift.auth.Utils;
+import org.jboss.logging.Logger;
+
+import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.HttpString;
+
+/**
+ *
+ */
+public class NamespaceHandler implements HttpHandler {
+
+    private static final Logger log = Logger.getLogger(NamespaceHandler.class);
+
+    private final HttpHandler containerHandler;
+
+    //
+    private static final String HAWKULAR_TENANT_HEADER_NAME = "Hawkular-Tenant";
+    private static final String TOKEN_FILE_LOCATION = "/var/run/secrets/kubernetes.io/serviceaccount/token";
+
+    private static final String KUBERNETES_MASTER_URL_SYSPROP = "KUBERNETES_MASTER_URL";
+    private static final String KUBERNETES_MASTER_URL_DEFAULT = "https://kubernetes.default.svc.cluster.local";
+    static final String KUBERNETES_MASTER_URL =
+            System.getProperty(KUBERNETES_MASTER_URL_SYSPROP, KUBERNETES_MASTER_URL_DEFAULT);
+    //
+
+    private final String token;
+
+    private final NamespaceListener namespaceListener;
+
+    private final Map<String,String> projectOverwrites;
+
+    public NamespaceHandler(HttpHandler containerHandler) {
+        log.info("Creating NamespaceHandler");
+        this.containerHandler = containerHandler;
+
+        try {
+            this.token = getToken();
+            log.info("Received token from " + TOKEN_FILE_LOCATION);
+        } catch (Exception e){
+            log.error("Error retrieving token from file " + TOKEN_FILE_LOCATION);
+            throw new RuntimeException(e);
+        }
+
+        namespaceListener = new NamespaceListener(token);
+        NamespaceOverrideMapper nom = new NamespaceOverrideMapper(token);
+        log.info("Received namespace overrides: " + nom.overrides);
+
+        this.projectOverwrites = nom.overrides;
+    }
+
+    private String getToken() throws IOException {
+        return new String(Files.readAllBytes(Paths.get(TOKEN_FILE_LOCATION)));
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange serverExchange) throws Exception {
+        String originalTenantName = getTenantName(serverExchange);
+        if (originalTenantName == null || originalTenantName.startsWith("_")) {
+            containerHandler.handleRequest(serverExchange);
+        } else {
+
+            StringJoiner namespaceName = new StringJoiner(",");
+            String[] splitted = originalTenantName.split(",");
+
+            for (String namespace: originalTenantName.split(",")) {
+                String namespaceID = namespaceListener.getNamespaceId(namespace);
+
+                if (namespaceID != null) {
+                    String tenantName;
+
+                    if (projectOverwrites.containsKey(namespaceID)) {
+                        tenantName = projectOverwrites.get(namespaceID);
+                    } else {
+                        tenantName = namespace + ":" + namespaceID;
+                    }
+                    namespaceName.add(tenantName);
+
+
+                } else {
+                    log.info("Could not determine a namespace id for namespace " + namespaceID + ". Cannot process request. Returning an INTERNAL_SERVER_ERROR.");
+                    Utils.endExchange(serverExchange, INTERNAL_SERVER_ERROR, "Could not determine a namespace id for namespace " + namespaceID);
+                    return;
+                }
+            }
+
+            serverExchange.getRequestHeaders().remove(HAWKULAR_TENANT_HEADER_NAME);
+            serverExchange.getRequestHeaders().put(new HttpString(HAWKULAR_TENANT_HEADER_NAME), namespaceName.toString());
+
+            containerHandler.handleRequest(serverExchange);
+        }
+    }
+
+    private String getTenantName(HttpServerExchange httpServerExchange) {
+        String tenantName = httpServerExchange.getRequestHeaders().getFirst(HAWKULAR_TENANT_HEADER_NAME);
+        return tenantName;
+    }
+
+    public void stop() {
+        if (namespaceListener != null) {
+            namespaceListener.stop();
+        }
+    }
+}
--- a/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
+++ b/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.hawkular.openshift.auth.org.hawkular.openshift.namespace;
+
+import java.io.InputStream;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.net.ssl.HttpsURLConnection;
+
+import org.jboss.logging.Logger;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ *
+ */
+public class NamespaceListener {
+
+    private static final Logger log = Logger.getLogger(NamespaceListener.class);
+
+    // The authentication token used by the filter to connect to the OpenShift Master API
+    private final String token;
+
+    // Maps the namespace name to the namespace id
+    private final Map<String, String> namespaces;
+
+    private boolean run = true;
+
+    /**
+     *
+     * @param token The token to be used by the
+     */
+    public NamespaceListener(String token) {
+        log.debug("Creating NamespaceListener");
+        this.token = token;
+
+        namespaces = new HashMap<>();
+        log.debug("Namespace listener starting.");
+        new Thread(new Listener()).start();
+    }
+
+    public void stop() {
+        log.debug("Namespace listener shutting down.");
+        run = false;
+    }
+
+    /**
+     * Returns the namespace id for a particular namespace name
+     * @param namespaceName The name of the namespace
+     * @return The id of the namespace or null if the namespace does not exist
+     */
+    public String getNamespaceId(String namespaceName) {
+        return namespaces.get(namespaceName);
+    }
+
+    private class Listener implements Runnable {
+
+        @Override
+        public void run() {
+            while(run && !Thread.currentThread().isInterrupted()) {
+                try {
+                    String path = "api/v1/namespaces";
+                    URL url = new URL(NamespaceHandler.KUBERNETES_MASTER_URL + "/" + path + "?watch=true");
+
+                    HttpsURLConnection connection = (HttpsURLConnection) url.openConnection();
+                    connection.setRequestMethod("GET");
+
+                    connection.setRequestProperty("Accept", "application/json");
+                    connection.setRequestProperty("Authorization", "Bearer " + token);
+
+                    connection.setConnectTimeout(5000);
+
+                    // Since the connection is left open for new events, set to 0 which indicates an infinite timeout
+                    // Note that the server will eventually close this connection, so we will need to reconnect when this happens
+                    connection.setReadTimeout(0);
+
+                    connection.connect();
+                    int responseCode = connection.getResponseCode();
+
+                    if (responseCode == 200) {
+
+                        JsonFactory jsonFactory = new JsonFactory();
+                        jsonFactory.setCodec(new ObjectMapper());
+
+                        InputStream inputStream = connection.getInputStream();
+                        JsonParser jsonParser = jsonFactory.createParser(inputStream);
+
+                        // this will fetch all the available tenants and will then wait until new tenant events occur
+                        JsonNode jsonNode = jsonParser.readValueAsTree();
+
+                        // When the connection is closed, the response is null
+                        while (jsonNode != null) {
+                            JsonNode metaData = jsonNode.get("object").get("metadata");
+                            String projectName = metaData.get("name").asText();
+                            String projectID = metaData.get("uid").asText();
+
+                            String action = jsonNode.get("type").asText();
+
+                            if (action.equals("ADDED")) {
+                                log.debugf("Project %s [%s] has been added.", projectName, projectID);
+                                namespaces.put(projectName, projectID);
+                            } else if (action.equals("DELETED")) {
+                                log.debugf("Project %s [%s] has been removed.", projectName, projectID);
+                                namespaces.remove(projectName, projectID);
+                            }
+
+                            // this will read the next available tenant if available or wait until the next available
+                            // tenant event occurs
+                            jsonNode = jsonParser.readValueAsTree();
+                        }
+
+                        jsonParser.close();
+                        inputStream.close();
+                        namespaces.clear();
+
+                    } else {
+                        log.warnf("Error getting metadata. Response code %s", responseCode);
+                        try {
+                            // try again after a second
+                            Thread.sleep(1000);
+                        } catch (InterruptedException e) {
+                            log.error(e);
+                            stop();
+                        }
+                    }
+
+                } catch (Exception e) {
+                    log.error(e);
+                    try {
+                        // try again after a second
+                        Thread.sleep(1000);
+                    } catch (InterruptedException ie) {
+                        log.error(ie);
+                        stop();
+                    }
+                }
+            }
+        }
+    }
+}
--- a/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
+++ b/containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
@@ -1,0 +1,327 @@
+/*
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.openshift.auth.org.hawkular.openshift.namespace;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+
+import org.jboss.logging.Logger;
+
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.JdkSSLOptions;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.Row;
+import com.datastax.driver.core.SSLOptions;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.TokenRange;
+import com.datastax.driver.core.exceptions.InvalidQueryException;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ *
+ */
+public class NamespaceOverrideMapper {
+
+    private static final Logger log = Logger.getLogger(NamespaceOverrideMapper.class);
+
+    private Session session;
+
+    private final String token;
+
+    public Map<String,String> overrides;
+
+    public NamespaceOverrideMapper(String token) {
+        log.debug("Creating NamespaceOverrideMapper");
+
+        this.token = token;
+
+        boolean connected = false;
+        do {
+            try {
+                session = createSession();
+                connected = true;
+            } catch (Exception e) {
+                log.info("Could not connect to Cassandra. This could mean Cassandra is not yet up and running. Will try again. Error message: " + e.getLocalizedMessage());
+                try {
+                    Thread.sleep(2000);
+                } catch (InterruptedException e1) {
+                    e1.printStackTrace();
+                }
+            }
+        } while (!connected);
+        log.debug("Connection to Cassandra establised");
+
+        // ensure that the openshift_metrics keyspace exists
+        ensureKeyspace("openshift_metrics");
+
+        // if we already have a mappingTable stored in OpenShift, then use it.
+        // Otherwise we need to generate the overwrite table.
+        if (hasMappingTable()) {
+            log.debug("Mapping table already exists in Cassandra. Using this table");
+            this.overrides = getMappingTable();
+        } else {
+            log.debug("Mapping table does not exist in Cassandra. Creating Mapping");
+            // get the current list of tenants in Hawkular Metrics
+            Set<String> tenants = getTenants();
+
+            // map the current tenants to the current projects
+            Map<String, String> namespaceMapping = getNamespaceMapping(tenants);
+            log.debug("Namespace mapping: " + namespaceMapping);
+
+            this.overrides = namespaceMapping;
+
+            // write this mapping table back to Cassandra. We don't want to ever recalculate this table
+            log.debug("Writing the mapping table to Cassandra");
+            writeMappingTable(this.overrides);
+            log.debug("Mapping table stored in Cassandra");
+        }
+    }
+
+    private Session createSession() {
+        Cluster.Builder clusterBuilder = new Cluster.Builder();
+
+        String nodes = System.getProperty("hawkular.metrics.cassandra.nodes", "hawkular-cassandra");
+        Arrays.stream(nodes.split(",")).forEach(clusterBuilder::addContactPoint);
+
+        if (System.getProperty("hawkular.metrics.cassandra.use-ssl") != null && !System.getProperty("hawkular.metrics.cassandra.use-ssl").equals("false")) {
+            SSLOptions sslOptions = null;
+            try {
+                String[] defaultCipherSuites = {"TLS_RSA_WITH_AES_128_CBC_SHA", "TLS_RSA_WITH_AES_256_CBC_SHA"};
+                sslOptions = JdkSSLOptions.builder().withSSLContext(SSLContext.getDefault())
+                        .withCipherSuites(defaultCipherSuites).build();
+                clusterBuilder.withSSL(sslOptions);
+            } catch (NoSuchAlgorithmException e) {
+                throw new RuntimeException("SSL support is required but is not available in the JVM.", e);
+            }
+        }
+
+        Cluster cluster = clusterBuilder.build();
+        cluster.init();
+
+        Session session = cluster.connect();
+
+        return session;
+    }
+
+    /**
+     * Creates a keyspace if it does not already exist
+     * @param keyspace The name of the keyspace
+     */
+    private void ensureKeyspace(String keyspace) {
+        session.execute("CREATE KEYSPACE IF NOT EXISTS " + keyspace + " WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;");
+    }
+
+    private boolean hasMappingTable() {
+        if (session.getCluster().getMetadata().getKeyspace("openshift_metrics").getTable("metrics_mappings") == null) {
+            log.debug("metrics_mappings table does not exist");
+            return false;
+        } else {
+            log.debug("metrics_mapping table does exist");
+        }
+
+        ResultSet resultset = session.execute("SELECT project_name FROM openshift_metrics.metrics_mappings WHERE project_id='%succeeded'");
+
+        Row row = (resultset == null) ? null: resultset.one();
+        String succeeded = (row == null) ? "false" : row.getString("project_name");
+
+        if (succeeded.equalsIgnoreCase("true")) {
+            log.debug("metrics_mapping table succeeded when it was created.");
+            return true;
+        } else {
+            log.debug("metrics_mapping table was created, but it did not succeed the last time.");
+            return false;
+        }
+    }
+
+    private void createMappingTable() {
+        session.execute("CREATE TABLE IF NOT EXISTS openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text) WITH compaction = { 'class' : 'LeveledCompactionStrategy' };");
+    }
+
+    private Map<String,String> getMappingTable() {
+        // You only want to prepare the query once. Best to do it when you initialize the session.
+        PreparedStatement findMappings = session.prepare(
+                "SELECT project_id, project_name " +
+                        "FROM openshift_metrics.metrics_mappings " +
+                        "WHERE token(project_id) > ? AND token(project_id) <= ?");
+
+        Map<String,String> mappings = new HashMap<>();
+        if (hasMappingTable()) {
+            for (TokenRange tokenRange : getTokenRanges()) {
+                BoundStatement boundStatement = findMappings.bind().setToken(0, tokenRange.getStart())
+                        .setToken(1, tokenRange.getEnd());
+                ResultSet resultSet = session.execute(boundStatement);
+                resultSet.forEach(row -> mappings.put(row.getString(0), row.getString(1)));
+            }
+        }
+
+        mappings.remove("%succeeded");
+        return mappings;
+    }
+
+    private Set<TokenRange> getTokenRanges() {
+        Set<TokenRange> tokenRanges = new HashSet<>();
+        for (TokenRange tokenRange : session.getCluster().getMetadata().getTokenRanges()) {
+            tokenRanges.addAll(tokenRange.unwrap());
+        }
+        return tokenRanges;
+    }
+
+    private void writeMappingTable(Map<String,String> mappings) {
+        createMappingTable();
+
+        PreparedStatement pStatement = session.prepare("INSERT INTO openshift_metrics.metrics_mappings (project_id, project_name) VALUES(?,?);");
+
+        for (Map.Entry<String,String> entry: mappings.entrySet()) {
+            int maxRetries = 3;
+            int tries = 0;
+            boolean succeeded = false;
+            while (!succeeded && tries < maxRetries) {
+                try {
+                    log.debugf("trying to write value to metrics_mapping table: %s %s", entry.getKey(), entry.getValue());
+                    session.execute(pStatement.bind(entry.getKey(), entry.getValue()));
+                    log.debugf("wrote value to metrics_mapping table");
+                    succeeded = true;
+                } catch (Exception e) {
+                    log.errorf(e, "Error trying to insert entry into the openshift_metrics.metrics_mappings table (%s:%s). Will retry.", entry.getKey(), entry.getValue());
+                    try {
+                        Thread.sleep(1000);
+                    } catch (InterruptedException ie) {
+                        break;
+                    }
+                }
+                tries++;
+            }
+
+            if (!succeeded) {
+                throw new RuntimeException("Could not write all values to the openshift_metrics.metrics_mapping table. Aborting");
+            }
+        }
+
+        session.execute(pStatement.bind("%succeeded", "true"));
+    }
+
+    private Set<String> getTenants() {
+        Set<String> tenants = new HashSet<>();
+        ResultSet resultset = session.execute("SELECT * FROM system_schema.keyspaces WHERE keyspace_name = 'hawkular_metrics';");
+        if (!resultset.iterator().hasNext()) {
+            return tenants;
+        }
+
+        try {
+            // An invalid query exception will occur if the table does not exists.
+            // If the table does not exist, then no tenants have been stored yet so we just return the empty tenant set.
+            ResultSet resultSet = session.execute("SELECT DISTINCT tenant_id,type from hawkular_metrics.metrics_idx;");
+
+            Iterator<Row> ri = resultSet.iterator();
+            while (ri.hasNext()) {
+                Row row = ri.next();
+                String tenant = row.getString("tenant_id");
+                if (!tenant.startsWith("_") && !tenant.contains(":")) {
+                    tenants.add(tenant);
+                }
+            }
+        } catch (InvalidQueryException iqe) {
+            log.warn(iqe);
+        }
+        return tenants;
+    }
+
+    private Map<String,String> getNamespaceMapping(Set<String> tenants) {
+        Map<String,String> namespaceMap = new HashMap<>();
+
+        if (tenants == null || tenants.isEmpty()) {
+            return namespaceMap;
+        }
+
+        try {
+            String path = "api/v1/namespaces";
+            URL url = new URL(NamespaceHandler.KUBERNETES_MASTER_URL + "/" + path);
+
+            HttpsURLConnection connection = (HttpsURLConnection) url.openConnection();
+            connection.setRequestMethod("GET");
+
+            connection.setRequestProperty("Accept", "application/json");
+            connection.setRequestProperty("Authorization", "Bearer " + token);
+
+            // set to 0 which indicated an infinite timeout
+            connection.setConnectTimeout(0);
+            connection.setReadTimeout(0);
+
+            connection.connect();
+            int responseCode = connection.getResponseCode();
+
+            if (responseCode == 200) {
+
+                JsonFactory jsonFactory = new JsonFactory();
+                jsonFactory.setCodec(new ObjectMapper());
+
+                InputStream inputStream = connection.getInputStream();
+                JsonParser jsonParser = jsonFactory.createParser(inputStream);
+
+                JsonNode jsonNode = jsonParser.readValueAsTree();
+                // When the connection is closed, the response is null
+                if (jsonNode != null) {
+                    JsonNode items = jsonNode.get("items");
+                    if (items != null && items.isArray()) {
+                        for (JsonNode project : items) {
+                            JsonNode metaData = project.get("metadata");
+                            String projectName = metaData.get("name").asText();
+                            String projectID = metaData.get("uid").asText();
+
+                            if (tenants.contains(projectName)) {
+                                namespaceMap.put(projectID, projectName);
+                            }
+                        }
+                    }
+                }
+
+                jsonParser.close();
+                inputStream.close();
+
+            } else {
+                log.error("Error getting metadata from the OpenShift Master (" + responseCode + "). This may mean the 'hawkular' user does not have permission to watch projects. Waiting 2 seconds and will try again.");
+                try {
+                    Thread.sleep(2000);
+                } catch (InterruptedException e1) {
+                    e1.printStackTrace();
+                }
+            }
+
+
+        } catch (IOException e) {
+            log.error(e);
+        }
+
+        return namespaceMap;
+    }
+}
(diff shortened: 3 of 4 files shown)


mwringe:
@jsanda: can you please take a look?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
```diff
@@ -0,0 +1,128 @@
...
+                    String tenantName;
+
+                    if (projectOverwrites.containsKey(namespaceID)) {
+                        tenantName = projectOverwrites.get(namespaceID);
+                    } else {
+                        tenantName = splitted[i] + ":" + namespaceID;
+                    }
+
+                    if (i < splitted.length - 1) {
+                        tenantName += ",";
```
burmanm: This creates garbage I think.. in Java 8, use ``String.join`` or ``StringJoiner`` (simplifies the code also)

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     *
+     * @param token The token to be used by the
+     */
+    public NamespaceListener(String token) {
+        log.info("Creating NamespaceListener");
+        this.token = token;
+
+        namespaces = new HashMap<String, String>();
```
burmanm: <String, String> is redundant

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private String getToken() throws IOException {
+        return new String(Files.readAllBytes(Paths.get(TOKEN_FILE_LOCATION)));
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange serverExchange) throws Exception {
+        String originalTenantName = getTenantName(serverExchange);
+        if (originalTenantName == null || originalTenantName.startsWith("_")) {
```
burmanm: Has `_` been a restricted part in the project name previously?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public NamespaceHandler(HttpHandler containerHandler) {
+        log.info("Creating NamespaceHandler");
+        this.containerHandler = containerHandler;
+
+        try {
+            this.token = getToken();
+            log.info("Received token from " + TOKEN_FILE_LOCATION);
+        } catch (Exception e){
+            log.info("Error retrieving token from file " + TOKEN_FILE_LOCATION);
```
burmanm: log.error

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,149 @@
...
+        log.info("Namespace listener starting");
+        new Thread(new Listener()).start();
```
burmanm: How do you shutdown the thread?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,149 @@
...
+     */
+    public String getNamespaceId(String namespaceName) {
+        return namespaces.get(namespaceName);
+    }
+
+    private class Listener implements Runnable {
+
+        @Override
+        public void run() {
+            while(true) {
```
burmanm: The semantics of this thread are a bit unclear to me.. it runs forever (after the jsonNode is null, it restarts as there's no return anywhere) in an eternal loop (repolling as fast as it can). 

What's the blocking process here? Is ``jsonParser.readValueAsTree()`` blocking and waiting until data is available?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+        this.token = token;
+
+        boolean connected = false;
+        do {
+            try {
+                session = createSession();
+                connected = true;
+            } catch (Exception e) {
+                log.error("Could not connect to Cassandra. This could mean Cassandra is not yet up and running. Will try again.");
+                e.printStackTrace();
```
burmanm: This will spam the logs in the startup? Lets try to avoid e.printStackTrace()

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+    private static final Logger log = Logger.getLogger(NamespaceOverrideMapper.class);
+
+    private Session session;
+
+    private final String token;
+
+    public Map<String,String> overrides;
+
+    public NamespaceOverrideMapper(String token) {
+        log.info("Creating NamespaceOverrideMapper");
```
burmanm: log.debug

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,149 @@
...
+
+    // Maps the namespace name to the namespace id
+    private final Map<String, String> namespaces;
+
+    /**
+     *
+     * @param token The token to be used by the
+     */
+    public NamespaceListener(String token) {
+        log.info("Creating NamespaceListener");
```
burmanm: .debug

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,149 @@
...
+        this.token = token;
+
+        namespaces = new HashMap<String, String>();
+        log.info("Namespace listener starting");
```
burmanm: .debug

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+        // if we already have a mappingTable stored in OpenShift, then use it.
+        // Otherwise we need to generate the overwrite table.
+        if (hasMappingTable()) {
+            log.info("Mapping table already exists in Cassandra. Using this table");
+            this.overrides = getMappingTable();
+        } else {
+            log.info("Mapping table does not exist in Cassandra. Creating Mapping");
+            // get the current list of tenants in Hawkular Metrics
+            Set<String> tenants = getTenants();
```
burmanm: Is it guaranteed that this runs *after* the Hawkular-Metrics schema has been created?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+            // map the current tenants to the current projects
+            Map<String, String> namespaceMapping = getNamespaceMapping(tenants);
+            log.info("Namespace mapping: " + namespaceMapping);
```
burmanm: .debug

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+            this.overrides = namespaceMapping;
+
+            // write this mapping table back to Cassandra. We don't want to ever recalculate this table
+            log.info("Writing the mapping table to Cassandra");
```
burmanm: .debug and next one too

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+        }
+    }
+
+    private boolean hasMappingTable() {
+        ResultSet resultset = session.execute("SELECT table_name FROM system_schema.tables WHERE keyspace_name='openshift_metrics' AND table_name='metrics_mappings';");
+        return !resultset.all().isEmpty();
+    }
+
+    private void createMappingTable() {
+        ResultSet resultset = session.execute("SELECT table_name FROM system_schema.tables WHERE keyspace_name='openshift_metrics' AND table_name='metrics_mappings';");
```
burmanm: If there's already hasMappingTable, what's the point of rechecking?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+        if (resultset.all().isEmpty()) {
+            session.execute("CREATE TABLE openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text);");
```
burmanm: Are you sure about the primary key? Looking at the select queries, you might want to have both in the PK (as you never request without both)

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+        }
+
+        return mappings;
+    }
+
+    private void writeMappingTable(Map<String,String> mappings) {
+        createMappingTable();
+
+        StringBuffer buffer = new StringBuffer();
```
burmanm: ``StringBuilder`` ? I know JVM optimizes this, but still. This isn't thread-safe in any case.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+            session.execute("CREATE TABLE openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text);");
+        }
+    }
+
+    private Map<String,String> getMappingTable() {
+        Map<String,String> mappings = new HashMap<>();
+        if (hasMappingTable()) {
+            ResultSet resultSet = session.execute("SELECT project_id,project_name FROM openshift_metrics.metrics_mappings;");
+
+            for (Row row: resultSet.all()) {
```
burmanm: Is there a risk that this query requires PAGING instead of loading the whole table to memory?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+        buffer.append("BEGIN BATCH ");
+
+        for (Map.Entry<String,String> entry: mappings.entrySet()) {
+            buffer.append(" INSERT INTO openshift_metrics.metrics_mappings (project_id, project_name) VALUES('"+ entry.getKey() + "','" + entry.getValue() + "');");
+        }
+        buffer.append(" APPLY BATCH;");
+
+        ResultSet resultset = session.execute(buffer.toString());
+
+        System.err.println("WRITE RESULT SET");
```
burmanm: No System.err

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+        StringBuffer buffer = new StringBuffer();
+        buffer.append("BEGIN BATCH ");
+
+        for (Map.Entry<String,String> entry: mappings.entrySet()) {
+            buffer.append(" INSERT INTO openshift_metrics.metrics_mappings (project_id, project_name) VALUES('"+ entry.getKey() + "','" + entry.getValue() + "');");
+        }
+        buffer.append(" APPLY BATCH;");
+
+        ResultSet resultset = session.execute(buffer.toString());
```
burmanm: resultSet is never used, can be removed

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+
+        System.err.println("WRITE RESULT SET");
+    }
+
+    private Set<String> getTenants() {
+        Set<String> tenants = new HashSet<String>();
```
burmanm: ``HashSet<>()`` is enough

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+        ResultSet resultset = session.execute("SELECT * FROM system_schema.keyspaces WHERE keyspace_name = '" + "hawkular_metrics" + "';");
+        if (resultset.all().isEmpty()) {
+            return tenants;
+        }
+
+        ResultSet resultSet = session.execute("SELECT DISTINCT tenant_id,type from hawkular_metrics.metrics_idx;");
+
+        for (Row row:resultSet.all()) {
```
burmanm: ``resultSet.all()`` instead of iterator could return a very large batch of data. Not recommended.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,271 @@
...
+                try {
+                    Thread.sleep(2000);
+                } catch (InterruptedException e1) {
+                    e1.printStackTrace();
+                }
+            }
+
+
+        } catch (IOException e) {
+            System.err.println(e);
```
burmanm: No System.err

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
mwringe: It works fine for me with JDK 8. I will update it to use StringJoiner as it does simplify things.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
mwringe: Yep, it always has. We use it for our _system tenant, _hawkular_admin tenant, and is used by others for non project based tenants.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: Ok, I will have it print the localized message only

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: I would prefer to keep it as info so that we can see from the logs when its being started. But I can move it debug.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: Oops, that was a debug statement that made it into the PR. removing

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: removed

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: The REST endpoint for a watch will return all the current values and then it keeps the connection open until there is a change. Then it will send the change and continue to keep the connection open until something else changes.

The jsonParser.readValueAsTree() will be blocking here when the connection is left open waiting for changes.

I will add some comments to the code to make that clear.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: It shouldn't matter when its run. We currently check if the keyspace exists or not but we need to add some more checks here incase the keyspace exists but the the table has not yet been created.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: good point, I will update this to use hasMappingTable instead of duplicating the check

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: Its a table storing a key value pair, why would I want to make both key and value the PK? Especially if at some point I might want to query with the key to get the value.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok, will switch to paging.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: Currently we don't, it get shutdown when the server is shutdown. Adding up a stop method.

mwringe:
Ok, I updated with the requested changes.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+    private void createMappingTable() {
+        if (!hasMappingTable()) {
+            session.execute("CREATE TABLE openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text);");
+        }
+    }
+
+    private Map<String,String> getMappingTable() {
+        Map<String,String> mappings = new HashMap<>();
+        if (hasMappingTable()) {
+            ResultSet resultSet = session.execute("SELECT project_id,project_name FROM openshift_metrics.metrics_mappings;");
```
jsanda: I realize that we often deal with small Cassandra clusters, but this type of query which is essentially a distributed full table scan is generally avoided because it incurs more overhead as the cluster size increases, resulting in increasingly poor performance.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+        return mappings;
+    }
+
+    private void writeMappingTable(Map<String,String> mappings) {
+        createMappingTable();
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("BEGIN BATCH ");
+
+        for (Map.Entry<String,String> entry: mappings.entrySet()) {
```
jsanda: Do not use a batch statement here. You should instead use async writes. We can help you with that if want.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+        Set<String> tenants = new HashSet<>();
+        ResultSet resultset = session.execute("SELECT * FROM system_schema.keyspaces WHERE keyspace_name = '" + "hawkular_metrics" + "';");
+        if (!resultset.iterator().hasNext()) {
+            return tenants;
+        }
+
+        try {
+            // An invalid query exception will occur if the table does not exists.
+            // If the table does not exist, then no tenants have been stored yet so we just return the empty tenant set.
+            ResultSet resultSet = session.execute("SELECT DISTINCT tenant_id,type from hawkular_metrics.metrics_idx;");
```
jsanda: Unlike the other query which I said should be avoided, this query while it appears very similar, is often ok. You are only querying the partition key which is indexed and fast to query as it should not require touching SSTables.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: I think I also wanted to create the mapping table here as well but I didn't, which I guess I need to add to the batch part.

I want to make sure that that table and its metrics are all written together, or all fail. Is batch not the right approach to use here?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: so I need to pull out the whole result of this simple table, what would be a better approach then?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+            this.overrides = namespaceMapping;
+
+            // write this mapping table back to Cassandra. We don't want to ever recalculate this table
+            log.debug("Writing the mapping table to Cassandra");
+            writeMappingTable(this.overrides);
+            log.debug("Mapping table stored in Cassandra");
+        }
+    }
+
+    private Session createSession() {
```
jsanda: I am not sure if it is possible, but it would be better to reuse the same session that is used elsewhere in hawkular-metrics.

jsanda:
I don't know why I cannot find this comment on the github issue, but making
both key and value the primary key should make the type of query you are
doing more performant.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
jsanda: Creating the table is not done in the batch statement. And I don't think it can be part of a batch statement. Only the inserts are done via the batch statement. Individual async writes should be used.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    private boolean hasMappingTable() {
+        ResultSet resultset = session.execute("SELECT table_name FROM system_schema.tables WHERE keyspace_name='openshift_metrics' AND table_name='metrics_mappings';");
+        return resultset.iterator().hasNext();
+    }
+
+    private void createMappingTable() {
+        if (!hasMappingTable()) {
+            session.execute("CREATE TABLE openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text);");
```
jsanda: This table should probably use leveled compaction (LCS). Size tiered (STCS) is the default.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
burmanm: You should use TokenRange approach

burmanm: That way you could split the query to multiple parts (to multiple servers for example)

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+     */
+    public String getNamespaceId(String namespaceName) {
+        return namespaces.get(namespaceName);
+    }
+
+    private class Listener implements Runnable {
+
+        @Override
+        public void run() {
+            while(run) {
```
jsanda: I would change this to be ```while (run && !Thread.currentThread().isInterrupted())

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                        inputStream.close();
+                        namespaces.clear();
+
+                    } else {
+                        log.error("Error getting metadata. Response code " + responseCode);
+                        try {
+                            // try again after a second
+                            Thread.sleep(1000);
+                        } catch (InterruptedException e) {
+                            log.error(e);
```
jsanda: Best to call ```stop()``` here

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                        }
+                    }
+
+                } catch (IOException e) {
+                    log.error(e);
+                    try {
+                        // try again after a second
+                        Thread.sleep(1000);
+                    } catch (InterruptedException ie) {
+                        log.error(ie);
```
jsanda: Best to call ```stop()``` here

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+        log.debug("Creating NamespaceListener");
+        this.token = token;
+
+        namespaces = new HashMap<>();
+        log.debug("Namespace listener starting");
+        new Thread(new Listener()).start();
+    }
+
+    public void stop() {
+        run = false;
```
jsanda: It would be good to have an info message here that the listener is stopping.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                        log.error("Error getting metadata. Response code " + responseCode);
+                        try {
+                            // try again after a second
+                            Thread.sleep(1000);
+                        } catch (InterruptedException e) {
+                            log.error(e);
+                        }
+                    }
+
+                } catch (IOException e) {
```
jsanda: You should handle **all** exceptions here. If a RuntimeException is thrown without being handled the listener will die silently.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                        // When the connection is closed, the response is null
+                        while (jsonNode != null) {
+                            JsonNode metaData = jsonNode.get("object").get("metadata");
+                            String projectName = metaData.get("name").asText();
+                            String projectID = metaData.get("uid").asText();
+
+                            String action = jsonNode.get("type").asText();
+
+                            if (action.equals("ADDED")) {
+                                log.debug("Project " + projectName + "[" + projectID + "] has been added.");
```
jsanda: use ```debugf``` here instead and get rid of the string concatenation

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                                namespaces.put(projectName, projectID);
+                            } else if (action.equals("DELETED")) {
+                                log.debug("Project " + projectName + "[" + projectID + "] has been removed.");
```
jsanda: use ```debugf``` here instead and get rid of the string concatenation

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
```diff
@@ -0,0 +1,158 @@
...
+                            // tenant event occurs
+                            jsonNode = jsonParser.readValueAsTree();
+                        }
+
+                        jsonParser.close();
+                        inputStream.close();
+                        namespaces.clear();
+
+                    } else {
+                        log.error("Error getting metadata. Response code " + responseCode);
```
jsanda: Use ```errorf``` here instead and get rid of the string concatenation. Actually I would use ```warnf```.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+
+        this.token = token;
+
+        boolean connected = false;
+        do {
+            try {
+                session = createSession();
+                connected = true;
+            } catch (Exception e) {
+                log.error("Could not connect to Cassandra. This could mean Cassandra is not yet up and running. Will try again. Error message: " + e.getLocalizedMessage());
```
jsanda: Change this to ```info``.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+                try {
+                    Thread.sleep(2000);
+                } catch (InterruptedException e1) {
+                    e1.printStackTrace();
```
jsanda: We do not want to ignore InterruptedExceptions without good reason. If there is an interrupt, that probably means we should initiate shutdown.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+
+        return session;
+    }
+
+    /**
+     * Creates a keyspace if it does not already exist
+     * @param keyspace The name of the keyspace
+     */
+    private void ensureKeyspace(String keyspace) {
+        ResultSet resultset = session.execute("SELECT * FROM system_schema.keyspaces WHERE keyspace_name = '" + keyspace + "';");
```
jsanda: You can simply this method. You do not need to query system_schema.keyspaces. Use the ```IF NOT EXISTS``` clause, e.g.,

```
CREATE KEYSPACE IF NOT EXISTS foo WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}
```

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    private boolean hasMappingTable() {
+        ResultSet resultset = session.execute("SELECT table_name FROM system_schema.tables WHERE keyspace_name='openshift_metrics' AND table_name='metrics_mappings';");
+        return resultset.iterator().hasNext();
+    }
+
+    private void createMappingTable() {
+        if (!hasMappingTable()) {
```
jsanda: Same as creating the keyspace. Use the ```IF NOT EXISTS``` clause, e.g.,

```
CREATE TABLE IF NOT EXISTS openshift_metrics.metrics_mappings (project_id text PRIMARY KEY, project_name text)
```

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
```diff
@@ -0,0 +1,279 @@
...
+        if (hasMappingTable()) {
+            log.debug("Mapping table already exists in Cassandra. Using this table");
+            this.overrides = getMappingTable();
+        } else {
+            log.debug("Mapping table does not exist in Cassandra. Creating Mapping");
+            // get the current list of tenants in Hawkular Metrics
+            Set<String> tenants = getTenants();
+
+            // map the current tenants to the current projects
+            Map<String, String> namespaceMapping = getNamespaceMapping(tenants);
```
jsanda: ```getNamespaceMapping``` is potentially an expensive operation that could take some time to complete. If ```tenants``` is empty, can we skip it?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
jsanda: Now that I understand that this query is performed only once at start up I am not overly concerned about performance; however, it is susceptible to read timeouts. Taking this under consideration I think using TokenRange queries is the safer approach. There are two parts to this. First, get all of the token ranges. Secondly, do a query for each token range. 

Take a look at ```DataAccessImpl.getTokenRanges``` to see how to get all of the token ranges. 

Then you will want to iterate over the token ranges like this:

```java
for (TokenRange tokenRange : tokenRanges) {
    // This needs to be a prepared statement
    ResultSet resultSet = session.execute(
            "SELECT project_id, project_name " +
            "FROM openshift_metrics.metrics_mappings " +
            "WHERE token(project_id) > ? AND token(project_id) <= ?
}
```
and the parameters for the prepared statement will be ```TokenRange.getStart()``` and ```TokenRange.getEnd()``` respectively.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
```diff
@@ -0,0 +1,130 @@
...
+                    if (projectOverwrites.containsKey(namespaceID)) {
+                        tenantName = projectOverwrites.get(namespaceID);
+                    } else {
+                        tenantName = namespace + ":" + namespaceID;
+                    }
+                    namespaceName.add(tenantName);
+
+
+                } else {
+                    log.error("Could not determine a namespace id for namespace " + namespaceID + ". Cannot process request. Returning an INTERNAL_SERVER_ERROR.");
```
jsanda: I would reduce logging level to info here. Can you use the other version of ```endExchange``` that allows you to set a message in the response. The message could be the same as the log message.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
jsanda: I realize that populating the metrics_mappings table is a one time operation; however, you still do not want to use a batch statement here, particularly one that contains all of the writes. It is not that you need all writes to either succeed or fail, but rather some retry logic. There should be some retry logic in place. And after so many retries the writes still fail, then I would propagate an exception up the stack to prevent the server from starting since we are not in a good state. When the server restarts we will perform the writes again which is fine because they are idempotent.

jsanda:
@mwringe I have finished my review.

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceHandler.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: checking if its empty is basically the first thing that getNamespaceMapping does

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: Sorry, what do you want this to do exactly? like just throw a runtime exception here?

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
jsanda: No, I would call ```stop()```

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceListener.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok

containers/hawkular-openshift-security-filter/src/main/java/org/hawkular/openshift/auth/org/hawkular/openshift/namespace/NamespaceOverrideMapper.java
mwringe: ok, updated. I have also added a marked entry to this table (%succeeded) which if set to true meant that the table was properly updated last time. If its set to false (or the table does not exist) then it will be recreated on the next restart.

mwringe:
Ok, I pushed out an update with the latest required changed.

@jsanda: is there anything else we need to do in order to get this merged?

jsanda:
I would like to do at least a quick smoke test particularly since there are no automated tests. I can build your branch and get it deploy in my local openshift instance. Any suggestions on steps to take to do some basic verification?

mwringe:
I have been waiting for this pr to go through before submitting the origin-metrics pr and the openshift-ansible pr.

If you want to test this, I would build the ear and update the hawkular-metrics docker image to use that ear. Then grant the Hawkular Metrics user cluster-reader permissions (note: the ansible pr will create a much more refined permission level, we don't advocate running things as cluster level users, but its ok for this test).

Once you have those in place, you can test by doing the following
- have openshift running for a while in that its gathering metrics. You should have a couple of test projects, say 'foo' and 'bar' which having running pods that have metrics being gathered for them

- you will then need to scale down the Hawkular Metrics pod and have it use your custom pod with the changes.

- after this need Hawkular Metrics pod is running, you will need to make sure that all the current pods are still reporting metrics (even the metrics from before)

- curl the 'foo' tenant and get a list of the metrics it contains

- delete the 'foo' project, create a new 'foo' project and deploy a pod there so that metrics are being collected

- curl the 'foo' tenant again and make sure it only has metrics for the new pod and not for the old pod

You could also go more indepth and start to look at the tables in Cassandra, but I don't know if that will really help you or not.

mwringe:
Is there anything else which needs to be done before we can get this merged?

mwringe:
This has been updated to match what is in the 0.27 branch.

jsanda:
@mwringe did you rebase against the latest master? I restarted travis, but we are still seeing the same failure, which should be resolved and is unrelated to your PR.

PR #849, opened 2017-07-18
For Filip, #2

burmanm:


FilipB:
Exception is not visible after 5 hours

PR #850, opened 2017-07-18
[HWKMETRICS-695] Fix NPE and several other issues

jsanda:
HawkularObjectNameFactory previously assumed every metric would have meta
data. TokenAuthenticator was registering metrics without meta data. I have
updated TokenAuthenticator so that it creates meta data. I have also
updated HawkularObjectNameFactory to handle the case where there is no
meta data.

There is another important change around the meta data. If a metric does
not have meta data, then its tags and data points will not get persisted.
A warning will get logged by HawkularMetricsRegistryListener about the
lack of meta data.

I have refactored the way in which internal metrics are registered.
Previously everything was done eagerly at start up. Now only meta data
is registered eagerly. Registering meta data will not cause anything to
be persisted in Cassandra. Metrics are registered lazily upon access.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/dropwizard/RESTMetrics.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/dropwizard/RESTMetrics.java
@@ -16,19 +16,18 @@
  */
 package org.hawkular.metrics.api.jaxrs.dropwizard;
 
-import java.util.HashMap;
-import java.util.Map;
-
 import javax.enterprise.context.ApplicationScoped;
 
 import org.hawkular.metrics.api.jaxrs.util.MetricRegistryProvider;
 import org.hawkular.metrics.core.dropwizard.HawkularMetricRegistry;
+import org.hawkular.metrics.core.dropwizard.MetricNameService;
 
 import com.codahale.metrics.Timer;
 
 /**
- * Registers metrics for REST endpoints. The metrics here are based on the URI paths with parameters, not the path
- * variables. A client for example might send GET /hawkular/metrics/gauges/MyMetric/raw. The matching metric will be
+ * Registers metric meta data for REST endpoints. The actual metrics are registered when the end points are invoked for
+ * the first time. The metrics here are based on the URI paths with parameters, not the path variables. A client for
+ * example might send GET /hawkular/metrics/gauges/MyMetric/raw. The matching metric will be
  * GET gauges/id/raw. The base path of /hawkular/metrics is omitted.
  *
  * @author jsanda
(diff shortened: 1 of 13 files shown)


mwringe:
LGTM

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/dropwizard/RESTMetrics.java
```diff
@@ -16,118 +16,122 @@
...
+    public void setMetricNameService(MetricNameService metricNameService) {
+        this.metricNameService = metricNameService;
+    }
+
+    public void initMetrics() {
+        String hostname = metricNameService.getHostName();
+
+        register(RESTMetaData.forRead(HTTPMethod.GET, "status", hostname));
+
+        register(RESTMetaData.forWrite(HTTPMethod.POST, "gauges", hostname));
```
burmanm: While this is like before.. this is horrible. What are the odds that these go out-of-sync with the implementation? Could we autocreate these with annotation processor?

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
```diff
@@ -66,7 +66,7 @@
     ADMIN_TENANT("hawkular.metrics.admin-tenant", "admin", "ADMIN_TENANT", false),
     METRICS_REPORTING_HOSTNAME("hawkular.metrics.reporting.hostname", null, "METRICS_REPORTING_HOSTNAME", false),
     METRICS_REPORTING_ENABLED("hawkular.metrics.reporting.enabled", null, "METRICS_REPORTING_ENABLED", true),
-    METRICS_REPORTING_COLLECTION_INTERVAL("hawkular.metrics.reporting.collection-interval", "180",
+    METRICS_REPORTING_COLLECTION_INTERVAL("hawkular.metrics.reporting.collection-interval", "300",
```
burmanm: Why 5 minutes instead of 3 minutes? Ticket does not mention this change.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/HawkularMetricsRegistryListener.java
```diff
@@ -68,56 +70,147 @@ public void setMetricNameService(MetricNameService metricNameService) {
...
+        MetaData metaData = metricRegistry.getMetaData(name);
+        if (metaData == null) {
+            logger.warnf("Did not find meta data for histogram %s that has been added to the registry", name);
+            return;
+        }
+
+        String tenantId = metricNameService.getTenantId();
+        String metricName = metricNameService.createMetricName(metaData);
+
+        createMetrics(metricName, asList(
```
burmanm: Why aren't these percentiles just calculated by metrics when needed?

jsanda: To be honest, I am not entirely sure what we want to store from the aggregated DropWizard metrics like timers and meters. I do not however think we need to figure that out in this PR. We could just store the raw gauge and counter values, but then that begs the question of why use the DropWizards metrics library at all.

 I think there is value in storing what are essentially pre-computed percentiles. We can easily view graphs of those percentiles with the Grafana plugin or the new UI being developed. We can also compute the min, max, avg of p99 for instance. We can make changes to do these things ad hoc, but it is not currently possible.

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
jsanda: 5 minutes is arbitrary just as 3 minutes was arbitrary. I want to error on the side of caution with respect to how much of internal metrics we store.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/dropwizard/RESTMetrics.java
jsanda: Yeah, these need to be dynamically generated. The endpoints are not going to change for the next 0.27.x release which is main concern right now. I would prefer to do this in a follow ticket because I need to get this merged asap.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/HawkularObjectNameFactory.java
```diff
@@ -51,9 +50,15 @@ private boolean needsQuote(String propertyValue){
...
             Hashtable<String, String> table = new Hashtable<>();
-            table.putAll(metaDataMap.get(name).getTags());
-            table.replaceAll((k,v) -> needsQuote(v) ? ObjectName.quote(v): v );
-            return new ObjectName(domain, table);
+            MetaData metaData = metricRegistry.getMetaData(name);
+            if (metaData == null) {
+                LOGGER.warnf("No meta data found for %s", name);
+                return new ObjectName(domain, type, name);
+            } else {
+                table.putAll(metricRegistry.getMetaData(name).getTags());
```
burmanm: This part of the code is a bit confusing for reader. First, there's already ``metricRegistry.getMetadata(name)``. That's in the ``metaData`` object. Also, why first do ``putAll`` and then ``replaceAll`` ? Makes the reader think there might be something else in the table also.

I would personally rather see it like: ``metaData.stream().forEach(needsQuote..(); put)`` or something that indicates the flow more clearly.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/HawkularMetricsRegistryListener.java
```diff
@@ -154,4 +275,15 @@ private void createMetrics(String metricName, List<Metric<?>> metrics) {
                 );
     }
 
+    private void deleteMetrics(String metricName, List<MetricId<?>> metrics) {
```
burmanm: Why ``List<MetricId<?>>`` and not ``Observable<MetricId<?>>`` ? The above onCounterRemoved etc should just instead of creating a List create an Observable.

burmanm: Also, what's the point of metrIcName here?

burmanm: This could also be a ``Transformer`` or something else reusable that's then applied to the streams from the ``onMetricType`` methods. Would look more reactive.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/HawkularObjectNameFactory.java
jsanda: The ObjectName constructor unfortunately takes a Hashtable and not a Map; so, we are forced into copying the tags into ```table```.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/dropwizard/HawkularMetricsRegistryListener.java
jsanda: > what's the point of metricName here?

Yes it is just for logging. It is the same for each of the constituent metrics whereas the metric id is of course unique for each one.

I agree about the refactoring, but I am going to hold off for the moment. I am going to introduce another ticket that will involve some changes to this code including the remove of some of it. It makes more sense to do the refactoring there.

stefannegrea:
Thank you @jsanda!

PR #852, opened 2017-07-20
[HWKMETRICS-695] Fix NPE and several other issues

jsanda:
HawkularObjectNameFactory previously assumed every metric would have meta
data. TokenAuthenticator was registering metrics without meta data. I have
updated TokenAuthenticator so that it creates meta data. I have also
updated HawkularObjectNameFactory to handle the case where there is no
meta data.

There is another important change around the meta data. If a metric does
not have meta data, then its tags and data points will not get persisted.
A warning will get logged by HawkularMetricsRegistryListener about the
lack of meta data.

I have refactored the way in which internal metrics are registered.
Previously everything was done eagerly at start up. Now only meta data
is registered eagerly. Registering meta data will not cause anything to
be persisted in Cassandra. Metrics are registered lazily upon access.


jsanda:
@burmanm can you review this?

stefannegrea:
Merging since it's contains the same changes as PR #850

stefannegrea:
Thank you @jsanda!

PR #854, opened 2017-07-24
[HWKMETRICS-597] Add backpressure to the RxSession scheduling

burmanm:
If there are more than 1024 inflight requests on any connected host, wait until scheduling more statements.

### DIFF
--- a/core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
+++ b/core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,13 +17,17 @@
 
 package org.hawkular.rx.cassandra.driver;
 
+import java.util.function.Function;
+
 import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Host;
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.RegularStatement;
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.ResultSetFuture;
 import com.datastax.driver.core.Row;
 import com.datastax.driver.core.Session;
+import com.datastax.driver.core.SimpleStatement;
 import com.datastax.driver.core.Statement;
 import com.google.common.util.concurrent.ListenableFuture;
 
@@ -38,6 +42,7 @@ import rx.schedulers.Schedulers;
 public class RxSessionImpl implements RxSession {
 
     private Session session;
+    private static int MAXIMUM_INFLIGHT_REQUESTS = 1024;
 
     public RxSessionImpl(Session session) {
         this.session = session;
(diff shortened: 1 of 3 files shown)


burmanm:
Rebased from master

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
```diff
@@ -38,6 +42,7 @@
 public class RxSessionImpl implements RxSession {
 
     private Session session;
+    private static int MAXIMUM_INFLIGHT_REQUESTS = 1024;
```
stefannegrea: Why 1024, and not larger or smaller?

burmanm: Currently this is what we're using with the driver (after more than 1024, we start pushing to the buffer pool of the driver). 

And I couldn't find a way to request the current setting from anywhere (and DataAccessImpl can't see anything else than Session)

burmanm:
Changed to follow our settings (you can modify the maxRequests with -Dhawkular.metrics.cassandra.max-requests-per-connection) and rebased from master

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
```diff
@@ -38,9 +43,14 @@
 public class RxSessionImpl implements RxSession {
 
     private Session session;
+    private int maximumInflightRequests;
 
     public RxSessionImpl(Session session) {
         this.session = session;
+        maximumInflightRequests = Math.min(
```
jsanda: The max number of inflight requests overall is num_hosts * max_connections_per_host * max_requests_per_connection

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
```diff
@@ -54,10 +64,32 @@ public RxSession init() {
...
+    public static Function<Session, Integer> mostInFlightRequests = (session) -> {
+        int inFlights = 0;
+        for (Host host : session.getState().getConnectedHosts()) {
+            inFlights = Math.max(inFlights, session.getState().getInFlightQueries(host));
+        }
+        return inFlights;
+    };
+
+    private Observable<ResultSet> scheduleStatement(Statement st, Scheduler scheduler) {
+        while(true) {
```
jsanda: Would it be better to use a semaphore here?

jsanda:
@burmanm I added some comments on this yesterday.

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
burmanm: I found this method faster in my benchmarks.

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
burmanm: We don't know what host is being targeted in this case and I'm not sure if there's a cheap way to find out either (we would have to maintain a list of TokenRanges per host and calculate hash for each query and then check which hosts that might match)

jsanda: I am not sure I understand. ```maximumInFlightRequests``` is set to the smaller of the max local or remote connections, but each connection supports concurrent requests. Why is the max requests per connection not taken into account?

>We don't know what host is being targeted

Actually we can easily determine that with ```LoadBalancingPolicy.newQueryPlan(String loggedKeyspace, Statement statement)``` but I do not know that it is necessary. I think it is sufficient to see if there is an available connection with **any** host. I say this because we use a TokenAwarePolicy with a child load balancing policy of DCAwareRoundRobinPolicy. Even if resources are not available to send the query directly to the replica, I believe the driver will fall back to a non-replica node.

core/rx-java-driver/src/main/java/org/hawkular/rx/cassandra/driver/RxSessionImpl.java
jsanda: Do you have results or tests that you can share?

burmanm:
I modified it to do the lookup now and use maxConnections * maxRequests. We should lower the maxrequests per connection to a lower value though (256 should be fine).

burmanm:
To the question of "semaphores", in this solution I don't think they're a good approach as that would create two layers of throttling management - the complexities in maintaining the semaphores and their counts matched to the potentially changing Hosts and TokenRanges if not in my opinion worth it.

I also did some benchmarks on Semaphore vs. while(true) { Thread.sleep(0,1) } and found no measurable difference in CPU overhead. So I think leaving the inflight count management just to the client makes more sense.

jsanda:
I am inclined to think that the busy loop approach will incur more CPU overhead some times just on the basis of the potential constant looping. We should be probably do some more extensive testing to see the impact.

>To the question of "semaphores", in this solution I don't think they're a good approach as that would create two layers of throttling management - the complexities in maintaining the semaphores and their counts matched to the potentially changing Hosts and TokenRanges if not in my opinion worth it.

Can you explain this? A semaphore would also address a race condition that is not handled. It is entirely possible for multiple threads to get back a result of true from ```availableInFlightSlots```, and there won't be available resources for one or more of those threads.

burmanm:
Because it's not really busy looping, it's yielding the thread with that ``Thread.sleep(0, 1)`` to the scheduler. On my machine, there was zero difference in CPU usage compared to the semaphores (otherwise the performance would suffer when Cassandra and Hawkular-Metrics are running on the same node).

And the same would have to happen with semaphore**s** also. You would need to have one semaphore per available host and loop between them to see which one has an open slot. If you block to the first one, the driver will always wait for that same node (regardless of where the writes are actually going, so the semaphore could be full without a single write going to that node).

PR #855, opened 2017-07-24
Add missing DataAccess.shutdown() requests

burmanm:
For tests only, avoids multiple SchemaChangeListeners at the same time

PR #856, opened 2017-07-26
Change Cassandra settings for Travis for faster tests

burmanm:
This removes the fsync from the schema changes, so it speeds up the Cassandra's operation in the testing environment considerably.

### DIFF
--- a/core/schema/src/main/resources/org/hawkular/schema/bootstrap.groovy
+++ b/core/schema/src/main/resources/org/hawkular/schema/bootstrap.groovy
@@ -48,8 +48,14 @@ if (reset) {
   executeCQL("DROP KEYSPACE IF EXISTS $keyspace", 20000)
 }
 
-executeCQL("CREATE KEYSPACE IF NOT EXISTS $keyspace WITH replication = {'class': 'SimpleStrategy', " +
-    "'replication_factor': $replicationFactor}")
+def keyspaceCreate = "CREATE KEYSPACE IF NOT EXISTS $keyspace WITH replication = {'class': 'SimpleStrategy', " +
+    "'replication_factor': $replicationFactor}"
+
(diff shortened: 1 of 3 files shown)


burmanm:
This is based on the PR #855  also

stefannegrea:
@burmanm, can you please rebase on master?

burmanm:
Done, although I'm not sure why (there was no conflict).

core/schema/src/main/resources/org/hawkular/schema/bootstrap.groovy
```diff
@@ -48,8 +48,14 @@ if (reset) {
   executeCQL("DROP KEYSPACE IF EXISTS $keyspace", 20000)
 }
 
-executeCQL("CREATE KEYSPACE IF NOT EXISTS $keyspace WITH replication = {'class': 'SimpleStrategy', " +
-    "'replication_factor': $replicationFactor}")
+def keyspaceCreate = "CREATE KEYSPACE IF NOT EXISTS $keyspace WITH replication = {'class': 'SimpleStrategy', " +
+    "'replication_factor': $replicationFactor}"
+
+if (keyspace.equals("hawkulartest") || keyspace.equals("hawkular_metrics_rest_tests")) {
```
stefannegrea: Would it be better to make this a general configuration? Right now you it is very specific to two keyspaces, I would make this configurable via flag and not based on the keyspace name.

burmanm: I don't know a way to find out if we're running under a test or not. These are the defaults we use in the tests.

jsanda: In theory I also prefer making it configurable. In practice, I cannot recall when I have actually used different keyspaces for tests.

PR #863, opened 2017-08-09
HWKMETRICS-721

stefannegrea:


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/MetricsServiceImpl.java
```diff
@@ -1072,21 +1074,23 @@ public void shutdown() {
     }
 
     @Override
-    @SuppressWarnings("unchecked")
     public <T> Observable<Void> deleteMetric(MetricId<T> id) {
+        Observable<Void> result = dataAccess.getMetricTags(id)
```
jsanda: How come deleteMetric is reverted?

stefannegrea: It was part of your original commit for logged batches. I will revert this part of the code back to master.

stefannegrea:
This is now ready to merge and brings master on parity with release/0.27.0 branch.

PR #864, opened 2017-08-10
HWKMETRICS-721 release

stefannegrea:


jsanda:
The changes look good, but I do not think we can merge this without some changes to heapster.

stefannegrea:
@jsanda, is the fix not complete without changes to Heapster? Or this change completely breaks Heapster integration?

jsanda:
I will need @burmanm to confirm, but I think there are two places when tags may get updated:

* When the sink is created, which presumably is start up
* When new metrics are added (only tags for new metrics are sent)

We cannot require a restart of heapster to get the tags in sync. There needs to be a mechanism for heapster to automatically retry failed updated and do so until either heapster is shutdown or the updates succeed. If heapster is shutdown, the tag updates should get sent again upon restart.

jsanda:
There is something else we might consider doing to help retry scenarios. If the updates to metrics_tags_index succeed and the updates to metrics_idx fail, then we could return a specific error code to let heapster know some of the updates succeeded. Then on a subsequent retry heapster could include a flag in the request that lets us know we only need to update metrics_idx. This would help reduce the load on Cassandra and increase the likelihood of the retry succeeding.

stefannegrea:
@jsanda, there are several statements executed to insert tags in the tags index and only a single statement for the metrics index. So most of the time will be spent inserting tags in the tags index. In case of error, I would prefer the user to retry everything. If the entire call succeeds then we are guaranteed that the two tables have the information synced.

If we are to create special error messages depending on what statement fails, then we would need to track all the inserts/updates in this method to be consistent. 

Does Heapster retry immediately on failure? Or it puts the failed metric in a queue and retries later?

jsanda:
I am not suggesting we track the success or failure of each update. I am suggesting that we take note of whether **all** updates to metrics_tags_index succeed. It is an easy optimization that increases the likelihood of retries succeeding.

jsanda:
@mwringe can you confirm that Heapster will retry the tag updates? It looks like I missed that when looking at the code.

PR #866, opened 2017-08-11
UI scm: use tag instead of branch; update to 0.0.2

jotak:
Keeping using branch for SCM would inevitably cause difficulties later


stefannegrea:
Thank you @jotak!

PR #874, opened 2017-08-16
Set default JDK to openjdk8

josejulio:


rhqci:

stefannegrea:
ok to test

PR #878, opened 2017-08-22
[HWKMETRICS-729] Update compressor to use improved Gorilla compressor

burmanm:
Improves the performance of the compression job (although the bottleneck is the Cassandra) and improves the compression ratio for availability / counter types.


stefannegrea:
@burmanm, does this break backwards compatibility? Does older data need to be migrated?

burmanm:
No, it supports reading old data (we have a header in the data that tells how it was compressed)

burmanm:
I'll close this to test a little bit without notifications

PR #886, opened 2017-09-15
[HWKMETRICS-734] Use the same version of netty delivered with the Haw…

stefannegrea:
…kular Alerting artifacts.


stefannegrea:
The dependency tree is really wrong for this release. I think we need to re-release Alerting first and only then we might have a chance of getting the correct tree. This needs a lot more investigation for the master branch.

stefannegrea:
Fixed the dependency mismatch by updating Hawkular Alerting to use Hawkular Parent version 55.

jshaughn:
LGTM.

PR #887, opened 2017-09-19
HWKMETRICS-736 (openshift) add explicit Cache-Control header

jotak:
Default to no-cache, but also add a config key to possibly override it.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CacheControlFilter.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/CacheControlFilter.java
@@ -37,8 +37,6 @@ import org.hawkular.metrics.api.jaxrs.config.ConfigurationProperty;
 @Provider
 public class CacheControlFilter implements ContainerResponseFilter {
 
-    public static final String HEADER_KEY = "Cache-Control";
-
     @Inject
     @Configurable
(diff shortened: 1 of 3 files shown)


jotak:
Question: since it's related to ticket https://bugzilla.redhat.com/show_bug.cgi?id=1492011 (openshift-online) , I assume this is for branch release/0.27.0 , is that correct?

Also, I'm adding here that header for all responses, regardless if it's GET, POST or something else... Another option would be to create several config keys, one per http method, but I think it's ok to go with the most simple for now.

stefannegrea:
@jotak, we need a JIRA opened for all release branch PRs. Do you mind opening a JIRA and updating the text of the commit and PR with it?

stefannegrea:
@jotak, please open a PR for master branch too.

jotak:
I've updated the PR because I noticed a curious behaviour: the headers from filters are not set on the jax-rs endpoint for serving static HTML (maybe due to the redirect?).

So they're now added explicitly, both CORS and Cache-Control headers.

jotak:
@stefannegrea , I will patch the master branch too when this one is reviewed and accepted, ok?

jotak:
Adding "Vary: Origin,Accept-Encoding" as well so that client is aware that the response may differ according to origin and expected encoding (ex. /hawkular/metrics has content negotiation html/json)

mwringe:
LGTM

PR #888, opened 2017-09-20
HWKMETRICS-736 (openshift) add explicit Cache-Control header

jotak:
- Default to no-cache, but also add a config key to possibly override it
- For static HTML, use "private, max-age=86400"
- Explicitly add headers before redirect in BaseHandler on "/" when serving static html
- Add also "Vary" header for Origin and Accept-encoding


jotak:
About travis failure: I can't see the full logs so I don't know why it fails... but it passes at home.

mwringe:
LGTM

stefannegrea:
Thank you @jotak!

PR #909, opened 2018-02-10
[HWKMETRICS-751] Improve simple tag query performance

burmanm:
Enables the A-cost query execution (exact matches). This breaks previous behavior and as such can be disabled with a flag.

However, the advantage with large tags partitions should be huge. Needs a better regexp matching function still.

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccess.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccess.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2018 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -118,7 +118,7 @@ public interface DataAccess {
 
     Observable<Row> findMetricsByTagName(String tenantId, String tag);
 
-    Observable<Row> findMetricsByTagNameValue(String tenantId, String tag, String tvalue);
+    Observable<Row> findMetricsByTagNameValue(String tenantId, String tag, String ... tvalues);
 
     Observable<Row> findAllMetricsFromTagsIndex();
 
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/DataAccessImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 Red Hat, Inc. and/or its affiliates
+ * Copyright 2014-2018 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -30,6 +30,7 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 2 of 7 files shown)


burmanm:
With Cassandra 3.11.1 and 1 million metrics with the same tags (thus a single partition - ~72MB), the results are as following. Querying a single metric with "pod_id:567432" (repeting for hundred times). Avg time per query:

* Before: >3000ms 
* After: <100ms

So around 30x improvement. I'll check with Cassandra 3.0.15 if I can get to enough close to same query times.

burmanm:
3.0.15 shows similar performance with 1M pods.

burmanm:
Tested with 4 million tags:

metrics_tags_idx:

Compacted partition maximum bytes: 186563160

metrics_idx:

Compacted partition maximum bytes: 322381140

Still under <100ms per operation. I won't be doing larger tests, but I will improve the regexp detection a bit (while investigating the odd test failures that have nothing to do with this PR)

jsanda:
The ```POST hawkular/metrics/m/stats/query``` endpoint is the one with the most performance problems. There are a couple queries in the OpenShift console of which I am aware that use this endpoint. The tags queries look like:

```
descriptor_name:memory/usage|cpu/usage_rate,type:pod_container,pod_id:871497e0-16b6-11e8-985b-54e1ad486be8|877c6263-16b6-11e8-985b-54e1ad486be8|87148f1b-16b6-11e8-985b-54e1ad486be8|87123431-16b6-11e8-985b-54e1ad486be8,container_name:nodejs
```

This PR still uses a regex for the ```pod_id``` filter. This is the part we need to optimize because the number of pod ids can be arbitrarily large.

burmanm:
Well, it really is regexp query.. ;). But that one can be optimized to use multiple single queries

jsanda:
I think the performance of multiple, single queries vs a single query that
does in memory filtering might actually perform worse as the result size
increases because a for N rows in the result set we would have N queries.

I think it would be interesting to see how a single query that uses the IN
clause performs. It would eliminate the need for in-memory filtering.

burmanm:
Not necessarily as we need to fetch from metrics_idx if we want to do in-memory filtering against other tags. To properly solve something like:

a=b AND c=d|e

Where a=b returns 1000 rows and c=d|e returns 2 rows would require more information than what we currently have. We would need to know an estimate (exact is not needed) of sizes they would return. Easy in single-node-mode though, since we could build a hyperloglog or equal to keep the estimated size. But probably out of context here.

jsanda:
Why would we have to fetch from metrics_idx? I thought we only needed to query metrics_idx when the ```!``` operator is used.

burmanm:
To get all the tags associated with the metric definition for proper filtering. For example, imagine a case where the query would be:

``a1=1, a2=2, a3=3, .. ,an=n``

If we used only metrics_tags_idx, we would have to do ``n`` queries to the Cassandra to fetch all the possible metric definition ids for filtering (and do intersect for each tag). With metrics_idx we could reduce this (we have no knowledge which one is cheaper obviously without estimates).

``a1 = 1`` to metrics_tags_idx and then metrics_idx to fetch all the tag definitions for each of the returned metrics_idx. Rest of the queries can be done on the Hawkular-Metrics by filtering through the tags-map to check if rest of tags match.

This can greatly reduce the amount of data fetched from the Cassandra. There are some test-cases which demonstrate this, but for a short overview from previous ``a1=1..`` example.

```
a1=1 -> return metricId A
a2=2 -> return metricId A, B
a3=3 -> return metricId A, B, C
a4=4 -> return metricId A, B, C, D
...
```

If we use metrics_idx, we would do ``a1=1`` query to Cassandra and fetch the metrics_idx for that. Since all the queries are ``AND`` queries we only need to check if metricId ``A`` has also ``a2=2, a3=3, ...``. No need to do those other queries against metrics_tags_idx

jsanda:
SimpleTagQueryParserTest is using TestNG annotation but the test phase is expecting JUnit annotations. I do not think it is getting executed.

jsanda:
Code changes look good, but there are a couple house keeping items we need. @burmanm can you please create a Jira ticket for this with a fix version of 0.30.0? And then please squash the commits and references the ticket. I will then create a ticket for the release/0.29.0 branch and merge the changes there.

PR #915, opened 2018-03-07
[HWKMETRICS-756] schema installer

jsanda:


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
```diff
@@ -0,0 +1,67 @@
...
+                    getVersion = session.prepare("SELECT value FROM " + keyspace + ".sys_config " +
+                            "WHERE config_id = ? AND name = ?");
+                }
+                ResultSet resultSet = session.execute(getVersion.bind(configId, configName));
+                if (!resultSet.isExhausted() && version.equals(resultSet.one().getString(0))) {
+                    logger.infof("Hawkular Metrics is up to date at version %s", version);
+                    return;
+                }
+            } catch (Exception e) {
+                if (logger.isDebugEnabled()) {
```
burmanm: .. what's the point of this? Why couldn't it just be an infof logging even if debug is enabled?

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
```diff
@@ -0,0 +1,67 @@
...
+                    logger.debug("Version check failed", e);
+                } else {
+                    logger.infof("Version check failed: %s", e.getMessage());
+                }
+                logger.infof("Trying again in %d ms", delay);
```
burmanm: This on the other hand I would move to debugf. And "Version check failed" if we want to put it should perhaps be WARN or ERROR instead..?

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
```diff
@@ -88,6 +88,9 @@
     INGEST_MAX_RETRIES("hawkular.metrics.ingestion.retry.max-retries", null, "INGEST_MAX_RETRIES", false),
     INGEST_MAX_RETRY_DELAY("hawkular.metrics.ingestion.retry.max-delay", null, "INGEST_MAX_RETRY_DELAY", false),
 
+    VERSION_CHECK_DELAY("hawkular.metrics.version.check.delay", "10", "VERSION_CHECK_DELAY", false),
```
burmanm: ..unit is..? seconds?

burmanm: Thread.sleep indicates milliseconds though.. I would increase this value and decrease the max retries.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -66,51 +65,58 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
         URI script = getScript();
         cassalog.execute(script, tags, vars);
 
-        session.execute("INSERT INTO " + keyspace + ".sys_config (config_id, name, value) VALUES " +
-                "('org.hawkular.metrics', 'version', '" + getNewHawkularMetricsVersion() + "')");
+        if (updateVersion) {
+            try {
+                updateVersion(session, keyspace, 0, 0);
+            } catch (Exception e) {
```
burmanm: InterruptedException is transformed to VersionUpdateException? I'm slightly confused with these Exception transformations through something that catches RuntimeExceptions also

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -66,51 +65,58 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
...
+                    updateVersion = session.prepare(
+                            "INSERT INTO " + keyspace + ".sys_config (config_id, name, value) VALUES (?, ?, ?)");
+                }
+                session.execute(updateVersion.bind(configId, configName, version));
+                logger.info("Updated system configuration to version {}", version);
+                break;
+            } catch (Exception e) {
+                retries++;
+                if (retries > maxRetries) {
+                    throw new RuntimeException("Failed to update version in system configuration after " + maxRetries +
```
burmanm: Shouldn't this throw VersionUpdateException? ;)

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -66,51 +65,58 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
...
-    @SuppressWarnings("unused")
-    private String getCurrentHawkularMetricsVersion(Session session, String keyspace) {
-        Statement statement = new SimpleStatement("SELECT value FROM " + keyspace + ".sys_config WHERE " +
-                "name = 'org.hawkular.metrics.version'");
-        ResultSet resultSet = session.execute(statement);
-        if (resultSet.isExhausted()) {
-            return null;
+        while (true) {
+            try {
+                if (updateVersion == null) {
```
burmanm: Why not do it before the loop? No point checking here each time if it was prepared? Also, is filling preparedStatementCache useful here? If this is only used in the startup.

core/schema/src/main/java/org/hawkular/metrics/schema/VersionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+            while (resources.hasMoreElements()) {
+                URL resource = resources.nextElement();
+                try (InputStream stream = resource.openStream()) {
+                    Manifest manifest = new Manifest(stream);
+                    String vendorId = manifest.getMainAttributes().getValue("Implementation-Vendor-Id");
+                    String implVersion = manifest.getMainAttributes().getValue("Implementation-Version");
+                    String gitSHA = manifest.getMainAttributes().getValue("Built-From-Git-SHA1");
+
+                    if (isValidManifest(vendorId, implVersion, gitSHA)) {
+                        return implVersion + "+" + gitSHA.substring(0, 10);
```
burmanm: 0, 10 ? What's the logic behind this selection? Theoretically git shortens to shortest possible unambiguous SHA1. But that can change. It just happens to be 9 bytes in our current repository.

core/schema/src/main/java/org/hawkular/metrics/schema/VersionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+                    }
+                }
+            }
+            throw new RuntimeException("Unable to determine implementation version for Hawkular Metrics");
+        } catch (IOException e) {
+            throw new RuntimeException("There was an I/O error reading META-INF/MANIFEST.MF", e);
```
burmanm: ..corrupted archive?

burmanm:
Additional question in the JIRA ticket (I couldn't tell from this PR)

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
jsanda: I only log the exception stack trace with the debug statement.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
jsanda: I think INFO is appropriate for the "Version checked failed" message because it is something we anticipate happening and handle. We can not assume that Cassandra will already be up and accepting client requests. If we could make that assumption, then I think WARN would be more appropriate.

api/metrics-api-util/src/main/java/org/hawkular/metrics/api/jaxrs/config/ConfigurationKey.java
jsanda: The value is in seconds. I could rename the vars to something like hawkular.metrics.version.check.delay-sec and VERSION_CHECK_DELAY_SEC. I thought a bit about what to use for defaults. The key for me is that I settled on a overall time of 3 minutes to way for schema updates to finish. If after 3 minutes schema updates are not finishes, there will be a container restart for hawkular-metrics. 

I don't want to go to high for the delay because I think upgrades are going to be quite common. In most upgrade scenarios we only apply a relatively small amount of schema changes that should complete quickly.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
jsanda: I will clean up the exception handling.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
jsanda: Yeah, there is no need to do it in the loop. This code is executed in one of two ways, either by calling ```SchemaService.run``` or by directly calling ```SchemaService.updateVersion```. In the former case we know that the schema already exists, and in the latter case we assume that it exists.  I will refactor this code.

core/schema/src/main/java/org/hawkular/metrics/schema/VersionUtil.java
jsanda: Based on what I read [here](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1), I think 10 characters is more than sufficient.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
burmanm: If we assume it's going to happen and we retry because of that, does it make sense to push it to the logs then? It only adds noise if it's something that should be ignored in any case (since it's not final state).

jsanda:
@burmanm can you do another review?

PR #923, opened 2018-03-27
[HWKMETRICS-765] remove DeleteExpiredMetrics job and related code

jsanda:


burmanm:
This PR does not remove expirationIndex writers from DataAccessImpl.updateMetricExpirationIndex and it's still called from MetricsServiceImpl compression code (this should result in a failure since the table is gone)

PR #926, opened 2018-03-28
[HWKMETRICS-761] Log REST endpoint details when there is an error

rubenvp8510:
Do not merge yet.

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/log/HttpErrorLoggerExtension.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/log/HttpErrorLoggerExtension.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2014-2018 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


rubenvp8510:
This has an error when tried to read the request body.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/log/HttpErrorLoggerExtension.java
```diff
@@ -0,0 +1,130 @@
...
+
+        private HttpHandler next;
+        Pattern postQueryPattern;
+
+        public HttpErrorHandler(HttpHandler next, Pattern postQueryPattern) {
+            this.next = next;
+            this.postQueryPattern = postQueryPattern;
+        }
+
+        private void getRequestBody(HttpServerExchange exchange) {
```
burmanm: This method allocates huge amount of stuff, once the bytes from the stream to the memory, then a byte array and then a String from that byte array.

rubenvp8510: Yeah I know, indeed this way of reading the request body doesn't work very well :(

jsanda:
@rubenvp8510 can you please lower the logging level to WARN.

rubenvp8510:
@jsanda Yes, I already set the level to WARN.

PR #927, opened 2018-03-28
[HWKMETRICS-771] compressBlock will check for all possible blocks to compress

burmanm:
WIP, will check for bugs later..

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/servlet/OpenshiftServlet.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/servlet/OpenshiftServlet.java
@@ -66,15 +66,29 @@ public class OpenshiftServlet extends HttpServlet {
         resp.setContentType("application/json");
         AsyncContext asyncContext = getAsyncContext(req);
 
+        PublishSubject<byte[]> byteSubject = PublishSubject.create();
+
         Observable<Metric<Object>> metricObservable = metricsService.scanAllMetricIndexes()
                 .filter(m -> m.getTags().containsKey(DESCRIPTOR_TAG))
                 .onBackpressureBuffer();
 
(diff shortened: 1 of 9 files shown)


jsanda:
I have restarted travis several times. It consistently fails with the now infamous:

```
Failed tests: 
  DataAccessITest.testFindAllDataFromBucket:271 Number of onNext events differ; expected: 10000, actual: 0 (1 completion)
```

jsanda:
With the new flatMap call in MetricsServiceImpl.compressBlock, could we compress multiple tables concurrently? Even aside from the job scheduler bug, I think it is quite possible to end up with tables that do not compressed when they are supposed to. If there is any error, e.g., read or write timeout, the observables will fail fast.

jsanda:
I would like to see some logging in MetricsServiceImpl.compressBlock:

* Log a warning on error
* Debug message that lists the timestamps emitted from datAccess.findExpiredTables

burmanm:
There's concurrency for reading data for a single table:

public Completable compressBlock(long jobStartTimeSlice, int pageSize, int maxConcurrency) {

But it's not used (I didn't find it necessary, the compression job is quite quick and if we increase concurrency we might harm the system's stability)

jsanda:
> But it's not used (I didn't find it necessary, the compression job is quite quick and if we increase concurrency we might harm the system's stability)

Increased concurrency is my concern as well and why I ask. I am referring to this outer most flatMap:

```java
    public Completable compressBlock(long jobStartTimeSlice, int pageSize, int maxConcurrency) {
        return Completable.fromObservable(
                Observable.from(dataAccess.findExpiredTables(jobStartTimeSlice))
                        // Will this flatMap result in concurrent execution?
                        .flatMap(startTimeSlice ->
                        // ...
                        );
        // ....
        );
}
```

jsanda:
@burmanm forgot to ask before, can you please create a ticket for the PR? Then I will merge. Thanks.

PR #928, opened 2018-03-29
[HWKMETRICS-763] fix temp tables not getting dropped

jsanda:


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TempTablesCleaner.java
```diff
@@ -0,0 +1,115 @@
...
+    private volatile boolean finished;
+
+    private static final String DROP_TABLE_CQL = "DROP TABLE IF EXISTS %s";
+
+    public TempTablesCleaner(RxSession session, DataAccessImpl dataAccess, String keyspace, int ttl) {
+        this.session = session;
+        this.dataAccess = dataAccess;
+        this.ttl = Days.days(ttl).toStandardDuration().getMillis();
+
+        findTables = session.getSession().prepare(
```
burmanm: Other option could be to use the PR #927 findExpiredTables.

jsanda: Doesn't findExpiredTables just return timestamps?

burmanm: Sure, but for those we have ``dataAccess.dropTempTable(startTimeSlice)``

burmanm: mm, actually expired tables is actually just pushing tables that should have been compressed, not dropped completely..

jsanda:
I guess it is a bit confusing that “expired” is used both here and in
DataAccessImpl and refers to different things.

core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/TempTablesCleaner.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private Observable<String> dropTable(String table) {
+        return session.execute(String.format(DROP_TABLE_CQL, table))
+                .map(resultSet -> table)
+                .onErrorResumeNext(t -> {
+                    // If there is an error, we do not retry because it is possible that the table has already been
+                    // dropped. We will instead wait until findTables runs again and retry dropping the table then
+                    // if dropping it did indeed fail for some reason.
+                    logger.infof(t, "Failed to drop %s", table);
+                    return Observable.just("");
```
burmanm: Would ``Observable.empty()`` instead and removing the filter part work as well?

jsanda:
@burmanm I made the changes.

PR #930, opened 2018-04-02
[HWKMETRICS-770] release/0.30.0 branch

jsanda:


api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/handler/MetricHandler.java
```diff
@@ -302,6 +306,10 @@ public void addMetricsData(
     public void findStats(@Suspended AsyncResponse asyncResponse, StatsQueryRequest query) {
         try {
             checkRequiredParams(query);
+
```
rubenvp8510: This always logs body, even if this call doesn't fail (of course if the log level is on debug), that will produce some noise on logs.

PR #952, opened 2018-04-18
[HWKMETRICS-783] Schema installer should not use git hash for schema …

rubenvp8510:
…version

### DIFF
--- a/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
+++ b/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
@@ -16,12 +16,12 @@
  */
 package org.hawkular.metrics.api.jaxrs.util;
 
+import org.hawkular.metrics.schema.SchemaService;
 import org.jboss.logging.Logger;
 
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.Session;
-
 /**
  * @author jsanda
  */
@@ -29,12 +29,14 @@ public class SchemaVersionChecker {
 
     private static final Logger logger = Logger.getLogger(SchemaVersionChecker.class);
 
-    public void waitForSchemaUpdates(Session session, String keyspace, String version, long delay, int maxRetries)
+    public void waitForSchemaUpdates(Session session, String keyspace, long delay, int maxRetries)
             throws InterruptedException {
         String configId = "org.hawkular.metrics";
-        String configName = "version";
+        String configName = "schema-version";
         int retries = 0;
         PreparedStatement getVersion = null;
+        SchemaService schemaService = new SchemaService(session, keyspace);
+        String version = schemaService.getSchemaVersion();
 
         while (maxRetries > retries) {
             try {
@@ -43,9 +45,17 @@ public class SchemaVersionChecker {
                             "WHERE config_id = ? AND name = ?");
                 }
                 ResultSet resultSet = session.execute(getVersion.bind(configId, configName));
-                if (!resultSet.isExhausted() && version.equals(resultSet.one().getString(0))) {
-                    logger.infof("Hawkular Metrics is up to date at version %s", version);
-                    return;
+                if (!resultSet.isExhausted()) {
+                    String schemaVersion = resultSet.one().getString(0);
+                    if(version.equals(schemaVersion)) {
+                        logger.infof("Hawkular Metrics schema is up to date at version %s", version);
+                        return;
+                    } else {
+                        logger.infof("Hawkular Metrics schema version does not match," +
+                                        " should be %s, present: %s, Trying again in %d ms",
+                                version, schemaVersion, delay);
+
+                    }
                 }
             } catch (Exception e) {
                 if (logger.isDebugEnabled()) {
--- a/core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
+++ b/core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
@@ -25,11 +25,13 @@ import java.util.Map;
 
 import org.cassalog.core.Cassalog;
 import org.cassalog.core.CassalogBuilder;
+import org.cassalog.core.ChangeSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.Session;
-import com.datastax.driver.core.SimpleStatement;
 import com.google.common.collect.ImmutableMap;
 
 /**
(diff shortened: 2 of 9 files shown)


core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -50,56 +51,91 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
...
+    public String getSchemaVersion(Session session, String keyspace) {
+        URI script = getScript();
+        CassalogBuilder builder = new CassalogBuilder();
+        Cassalog cassalog = builder.withKeyspace(keyspace).withSession(session).build();
+        Map<String, ?> vars = this.getVars(session, keyspace, false, 1);
+        List<ChangeSet> changeSets = cassalog.load(script, this.getTags(), vars);
+        if (changeSets.size() > 0) {
+            return changeSets.get(changeSets.size() - 1).getVersion();
+        }
+        return "";
```
jsanda: I feel like returning an empty string here is akin to swallowing an exception. In other words, ```changeSets``` should not be empty. If it is, then something is wrong, and I think we ought to throw an exception.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -50,56 +51,91 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
...
+        String version = this.getSchemaVersion(session, keyspace);
+        writeVersion("schema-version", version, session, keyspace, delay, maxRetries);
+    }
+
+    private void writeVersion(String versionType, String version, Session session, String keyspace, long delay, int
+            maxRetries) {
+        String configId = "org.hawkular.metrics";
         SimpleStatement updateVersion = new SimpleStatement(
-                "INSERT INTO sys_config (config_id, name, value) VALUES (?, ?, ?)", configId, configName, version)
+                "INSERT INTO sys_config (config_id, name, value) VALUES (?, ?, ?)", configId, versionType, version)
```
jsanda: Since this query will not get executed multiple times, we should make it a PreparedStatement.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -50,56 +51,91 @@ public void run(Session session, String keyspace, boolean resetDB, int replicati
     public void run(Session session, String keyspace, boolean resetDB, int replicationFactor, boolean updateVersion) {
         CassalogBuilder builder = new CassalogBuilder();
         Cassalog cassalog = builder.withKeyspace(keyspace).withSession(session).build();
-        Map<String, ?> vars  = ImmutableMap.of(
+        URI script = getScript();
+        Map<String, ?> vars = this.getVars(session, keyspace, resetDB, replicationFactor);
+        cassalog.execute(script, this.getTags(), vars);
```
jsanda: We want to update the schema version after each update is applied. If ```updateVersion``` is true then you will want to use the new version of the execute method, 
```java
void execute(URI script, Collection tags, Map vars, ChangeSetListener listener)
```

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
rubenvp8510: +1, I can throw an exception and log an error here, just in case.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
jsanda: I went back and looked at my notes in Trello to see why I thought we needed to use the ChangeSetListener. It is not clear to me now, but more importantly, I agree with your assessment that it is not needed.

api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/util/SchemaVersionChecker.java
```diff
@@ -43,9 +45,17 @@ public void waitForSchemaUpdates(Session session, String keyspace, String versio
                             "WHERE config_id = ? AND name = ?");
                 }
                 ResultSet resultSet = session.execute(getVersion.bind(configId, configName));
-                if (!resultSet.isExhausted() && version.equals(resultSet.one().getString(0))) {
-                    logger.infof("Hawkular Metrics is up to date at version %s", version);
-                    return;
+                if (!resultSet.isExhausted()) {
+                    String schemaVersion = resultSet.one().getString(0);
+                    if(version.equals(schemaVersion)) {
+                        logger.infof("Hawkular Metrics is up to date at version %s", version);
```
jsanda: I would change this slightly to, 

"Hawkular Metrics schema is up to date at version %s"

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -39,67 +41,125 @@
...
+    public void run(boolean resetDB) {
+        run(resetDB, 1);
     }
 
-    public void run(Session session, String keyspace, boolean resetDB, int replicationFactor, boolean updateVersion) {
+    public void run(boolean resetDB, int replicationFactor) {
+        run(resetDB, replicationFactor, false);
+    }
+
+    public void run(boolean resetDB, int replicationFactor, boolean updateVersion) {
```
jsanda: SchemaService has evolved a bit since it was originally written. If you want feel free to refactor this and inject ```resetDB```, ```replicationFactor```, and ```updateVersion``` in the constructor and/or setter methods. Then we won't have to call ```buildCassalogIfNeeded``` in multiple places.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
```diff
@@ -39,67 +41,125 @@
...
-                "keyspace", keyspace,
-                "reset", resetDB,
-                "session", session,
-                "replicationFactor", replicationFactor,
-                "logger", logger
-        );
+        cassalog = builder.withKeyspace(keyspace).withSession(session).build();
+    }
+
+    private Map<String, ?> getVars(boolean resetDB, int replicationFactor, boolean createSchema) {
```
jsanda: I don't think we want the ```createSchema``` variable. It is hard coded to true, so I don't see any need for it. It could potentially break the schema installs/upgrade which I will describe below.

core/schema/src/main/resources/org/hawkular/schema/bootstrap.groovy
```diff
@@ -55,19 +55,20 @@ if (keyspace.equals("hawkulartest") || keyspace.equals("hawkular_metrics_rest_te
   keyspaceCreate = "$keyspaceCreate AND DURABLE_WRITES = 'false'"
 }
 
-executeCQL(keyspaceCreate)
-executeCQL("USE $keyspace")
+if(createSchema) {
```
jsanda: There is another reason we do not want ```createSchema```. In the event that ```resetDB``` is true and ```createSchema``` is false, then schema installation breaks. Each of the commands in this script is idempontent. It is safe for them to be run multiple times.

core/schema/src/main/java/org/hawkular/metrics/schema/SchemaService.java
jsanda: They keyspace does not have to exist in advance. Building Cassalog only initializes its properties. Cassalog does not try connecting to Cassandra on initialization.

core/schema/src/main/resources/org/hawkular/schema/bootstrap.groovy
```diff
@@ -44,30 +44,35 @@ def createType(String type, String cql) {
   }
 }
 
-if (reset) {
-  executeCQL("DROP KEYSPACE IF EXISTS $keyspace", 20000)
-}
+bootstrap {
+  println "BOOTSTRAPPING"
```
jsanda: @rubenvp8510 can you remove this ```println``` and then I will merge. Thanks!

rubenvp8510:
Ok, I removed the println and put the cassalog build on constructor, if/when this is on green we can merge :)

PR #960, opened 2018-05-01
Optimize Openshift tag queries

burmanm:
@jsanda Something like this (untested)

### DIFF
--- a/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/tags/SimpleTagQueryParser.java
+++ b/core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/tags/SimpleTagQueryParser.java
@@ -118,6 +118,10 @@ public class SimpleTagQueryParser {
 
         enum RegExpOptimizer {OR_SINGLE_SEEK, NONE }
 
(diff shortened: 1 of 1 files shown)


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/tags/SimpleTagQueryParser.java
```diff
@@ -199,13 +216,37 @@ public static boolean isRegExp(String tagValuesQuery) {
         Map<Long, List<Query>> costSortedMap =
                 QueryOptimizer.reOrderTagsQuery(tagsQueries, enableACostQueries);
 
+        List<Query> openshiftQuery = costSortedMap.get(QueryOptimizer.OPENSHIFT_OPTIMIZED_TAG_QUERY);
```
jsanda: This should be:
```
costSortedMap.get(QueryOptimizer.GROUP_OPENSHIFT_OPTIMIZATION)
```

burmanm: Ah, now the devil of autocorrection hits even the IDEs ;)

burmanm:
Closed in favor of #961

PR #966, opened 2018-05-24
[HWKMETRICS-789] fix filtering on groupAOREntries to prevent NPE

jsanda:
Also update test to cover scenario that triggered the NPE.


core/metrics-core-service/src/main/java/org/hawkular/metrics/core/service/tags/SimpleTagQueryParser.java
```diff
@@ -552,13 +565,18 @@ private void logQuery(String tenantId, Query query, String queryType, Set<? exte
...
+            try {
+                String tagValue = tMetric.getTags().get(tagName);
+                if(tagValue != null) {
+                    boolean positive = (!regexp.startsWith("!"));
+                    Pattern p = PatternUtil.filterPattern(regexp);
+                    return positive == p.matcher(tagValue).matches(); // XNOR
+                }
+                return false;
+            } catch (Exception e) {
+                logger.warn("tagValueFilter failed", e);
```
burmanm: This breaks the semantics. You will never catch the doOnError since you don't actually propagate the Exception.

PR #992, opened 2018-09-19
Revert "[HWKMETRICS-793] Create stand alone schema installer (#978)"

rubenvp8510:
This reverts commit 90aebd4b00fbf6182b0fdc1f13c568d55d12c614.

@jsanda I need to revert his commit due issues backporting the schema installer, but I think we want to alerts removed from this, because of this I think we need to do another release of this just without alerts but also without this backport. WDYT?


jsanda:
Sounds good me. Travis is complaining about a license header missing.

PR #993, opened 2018-11-05
Revert "[BZ 1590451] Schema Installer. (#988)"

rubenvp8510:
We backported the schema installer to 0.28.0. but due to some issues related to installation and imagen absence, we are reverting it.

I'm doing this PR because there are some fixes on the latest release, but I need those fixes but no the schema installer to be included in the next release for 0.28 branch, this would be needed for other old branches where the schema installer was backported.

Unfortunately for this branch the commit was squashed with other commits that removes unused code, in this PR I'm preserving the remotion of the unused code and tried to revert ONLY the schema installer stuff.


jsanda:
LGTM

PR #994, opened 2018-11-07
[HWKMETRICS-801] return a proper JSON object as expected by Heapster …

jmartisk:
https://issues.jboss.org/browse/HWKMETRICS-801
https://bugzilla.redhat.com/show_bug.cgi?id=1506736

BTW, @jsanda @rubenvp8510  shouldn't we first change the project version on this branch to 0.31.1-SNAPSHOT or something like that? I kinda don't like the approach where we keep the previous Final version in pom.xml until we release a new Final version... Just a suggestion.


rubenvp8510:
Now you can rebase and tests should pass, when this is on green we can merge

jmartisk:
Green now. Thanks for fixing the tests.

PR #996, opened 2018-11-12
Fix mirror in order to get correct C* version

rubenvp8510:
@jmartisk Could you please review?

This should solve the travis issue we are seeing in your PR:  https://github.com/hawkular/hawkular-metrics/pull/994

### DIFF
--- a/.travis.install.cassandra.sh
+++ b/.travis.install.cassandra.sh
@@ -24,11 +24,9 @@ cd "${HOME}"
 C_MAJOR="3"
 C_MINOR="0"
 
-# Find the closest Apache mirror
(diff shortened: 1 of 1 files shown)


jmartisk:
LGTM, this mirror seems to have all versions. Don't know why apache.cs.utah.edu has only some specific versions.. Should we do this for older maintenance branches as well? I guess yeah

burmanm:
Cassandra downloads only have the newest version. But we use an older version than the current one -> it's in the archives.

rubenvp8510:
@jmartisk it seems to me like you right about we need to do it for other branches. I'll merge this one :)

PR #1006, opened 2018-12-05
Validate when scheduled jobs has null values

rubenvp8510:


job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
```diff
@@ -195,14 +213,20 @@ public Completable deleteJob(UUID jobId, rx.Scheduler scheduler) {
...
+            logger.warn("Tried to insert job on scheduled jobs with invalid values");
+        }
         return session.execute(insertScheduled.bind(timeSlice, job.getJobId(), job.getJobType(), job.getJobName(),
                 job.getParameters().getMap(), getTriggerValue(session, job.getTrigger())));
     }
 
     public Observable<ResultSet> updateStatusToFinished(Date timeSlice, UUID jobId) {
-        return session.execute(updateStatus.bind((byte) 1, timeSlice, jobId))
+        // First check if the job exists
+        return session.executeAndFetch(findByIdAndSlice.bind(timeSlice, jobId)).flatMap(
```
jsanda: I would consider adding some logging in here. The expectation is that there will be a row for `timeSlice` and `jobId`. At a minimum, I would log a warning that the row does not exist. Then the question is, what if anything else should be done?

rubenvp8510: I think for now just logging and then ignore it and of course don't try to do the upsert operation.

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
```diff
@@ -195,14 +213,28 @@ public Completable deleteJob(UUID jobId, rx.Scheduler scheduler) {
...
                 job.getParameters().getMap(), getTriggerValue(session, job.getTrigger())));
     }
 
     public Observable<ResultSet> updateStatusToFinished(Date timeSlice, UUID jobId) {
-        return session.execute(updateStatus.bind((byte) 1, timeSlice, jobId))
-                .doOnError(t -> logger.warnf("There was an error updating the status to finished for %s in time " +
-                        "slice [%s]", jobId, timeSlice.getTime()));
+        // First check if the job exists
+        return session.executeAndFetch(findByIdAndSlice.bind(timeSlice, jobId))
+                .isEmpty()
```
rubenvp8510: I don't have too much experience on javarx, not sure if this is a good way to do the things.

rubenvp8510:
@jsanda I've already updated the PR with all comments we talked about. 

Thanks for taking the time to review this!

job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
```diff
@@ -66,9 +68,19 @@
...
 
+    private PreparedStatement findByIdAndSlice;
+
     private static final Function<Map<String, String>, Completable> SAVE_PARAMS_NO_OP =
             params -> Completable.complete();
 
+    // Fields that shouldn't be null: time_slice, job_id, job_type, job_name, trigger
+    private Func1<Row, Boolean> filterAllScheduledQuery = row -> !(row.isNull(0) || row.isNull(1)
+            || row.isNull(2) || row.isNull(3) || row.isNull(5));
+
```
jsanda: The functions look good. I'll mention one thing to consider. The filters should help avoid the NPEs but will do so silently. You could add some additional logging at the expense of making the filter functions a bit more verbose, e.g., 

```java
private Func1<Row, Boolean> filterAllScheduledQuery = row -> {
    boolean isNull = !(row.isNull(0) || row.isNull(1) || row.isNull(2) || row.isNull(3) || row.isNull(5));
    if (isNull) {
        // log column names/values...
    }
    return isNull;
}
```

rubenvp8510:
I've already addressed all comments, is ready for other review and merge.

PR #1007, opened 2018-12-17
[HWKMETRICS-802] Validate when scheduled jobs has null values

rubenvp8510:


job-scheduler/src/main/java/org/hawkular/metrics/scheduler/impl/JobsService.java
```diff
@@ -195,14 +217,29 @@ public Completable deleteJob(UUID jobId, rx.Scheduler scheduler) {
...
                 job.getParameters().getMap(), getTriggerValue(session, job.getTrigger())));
     }
 
     public Observable<ResultSet> updateStatusToFinished(Date timeSlice, UUID jobId) {
-        return session.execute(updateStatus.bind((byte) 1, timeSlice, jobId))
-                .doOnError(t -> logger.warnf("There was an error updating the status to finished for %s in time " +
-                        "slice [%s]", jobId, timeSlice.getTime()));
+        // First check if the job exists
+        return session.executeAndFetch(findByIdAndSlice.bind(timeSlice, jobId))
+                .isEmpty()
```
burmanm: I'll comment on this PR as the other one were you presented the question was merged already. I'm not sure what you try to achieve here, since if the previous part does not emit anything, then there's nothing to flatMap in the next section either. So why transform the end event to a boolean?

rubenvp8510: I want to validate if the `findByIdAndSlice` statement returns empty, just for logging purposes. In the case of non-empty I will execute the `updateStatus` statement is in this case when I need to use flatMap.

Essentially I'm transforming a boolean to emitted rows from updateStatus only if the boolean true, otherwise, I'm returning an empty observable. I'm using the flatMap to flat the Observable returned by updateStatus.

Does that makes sense?

burmanm: ```.switchIfEmpty(Observable.empty().doOnComplete(() -> logger.info("Nothing was found!"))``` ?

rubenvp8510: Mm yeah, something like that :)

PR #1017, opened 2019-01-10
Move to 0.31.2.Final-SNAPSHOT

jmartisk:
Should we start doing this, @rubenvp8510 ? It's quite a usual thing to do after releasing a version, I noticed Hawkular Metrics doesn't do that though. If it's for some reason an issue (for example for building development OpenShift images or so), we don't have to do this...


rubenvp8510:
Hi @jmartisk this looks good, we can start working in this way. I don't see any inconvenient :)

PR #1019, opened 2019-05-09
Add SHA256 to password storage types for htpasswd file.

briantward:
Unfortunately I did not see any existing test framework for this package, so I tested the private method individually outside of this project.  I'll test it tomorrow as a patch on the release I need.


briantward:
closing for #1020 instead.

PR #1022, opened 2019-07-03
BZ 1632852 Add retry logic that keep installer running until success.

rubenvp8510:


core/schema-installer/src/main/java/org/hawkular/metrics/schema/Installer.java
```diff
@@ -76,34 +78,41 @@ public Installer() {
         replicationFactor = Integer.getInteger("hawkular.metrics.cassandra.replication-factor", 1);
         versionUpdateDelay = Long.getLong("hawkular.metrics.version-update.delay", 5) * 1000;
         versionUpdateMaxRetries = Integer.getInteger("hawkular.metrics.version-update.max-retries", 10);
+        waitRetry =  Long.getLong("hawkular.metrics.version-update.retry-delay", 10);
     }
 
     public void run() {
+        while (true) {
```
jmartisk: I don't quite like how this could end up running forever without stopping, shouldn't we make use of `versionUpdateMaxRetries` for this cycle as well somehow (or some new variable)?

core/schema-installer/src/main/java/org/hawkular/metrics/schema/Installer.java
```diff
@@ -76,34 +78,41 @@ public Installer() {
...
-
-            SchemaService schemaService = new SchemaService(session, keyspace);
-            schemaService.run(resetdb, replicationFactor, false);
+        Session session = initSession();
+        waitForAllNodesToBeUp(session);
 
-            JobsManager jobsManager = new JobsManager(session);
-            jobsManager.installJobs();
+        SchemaService schemaService = new SchemaService(session, keyspace);
+        schemaService.run(resetdb, replicationFactor, false);
```
jmartisk: Can we be sure that `schemaService` and `jobsManager` operations are idempotent and can be run multiple times? I mean, imagine a case that
- `schemaService.run` completes successfully
- `jobsManager.installJobs` completes sucessfully
- `schemaService.updateVersion` fails, so the whole installation is retried, therefore `schemaService.run` and `jobsManager.installJobs` will run again, even though it was completed successfully before! Not sure if that would install more jobs and if that could be an issue.

Even if they were idempotent, redoing all operations would take more time and increase the probability of another failure. Perhaps we could mark installation steps that are already done and skip them when we are retrying the installation?

rubenvp8510: I'm sure they are idempotent.

core/schema-installer/src/main/java/org/hawkular/metrics/schema/Installer.java
rubenvp8510: Ohh a silly error, TBH I don't like either, I'll put a counter here.

core/schema-installer/src/main/java/org/hawkular/metrics/schema/Installer.java
rubenvp8510: Not sure if there is a way to "mark" the steps, the schema version is managed by cassalog, IIRC it checks the version in a table , then follow some steps for creating tables, column families etc..,  also check for if certain table/column exists or not. That could be a way to "mark" the progress. Or not? other than that I don't know how to put marks on a schema installation process.

 On the other hand if the schema does not finish property, Hawkular Metrics won't run, so we can detect the problem and review the logs and say "hey, this is running forever, because it fails".. Most of the time I saw this fail is due a time-out on C* , if C* is not have good performance for schema modifications  we can assume that even if for some reason this completes successfully, Hawkular Metric will have problems.

Does that make sense for you?

jmartisk: By marking completed items I mean the four high-level steps here (`schemaService.run`, `jobsManager.installJobs`, `schemaService.updateVersion`, `schemaService.updateSchemaVersionSession`) as a whole, not on cassalog level. This could be achieved by simply having four boolean variables that get set to `true` after the respective item is complete so next time we won't have to repeat it. But maybe it's not necessary

rubenvp8510:
@jmartisk Do you think this is ready for be merged? or any other thoughts about this?

jmartisk:
LGTM

PR #1023, opened 2019-07-30
Separate htpasswd utility and package into jar

rubenvp8510:


containers/hawkular-htpasswd/src/main/java/org/hawkular/openshift/htpasswd/CreatePassword.java
```diff
@@ -0,0 +1,104 @@
...
+    private boolean created;
+
+    private PasswordManager passwordManager = new PasswordManager();
+
+    public CreatePassword() {
+        password = System.getProperty("htpasswd.password", "");
+        username = System.getProperty("htpasswd.username", "");
+        algorithm = System.getProperty("htpasswd.algorithm", SHA256_ALGORITHM);
+        fileName = System.getProperty("htpasswd.file", "");
+        created = Boolean.getBoolean("htpasswd.created");
```
jmartisk: Wouldn't `htpasswd.overwrite` be a more appropriate name? IIUC if this is `true` then an existing file should be overwritten, if `false` then a line should be appended, right?

containers/hawkular-htpasswd/src/main/java/org/hawkular/openshift/htpasswd/CreatePassword.java
```diff
@@ -0,0 +1,104 @@
...
+        if (output != null) {
+            try {
+                outputStream.write(username.getBytes());
+                outputStream.write(":".getBytes());
+                outputStream.write(output.getBytes());
+                outputStream.write("\n".getBytes());
+            } catch (IOException e) {
+                throw new RuntimeException(e.getMessage());
+            }
+        } else {
```
jmartisk: The `outputStream` should be closed (unless it's `System.out`). It's probably not a big problem, but just to be clean...

containers/hawkular-htpasswd/src/main/java/org/hawkular/openshift/htpasswd/CreatePassword.java
rubenvp8510: Is right, I took the name from htpasswd utility, but this is a better name.

